PR #25, opened 2025-08-15
User management

mobileoverlord:
Add support for managing real user accounts. Service accounts will be handled by `systemd-sysusers` however, real accounts need to modify or control the user account files in `etc`. Primarily `shadow` `passwd` `group` and `gshadow`. This PR adds an interface for designating a single extension in the config to manage users. 

For development purposes, we support setting accounts with no passwords. Here is an example on how to enable root with no password. For the remainder of the text, we will assume you are doing your user management in an `my-app` extension. We support two ways to declare and organize user and group information in the config. 

Inline syntax:
```
[ext.my-app.users]
root = {password = ""}
```

Table syntax
```
[ext.my-app.users.root]
password = ""
```

The longer form is useful is you want control over a lot of keys for that user since the config file format is TOML and TOML objects do not support multiline breaks 🙄

Here is an example of all the settings supported on users:
```
[ext.my-app.users.appuser]
password = "$6$salt$hash"
uid = 1001
gecos = "Application User"
home = "/opt/appuser"
shell = "/bin/sh"
groups = ["appuser"]
max_days = 90
warn_days = 7
```

This comprehensively covers whats would normally be accomplished via `useradd` / `usermod`. More comprehensive documentation will ensure to illustrate the options, their choices, and the defaults. For example, you can omit all uid / gid's and the system will automatically increment starting at 1000. 

Groups are similar:
Inline syntax
```
[ext.my-ext.groups]
appuser = { gid = 1000 }
```

Table syntax
```
[ext.my-ext.groups.appuser]
gid = 1000
members = ["appuser"]
```

During ext build, the rootfs databases will be copied into the ext sysroot and updated to include the user / group modifications.

### DIFF
--- a/docs/users-and-groups.md
+++ b/docs/users-and-groups.md
@@ -1,0 +1,170 @@
+# User and Group Management in Avocado Extensions
+
+Avocado supports comprehensive user and group management in extensions using a flexible TOML configuration syntax.
+
+## Mixed Syntax Approach (Recommended)
+
+You can use two approaches depending on complexity:
+
+### Simple Users/Groups: Inline Syntax
+For users or groups with minimal configuration, use inline table syntax:
+
+```toml
+[ext.my-extension.users]
+root = { password = "" }
+guest = { password = "*" }
+
+[ext.my-extension.groups]
+users = { gid = 1000 }
+staff = { gid = 20 }
+```
+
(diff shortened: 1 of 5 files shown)


mobileoverlord:
Depends on https://github.com/avocado-linux/meta-avocado/pull/152

PR #151, opened 2026-06-02
feat(connect): support explicit org selection in auth login

lee-reinhardt:
## Summary

Adds multi-org support to `avocado connect auth login`. Users who belong to more than one org can now choose which org their new CLI token will be scoped to, either via a `--org` flag for scripting or via an interactive picker.

## What's new

- `--org <id>` flag scopes the new token to a specific org non-interactively. UUID-only, matching the convention in `connect init` and other org-scoped commands.
- Interactive terminal picker prompts when the flag is absent and the user belongs to multiple orgs. Same picker pattern as `connect init` so the UX feels consistent across the auth and project-setup flows.
- Single-org users see no change — the picker is skipped and the lone org is selected automatically.
- JSON output mode (`--output json`) with multiple orgs and no `--org` returns a clear error rather than hanging on a prompt the caller can't answer.

## Mechanics

- Adds a new step between the browser callback and exchange: `POST /auth/cli/list-orgs {code}` returns the user's orgs and roles without consuming the one-time code (codes are stateless, so the same code is reused for the subsequent exchange call).
- `POST /auth/cli/exchange` now sends `organization_id` in the request body, letting the server scope the token at mint time.

## Server dependency

Requires a server that supports the new `list-orgs` endpoint and the optional `organization_id` param on `exchange`. The corresponding server change is shipping ahead of any CLI release.

## Tests

8 new unit tests cover the org-pick decision logic across every branch (zero/single/multi orgs, hint match/mismatch, JSON vs human output). The pure decision logic (`pick_org`) is split from the IO/UI (HTTP, stdin, stdout) so the tests target pure functions and run without a network or terminal.

## Backwards compatibility

Older `avocado` binaries that don't send `organization_id` continue to work against the new server unchanged — the server preserves the existing scoping behavior for that case. Users opt into the picker by upgrading the CLI.

## Draft status

Marked as draft pending the server-side change rolling out first, so CLIs that upgrade hit a server that already knows about the new endpoint.

### DIFF
--- a/src/commands/connect/auth.rs
+++ b/src/commands/connect/auth.rs
@@ -5,7 +5,9 @@ use chrono::Utc;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::net::TcpListener;
 
-use crate::commands::connect::client::{self, ConnectClient, ConnectConfig, Profile, ProfileUser};
+use crate::commands::connect::client::{
+    self, ConnectClient, ConnectConfig, OrgInfo, Profile, ProfileUser,
+};
 use crate::utils::output::{print_error, print_info, print_success, OutputLevel};
 use crate::utils::output_format::{emit_json_event, emit_json_object};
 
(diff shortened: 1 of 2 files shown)


jasontwong (approved):
LGTM — no blocking issues found.

PR #160, opened 2026-06-25
fix: honor target-<name>: override for extension sdk.packages

mobileoverlord:
## Problem

Per-target overrides inside an extension are resolved inconsistently:

- Runtime `packages:` overrides go through `resolve_overrides_in_value`, which accepts **both** the preferred `target-<name>:` key and the legacy bare `<name>:` key.
- `sdk.packages` overrides were collected by matching **only** the bare `<target>:` key (`ext_config_table.get(target)`), in both the local and external-extension SDK dependency collectors.

So a block like:

```yaml
extensions:
  avocado-ext-imx-npu:
    sdk:
      packages:
        nativesdk-uv: "*"
    target-imx93-frdm:        # preferred override form
      sdk:
        packages:
          nativesdk-ethos-u-vela: "*"
```

had its `target-imx93-frdm.sdk.packages` **silently dropped** — the per-target SDK build tool was never installed into the SDK, even though the same `target-` form worked fine for runtime `packages:`. This surfaced as a build-time `vela: command not found` despite the package being correctly declared.

## Fix

Add a shared `Config::merge_target_sdk_packages` helper that honors **both** the `target-<name>:` and bare `<name>:` keys (bare first, then prefixed so the preferred form wins), and call it from both the local (`get_extension_sdk_dependencies_with_config_path_and_target`) and external (`collect_external_extension_sdk_dependencies_with_target`) SDK collectors. This also de-dupes the two near-identical nested blocks.

The bare key continues to work, so this is backward compatible.

## Test

Adds `test_extension_sdk_dependencies_target_prefix`, a regression test asserting that a `target-<name>:` `sdk.packages` block is merged for the matching target and absent for a non-matching one. Existing bare-key tests still pass.

## Checks

`cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo build`, and `cargo test` all pass locally.

### DIFF
--- a/src/utils/config.rs
+++ b/src/utils/config.rs
@@ -4831,6 +4831,37 @@ impl Config {
 
     /// Get extension SDK dependencies from configuration, including nested external extension dependencies and target-specific dependencies
     /// Returns a HashMap where keys are extension names and values are their SDK dependencies
+    /// Merge an extension's per-target `sdk.packages` into `merged_deps`,
+    /// honoring BOTH the preferred `target-<name>:` override key and the legacy
+    /// bare `<name>:` key. This mirrors how runtime `packages:` overrides resolve
+    /// (via `resolve_overrides_in_value`), so a `target-`-keyed `sdk.packages`
+    /// block is not silently dropped when collecting SDK dependencies.
+    fn merge_target_sdk_packages(
+        ext_config_table: &serde_yaml::Mapping,
+        target: &str,
+        merged_deps: &mut HashMap<String, serde_yaml::Value>,
+    ) {
+        // Bare `<name>:` first, then `target-<name>:`, so the preferred prefixed
+        // form wins if both happen to be present.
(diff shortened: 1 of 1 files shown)


jetm (approved):
The refactor is correct and the simplification is a meaningful improvement over two near-identical deep `if let` chains.

Two minor points:

- **Doc comment placement** (`config.rs:4832`): the existing `///` lines for `get_extension_sdk_dependencies_with_config_path_and_target` are now attached to `merge_target_sdk_packages` (Rust `///` comments bind to the immediately following item). The public function ends up undocumented. Moving the helper insertion point above the existing doc block would fix this.
- **Precedence coverage**: `test_extension_sdk_dependencies_target_prefix` uses distinct packages in each block, so it does not exercise the bare-wins-then-prefixed-wins ordering. A case where the same key appears in both `<target>:` and `target-<target>:` with different values would verify the loop order is correct.

Neither is a blocker.

PR #158, opened 2026-06-24
Add --connect-sign flag to avocado deploy

jetm:
## Problem

Devices that have received a Connect OTA update trust the Connect server key
as their TUF root, so locally-signed TUF metadata is rejected. `avocado deploy`
has no way to sign metadata via Connect, leaving these devices stuck.

## Solution

Add `--connect-sign` to both `avocado deploy` entry points. When set, Phase 2
routes through a new `ConnectClient::sign_for_deploy` method that POSTs the
target list to `/api/orgs/{org}/signing/sign-for-deploy` and writes the four
returned JSON blobs to staging. The local path is unchanged; `--connect-sign`
is an explicit opt-in.

A nudge prints on local deploys when the project has `connect.server_key` set,
surfacing the flag without forcing it.

## Key changes

- `TargetFileInfo` gains `Serialize` so the targets array goes directly into the request body
- `ConnectClient::sign_for_deploy` and response types added, mirroring the `get_tuf_server_key` pattern
- Phase 2 in `deploy.rs` branches on `--connect-sign`: resolves org early, calls the API, writes the four TUF JSON files, prints 7-day expiry notice
- Phase 2 progress label switches to "Signing TUF metadata via Connect..." under `--connect-sign`
- `root.json` write moved outside the if/else so both signing paths stage it
- Flag wired to both `Commands::Deploy` and `RuntimeCommands::Deploy` in `main.rs`
- `config show --output json` now emits a per-runtime `signing_enabled` boolean so Avocado Desktop can gate "Sign via Connect" (unblocks ENG-2005)
- `.pre-commit-config.yaml` added: cargo fmt/clippy/check on commit, cargo audit on push

## Reviewer notes

**Gated on ENG-1963** — the `sign-for-deploy` endpoint is not live yet. Unit
tests pass; a live E2E needs three things in place:

1. ENG-1963 endpoint landed and reachable
2. A device that has taken a Connect OTA (rejects local signing as baseline)
3. An authenticated Connect session with a valid org

E2E sequence once ready:
```sh
# Confirm baseline failure
avocado deploy -r <runtime> -d <device-ip>

# Connect-signed deploy
avocado deploy --connect-sign -r <runtime> -d <device-ip>
# -> success + 7-day expiry notice; device applies without TUF errors
```

This PR also closes the ENG-2005 config-show contract: `config show --output json`
exposes `signing_enabled` per runtime, resolved identically to the deploy nudge
(per-runtime `signing.server_key`, falling back to `connect.server_key`).

[...]

PR #161, opened 2026-06-26
ext package: bake resolved version into the published avocado.yaml

mobileoverlord:
## Problem

The in-source program extensions (`avocado-ext-cli`, `avocado-ext-connect`, `avocado-ext-tunnels`) keep the extension version in sync with `Cargo.toml` by templating it from an env var:

```yaml
extensions:
  avocado-ext-connect:
    version: '{{ env.AVOCADO_EXT_VERSION }}'   # CI exports it from [package] version
```

That template only resolves while `AVOCADO_EXT_VERSION` is set — i.e. at **package / PR-test time**. The `ext package` RPM payload copied the **raw on-disk `avocado.yaml`**, so the literal template survived into the published package. Later, a **runtime build** consuming that extension has no `AVOCADO_EXT_VERSION` in scope, so it interpolated to `''` and failed:

```
Extension 'avocado-ext-connect' has invalid version ''. Version must be in
semantic versioning format (e.g., '1.0.0', '2.1.3')
```

## Fix

Resolve the required `version` field **at package time** and bake the concrete semver into the `avocado.yaml` that ships in the RPM payload. The resolved value is `metadata.version` — already interpolated through the normal config pipeline and semver-validated before packaging — so no new resolution logic, just persistence.

Only `extensions.<ext>.version` is rewritten, via a comment-preserving, indentation-scoped line edit (no serde round-trip, so comments and quoting are untouched). **Every other template** (`{{ avocado.target }}`, `{{ env.AVOCADO_DISTRO_RELEASE }}`, …) is intentionally left for the consumer to resolve at their build time — matching the "interpolate the required fields, defer the rest" model.

Implementation notes:
- The bake rewrites the single small staged `avocado.yaml` (≈1 KB) that's already in the staging dir — it adds **no** copy of the (potentially large) source tree.
- Remote/unreadable source configs (e.g. packaging a fetched remote ext) are left untouched rather than failing the package.

## Tests

New unit tests for the `bake_extension_version` helper:
- version-only replacement, with other templates **and** comments preserved
- nested `version:` keys (e.g. under `packages:`) left untouched
- multi-extension files bake only the named extension
- not-found error path

Local: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test` all green (17 package tests, incl. 4 new).

## Follow-up (not in this PR)

`ext package` currently copies the selected files twice (src → staging → rpmbuild buildroot). A separate refactor can have `%install` copy straight from the source into the buildroot (dropping the intermediate staging dir) to cut one copy for large source trees; it touches the rpmbuild flow so it should be validated against a real `avocado ext package` run.

### DIFF
--- a/src/commands/ext/package.rs
+++ b/src/commands/ext/package.rs
@@ -587,15 +587,14 @@ impl ExtPackageCommand {
                 .unwrap_or_else(|| "avocado.yaml".to_string());
             match fs::read_to_string(ext_config_path) {
                 Ok(text) => {
-                    let baked =
-                        bake_extension_version(&text, &self.extension, &metadata.version)
-                            .with_context(|| {
-                                format!(
-                                    "Failed to bake resolved version '{}' into the packaged \
+                    let baked = bake_extension_version(&text, &self.extension, &metadata.version)
+                        .with_context(|| {
+                        format!(
+                            "Failed to bake resolved version '{}' into the packaged \
                                      config for extension '{}'",
-                                    metadata.version, self.extension
-                                )
-                            })?;
+                            metadata.version, self.extension
+                        )
+                    })?;
                     let b64 = BASE64_STANDARD.encode(baked.as_bytes());
                     format!(
                         r#"
@@ -1027,9 +1026,7 @@ fn bake_extension_version(text: &str, ext_name: &str, version: &str) -> Result<S
     }
 
     if !replaced {
-        anyhow::bail!(
-            "could not locate `version:` for extension '{ext_name}' in its avocado.yaml"
-        );
+        anyhow::bail!("could not locate `version:` for extension '{ext_name}' in its avocado.yaml");
     }
 
     let mut result = out.join("\n");
--- a/src/commands/ext/package.rs
+++ b/src/commands/ext/package.rs
@@ -143,6 +143,14 @@ impl ExtPackageCommand {
             }
         };
 
+        // On-disk config filename (avocado.yaml or avocado.yml). Used both for the
+        // default packaged config and for the version bake so a `.yml` ext stages
+        // and bakes under its real name rather than a hardcoded `avocado.yaml`.
+        let cfg_basename = std::path::Path::new(&ext_config_path)
+            .file_name()
+            .map(|n| n.to_string_lossy().to_string())
+            .unwrap_or_else(|| "avocado.yaml".to_string());
+
         if self.verbose {
             match &extension_location {
                 ExtensionLocation::Local { name, config_path } => {
@@ -207,7 +215,8 @@ impl ExtPackageCommand {
         // Determine which files to package
         // Pass both merged config (for package_files), raw config (for all target overlays),
         // and full parsed config (for sdk.compile scripts)
-        let package_files = self.get_package_files(&ext_config, raw_ext_config.as_ref(), parsed);
+        let package_files =
+            self.get_package_files(&ext_config, raw_ext_config.as_ref(), parsed, &cfg_basename);
 
         if self.verbose {
             print_info(
@@ -232,6 +241,7 @@ impl ExtPackageCommand {
                 &target,
                 &ext_config_path,
                 &package_files,
+                &cfg_basename,
             )
             .await?;
 
@@ -306,6 +316,7 @@ impl ExtPackageCommand {
         ext_config: &serde_yaml::Value,
         raw_ext_config: Option<&serde_yaml::Value>,
         full_parsed_config: &serde_yaml::Value,
+        cfg_basename: &str,
     ) -> Vec<String> {
         // Check if package_files is explicitly defined
         if let Some(package_files) = ext_config.get("package_files") {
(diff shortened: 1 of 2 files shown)


jetm:
@mobileoverlord before the review below: I've now gone through every PR you asked me to look at, so I'd appreciate the same in return. Mine have been sitting without a review and they're the foundation the CI/CD lab work depends on. Fair's fair - here are mine, roughly in priority order:

- avocado-cli #158: https://github.com/avocado-linux/avocado-cli/pull/158
- meta-avocado #232: https://github.com/avocado-linux/meta-avocado/pull/232
- meta-avocado #229: https://github.com/avocado-linux/meta-avocado/pull/229
- meta-avocado #230: https://github.com/avocado-linux/meta-avocado/pull/230
- meta-avocado #227: https://github.com/avocado-linux/meta-avocado/pull/227
- meta-avocado #226: https://github.com/avocado-linux/meta-avocado/pull/226
- avocado-mcp #8: https://github.com/avocado-linux/avocado-mcp/pull/8

Now to #161. The direction is right and the base64 transport is a clean way to dodge shell-escaping the YAML, but I think a few things need addressing before this merges - two of them turn a currently-working `ext package` into a hard failure, and one is a command-injection path.

**Critical**

1. **Version sourced from a target/kernel override block hard-fails.** `metadata.version` comes from the *merged* config, so the resolved version can originate from a `target-<name>.version` sub-block. `bake_extension_version` only matches a `version:` that's a direct child of the extension, so in that case it `bail!`s and the `?` aborts a package that built fine before. It can also half-bake: a base `version:` template gets resolved while a `target-*.version` template survives into the payload and still interpolates to `''` downstream - the exact failure this is meant to prevent.

2. **Command injection through the version string.** `validate_semver` only checks the three numeric components; pre-release/build metadata after `-`/`+` isn't validated. `metadata.version` is then interpolated into the double-quoted `echo "Baked resolved version '...'"` in the rpmbuild script, so a version like `1.0.0-$(...)` executes in the build container. The baked YAML rides safely through base64 already; the `version` shouldn't be interpolated into the shell at all.

3. **Templated extension keys hard-fail.** Extension keys can be templates (`avocado-bsp-{{ avocado.target }}`), and `find_ext_in_mapping` resolves them by interpolating. `bake_extension_version` compares the key literally, so any template-keyed extension never matches and `bail!`s.

**Warning**

4. **`.yml` vs `.yaml` silently skips the bake.** `cfg_basename` follows the on-disk name, but staging copies the hardcoded `avocado.yaml`. If the config is `avocado.yml`, the `if [ -f "$STAGING_DIR/avocado.yml" ]` guard is false and the bake is skipped with no error, shipping the unresolved template.

5. **Remote extensions never get baked.** The host can't read the container-volume config path, so the bake section is empty and the template ships. It's noted as intentional, but it's the same downstream breakage, just for remote ext.

**Suggestion**

6. **Reuse `src/utils/config_edit.rs`.** We already have a comment-preserving avocado.yaml line-editor there, with quoted-key handling and the compact-indent fix from ENG-1868, plus tests. A `set_extension_field` built on `find_top_level_key` would replace the hand-rolled state machine and inherit those fixes instead of drifting from them. Resolving against the merged config (which fixes 1 and 3) fits naturally there too.

The tests are good - the nested-`version:` and multi-extension cases are exactly the right edges. Happy to pair on the `config_edit.rs` rework if useful.

mobileoverlord:
Thanks @jetm — this was a thorough review and most of it landed. Pushed f8b8d78 addressing it. Summary of what changed and where I pushed back:

**Root cause (your #1 + #3):** the bake matched the raw on-disk text literally while `metadata.version` is resolved from the *merged* config (`find_ext_in_mapping` + target overrides). Fixed at the source — matching now mirrors `find_ext_in_mapping` (direct key, then `{{ avocado.target }}` interpolation), so:

- **#3 templated keys** (`avocado-bsp-{{ avocado.target }}`) resolve instead of `bail!`ing. This was the widest regression, so good catch.
- **#1 target-block version**: when there's no direct-child `version:` we now *insert* a concrete base version instead of aborting, and we also bake `version:` inside `target-*`/`kernel-*` override blocks so a surviving template can't interpolate to `''` downstream. Deeper `packages.*.version` stays untouched.

**#2 command injection:** fixed at the injection point — the version string is no longer interpolated into the rpmbuild `echo`; the baked value rides the base64 payload only. I'd gently downgrade this from "critical": the baked YAML was already base64-safe, only the cosmetic log line was exposed, and `version` originates from CI's `AVOCADO_EXT_VERSION`/Cargo.toml rather than an untrusted party. Real defense-in-depth, worth fixing because it's free, but not a live exploit in the intended flow. (Left `validate_semver` alone — tightening it would reject legitimate pre-release/build metadata; removing the shell interpolation is the right layer.)

**#4 `.yml`/`.yaml`:** the on-disk basename now drives both staging and the bake guard, so a `.yml` ext stages and bakes under its real name.

**#6 reuse `config_edit.rs`:** done — the helper moved there next to the other comment-preserving editors, reusing `find_top_level_key`/`extract_package_name`/`leading_spaces`. Tests moved with it and gained template-key + target-block coverage. Resolving against the merged config (your #1/#3 fix) fits naturally there as you predicted.

**#5 remote ext:** left as-is/documented — the host can't read the container-volume config, and the empty-bake path avoids failing the package. Same downstream caveat as before, no regression. Open to revisiting if it bites in practice.

One note on #1: no config in the tree currently puts `version` in a target block, so the half-bake was theoretical — but I implemented the override-block baking anyway since it's cheap and makes the helper match the merge semantics rather than drift from them.

`cargo fmt` / `clippy -D warnings` / `cargo test` green (1070 tests; 7 bake tests incl. the new template-keyed and target-block cases).

And yes — pulling up your stack now, will get reviews back to you today.

jetm (approved):
@mobileoverlord Approving. Appreciate the quick turnaround and picking up the stack.

PR #165, opened 2026-07-03
lockfile: relocate to top-level avocado.lock

mobileoverlord:
## Summary

Promotes the lock file from `.avocado/lock.json` to a top-level, committed
`avocado.lock` (like `Cargo.lock` / `flake.lock`). This frees `.avocado/` to be
a purely-scratch, gitignored directory.

## Details

- `LockFile::get_path` now returns `<src>/avocado.lock`; a new `legacy_path()`
  points at the old `.avocado/lock.json`.
- `LockFile::load` migrates an existing legacy lock to the new path on first
  read (version-migrate → write `avocado.lock` → remove the old file), so
  existing projects upgrade transparently.
- `save` merges any legacy on-disk state before writing.
- connect upload / save / load resolve the lock via `get_path` with a legacy
  fallback, and the save/load bundle now carries `avocado.lock`.
- `init` gitignores the whole `.avocado/` scratch dir while keeping
  `avocado.lock` tracked.

## Testing

- Added a legacy → top-level migration test; updated the path assertion.

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -848,26 +848,45 @@ impl InitCommand {
                 format!("Failed to read existing '{}'", gitignore_path.display())
             })?;
 
-            // Check if .avocado-state is already in the .gitignore
-            if !existing_content.contains(".avocado-state") {
-                // Append to existing .gitignore
+            // Ensure both the legacy state file and the `.avocado/` scratch dir
+            // are ignored. `avocado.lock` is intentionally NOT ignored — it's a
+            // tracked, committed pin file (like Cargo.lock).
+            //
+            // Match whole lines (trimmed), not substrings: a substring test would
+            // treat a pre-existing `.avocado/lock.json` line as already covering
+            // `.avocado/` (so the broad scratch ignore is never added), and a bare
+            // `.avocado` entry would both fail a `.avocado/` substring test and be
+            // double-appended.
+            let existing_lines: std::collections::HashSet<&str> =
+                existing_content.lines().map(str::trim).collect();
+            let mut to_add = String::new();
+            if !existing_lines.contains(".avocado-state") {
+                to_add.push_str(".avocado-state\n");
+            }
+            if !existing_lines.contains(".avocado/") {
+                to_add.push_str(".avocado/\n");
+            }
+            if !to_add.is_empty() {
                 let mut updated_content = existing_content;
                 if !updated_content.ends_with('\n') {
                     updated_content.push('\n');
                 }
-                updated_content.push_str("\n# Avocado state files\n.avocado-state\n");
+                updated_content.push_str("\n# Avocado scratch + state (avocado.lock is tracked)\n");
+                updated_content.push_str(&to_add);
 
                 fs::write(&gitignore_path, updated_content)
                     .with_context(|| format!("Failed to update '{}'", gitignore_path.display()))?;
 
-                self.emit_info("✓ Updated .gitignore to ignore .avocado-state files.");
+                self.emit_info("✓ Updated .gitignore to ignore Avocado scratch files.");
             }
 
             return Ok(());
         }
 
-        // Create new .gitignore with Avocado-specific entries
-        let gitignore_content = "# Avocado state files\n.avocado-state\n";
+        // Create new .gitignore with Avocado-specific entries. `avocado.lock`
+        // is deliberately tracked (committed), so it is not listed here.
+        let gitignore_content =
+            "# Avocado scratch + state (avocado.lock is tracked)\n.avocado-state\n.avocado/\n";
 
         fs::write(&gitignore_path, gitignore_content).with_context(|| {
             format!(
--- a/src/commands/load.rs
+++ b/src/commands/load.rs
@@ -8,6 +8,7 @@ use std::io::Read;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
+use crate::utils::lockfile::LockFile;
 use crate::utils::output::{print_info, print_success, print_warning, OutputLevel};
 use crate::utils::volume::VolumeState;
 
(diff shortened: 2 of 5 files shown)


src/utils/lockfile.rs
```diff
@@ -500,20 +512,53 @@ impl LockFile {
...
+            let _ = fs::remove_file(&legacy_path);
+            crate::utils::output::print_info(
+                &format!(
+                    "Migrated lock file '{}' -> '{}'",
+                    legacy_path.display(),
+                    path.display()
+                ),
+                crate::utils::output::OutputLevel::Normal,
+            );
         }
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: the legacy → `avocado.lock` migration write is now guarded by `LOCKFILE_SAVE_GATE.try_lock()` (not `lock()`, since `load()` runs inside `save()` while the gate is held — a blocking lock would deadlock). If the gate is busy the eager migration is skipped (the in-flight `save()` persists `avocado.lock` anyway, and a later standalone `load()` finishes dropping the legacy file). Legacy-file removal failures now surface as a warning instead of being ignored.

src/utils/lockfile.rs
```diff
@@ -799,7 +844,10 @@ impl LockFile {
         // top so concurrent writers' updates accumulate instead of
         // clobbering each other.
         let path = Self::get_path(src_dir);
-        let to_write = if path.exists() {
+        // Merge with whatever is on disk, including a not-yet-migrated legacy
+        // `.avocado/lock.json` (load() handles relocating it), so a save that
+        // happens before any explicit load doesn't clobber existing pins.
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: reworded the comment to state that `load()` relocates the legacy file only when `avocado.lock` is absent, and that an existing `avocado.lock` takes precedence (the legacy file is ignored, not merged).

src/commands/save.rs
```diff
@@ -211,6 +211,12 @@ fn build_save_archive(
         tar.append_dir_all(format!("{ARCHIVE_PREFIX}/config/.avocado"), &avocado_dir)?;
     }
 
+    // 3b. Add the top-level avocado.lock (relocated out of .avocado/) if present.
+    let lockfile = config_dir.join("avocado.lock");
+    if lockfile.is_file() {
+        tar.append_path_with_name(&lockfile, format!("{ARCHIVE_PREFIX}/config/avocado.lock"))?;
+    }
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: now uses `LockFile::get_path(config_dir)` instead of hardcoding the lockfile name/path.

src/commands/load.rs
```diff
@@ -207,6 +207,15 @@ fn import_archive(
...
     }
 
+    // Restore the top-level avocado.lock (newer bundles; older bundles carry it
+    // inside the restored .avocado/ as lock.json and migrate on first load).
+    let archive_lockfile = archive_config_dir.join("avocado.lock");
+    if archive_lockfile.is_file() {
+        let dest_lockfile = config_dir.join("avocado.lock");
+        fs::copy(&archive_lockfile, &dest_lockfile)
+            .with_context(|| "Failed to restore avocado.lock")?;
+    }
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: the restore destination now uses `LockFile::get_path(config_dir)` so it stays in sync with the canonical definition.

src/commands/load.rs
```diff
@@ -207,6 +207,15 @@ fn import_archive(
             .with_context(|| "Failed to restore .avocado/ directory")?;
     }
 
+    // Restore the top-level avocado.lock (newer bundles; older bundles carry it
+    // inside the restored .avocado/ as lock.json and migrate on first load).
+    let archive_lockfile = archive_config_dir.join("avocado.lock");
+    if archive_lockfile.is_file() {
```
jetm: The `avocado.lock` restore is unconditional — only an `is_file()` check, with none of the `config_path.exists() && !force` guarding that protects `avocado.yaml` just above. A non-`--force` `avocado load` silently overwrites the user's local lock pins while reporting nothing changed. The same block also means an old-format bundle (lock only in `.avocado/lock.json`, no top-level file) has its pins silently ignored when the project already has a local `avocado.lock`, since `load()` prefers the top-level path and never reads the restored legacy file. Suggest guarding the copy with the same `!force` check and handling the legacy-only bundle.

src/utils/lockfile.rs
```diff
@@ -500,20 +510,70 @@ impl LockFile {
...
+        // top-level `avocado.lock`, then drop the old file. The version is
+        // already current after `parse_and_migrate`, so this just moves it.
+        //
+        // Guard the write with the save gate (via `try_lock`, not `lock`) so it
+        // can't race a concurrent `save()`. `try_lock` is deliberate: `load()`
+        // is itself called from within `save()` while the gate is held, and a
+        // blocking lock would deadlock. If the gate is busy we skip the eager
+        // migration — the in-flight `save()` persists `avocado.lock` anyway, and
+        // a later standalone `load()` will finish dropping the legacy file.
+        if from_legacy {
```
jetm: The comment promises "a later standalone `load()` will finish dropping the legacy file," but that can't happen on a save-first flow: `save()` holds `LOCKFILE_SAVE_GATE`, then calls `load()` whose `try_lock` fails (non-reentrant), so migration is skipped and `save()` writes `avocado.lock`. Every later `load()` then takes the `path.exists()` branch and never revisits `legacy_path`, so `.avocado/lock.json` is orphaned permanently.

src/commands/save.rs
```diff
@@ -211,6 +211,12 @@ fn build_save_archive(
         tar.append_dir_all(format!("{ARCHIVE_PREFIX}/config/.avocado"), &avocado_dir)?;
     }
 
+    // 3b. Add the top-level avocado.lock (relocated out of .avocado/) if present.
+    let lockfile = crate::utils::lockfile::LockFile::get_path(config_dir);
+    if lockfile.is_file() {
+        tar.append_path_with_name(&lockfile, format!("{ARCHIVE_PREFIX}/config/avocado.lock"))?;
```
jetm, line 230: If the legacy `.avocado/lock.json` was orphaned (see the migration note in `lockfile.rs`), the `append_dir_all` of `.avocado` above already bundled it, and this line adds the top-level `avocado.lock` too — so the archive ships two lock files that can disagree, and which one wins on restore depends on ordering. Consider not packing the legacy file, or migrating it before archiving.

src/commands/init.rs
```diff
@@ -848,26 +848,37 @@ impl InitCommand {
...
-            // Check if .avocado-state is already in the .gitignore
+            // Ensure both the legacy state file and the `.avocado/` scratch dir
+            // are ignored. `avocado.lock` is intentionally NOT ignored — it's a
+            // tracked, committed pin file (like Cargo.lock).
+            let mut to_add = String::new();
             if !existing_content.contains(".avocado-state") {
-                // Append to existing .gitignore
+                to_add.push_str(".avocado-state\n");
+            }
+            if !existing_content.contains(".avocado/") {
```
jetm: `existing_content.contains(".avocado/")` is a substring test: a pre-existing specific entry like `.avocado/lock.json` satisfies it, so the broad `.avocado/` scratch ignore is never added and other scratch under `.avocado/` stays committable — the opposite of this PR's goal. A bare `.avocado` entry (no slash) also fails the check and double-appends. Matching line-wise (split, trim, compare) is more robust.

jetm (commented):
Reviewed the lockfile relocation. The migration/restore path has a few correctness issues, plus a `.gitignore` matching nit — details inline.

src/commands/load.rs
mobileoverlord: Fixed: the lock restore is now guarded by the same `!force` check as avocado.yaml (won't clobber a local `avocado.lock` without `--force`). It also falls back to a legacy `.avocado/lock.json` in the archive — normalizing it to the top-level `avocado.lock` — so an older legacy-only bundle's pins aren't ignored, and drops any restored `.avocado/lock.json` afterward so the two can't diverge.

src/utils/lockfile.rs
mobileoverlord: Fixed: `load()` is now read-only — no write/relocate — so the try_lock skip that orphaned the legacy file is gone. Physical migration (write `avocado.lock` + remove `.avocado/lock.json`) moved into `write_to_disk`, which runs on every `save()` under the gate. The save-first flow now migrates instead of orphaning; the misleading comment is removed.

src/commands/save.rs
mobileoverlord: Fixed: `save` migrates any legacy `.avocado/lock.json` to the top-level `avocado.lock` before archiving (load + save_replacing, which drops the legacy file), so the bundle ships exactly one canonical lock and the `.avocado/` archive no longer also carries a legacy lock.

src/commands/init.rs
mobileoverlord: Fixed: the check is now line-wise (split + trim + exact match) rather than a substring test — a pre-existing `.avocado/lock.json` line no longer masks adding the broad `.avocado/` ignore, and a bare `.avocado` entry isn't double-appended.

jetm (approved):
All review findings addressed — the lock restore now honors `!force`, legacy bundles fall back and are cleaned up, the migration is read-only in `load()`, and `.gitignore` matching is line-wise. CI green. LGTM.

PR #166, opened 2026-07-03
overlay: opt-in build-time preprocessing of overlay files

mobileoverlord:
## Summary

Adds opt-in `{{ … }}` preprocessing of overlay file *contents* at build time,
so values (e.g. a secret such as an API/claim token) can be templated into
overlay files without mutating the working tree.

## Usage

```yaml
overlay:
  dir: overlay
  preprocess:                 # opt-in; default off = byte-identical to today
    - "etc/some/config.toml"  # globs, or `preprocess: true` for all UTF-8 files
```

The committed overlay file holds e.g. `token = "{{ env.SOME_TOKEN }}"`; the
value is substituted only into the built image, never back into the tree.

## Details

- Reuses the existing `env.` / `config.` / `avocado.` interpolation engine via a
  new `preprocess_text` (no new templating language).
- Overlay files are copied verbatim inside the SDK container, so a processed
  copy is materialized on the host into `.avocado/overlay-staging/<label>/`
  (gitignored scratch, inside the `/opt/src` bind mount) and the copy is pointed
  there.
- Binary (non-UTF-8) files pass through verbatim; symlinks and mode bits are
  preserved.
- Wired into the rootfs/initramfs and sysext/confext overlay copy sites. Local
  overlays only; remote-extension overlays warn + skip.
- Stamps fold a post-preprocess content digest into the rootfs/initramfs/ext
  input hashes when preprocessing is enabled, so a changed template value or an
  edited overlay file invalidates the build. Verbatim overlays are unchanged.

## Testing

- Unit tests for `preprocess_text`, glob matching, binary passthrough, and stamp
  invalidation on a changed template value.

### DIFF
--- a/src/commands/ext/build.rs
+++ b/src/commands/ext/build.rs
@@ -21,6 +21,8 @@ use crate::utils::tui::{TaskId, TuiGuard};
 struct OverlayConfig {
     dir: String,
     mode: OverlayMode,
+    /// Opt-in `{{ }}` preprocessing of overlay file contents.
+    preprocess: crate::utils::overlay_preprocess::PreprocessSpec,
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -292,8 +294,14 @@ impl ExtBuildCommand {
             // is matched against the entry for its (component, command) pair.
             let project_root = config.project_root(&self.config_path);
             let install_inputs = compute_ext_install_input_hash(parsed, &self.extension).ok();
-            let build_inputs =
-                compute_ext_build_input_hash(parsed, &self.extension, &project_root).ok();
+            let build_inputs = compute_ext_build_input_hash(
+                parsed,
+                &self.extension,
+                &project_root,
+                Some(target.as_str()),
+                self.runtime.as_deref(),
+            )
+            .ok();
             let mut current_inputs: Vec<crate::utils::stamps::CurrentInput<'_>> = Vec::new();
             if let Some(ref i) = install_inputs {
                 current_inputs.push((StampComponent::Extension, StampCommand::Install, i));
@@ -554,12 +562,14 @@ impl ExtBuildCommand {
             })?;
 
         // Get overlay configuration
-        let overlay_config = ext_config.get("overlay").map(|v| {
+        let mut overlay_config = ext_config.get("overlay").map(|v| {
+            use crate::utils::overlay_preprocess::PreprocessSpec;
             if let Some(dir_str) = v.as_str() {
                 // Simple string format: overlay = "directory"
                 OverlayConfig {
                     dir: dir_str.to_string(),
                     mode: OverlayMode::Merge, // Default to merge mode
+                    preprocess: PreprocessSpec::None,
                 }
             } else if let Some(table) = v.as_mapping() {
                 // Table format: overlay = {dir = "directory", mode = "opaque"}
@@ -574,12 +584,17 @@ impl ExtBuildCommand {
                     _ => OverlayMode::Merge, // Default to merge mode
                 };
 
-                OverlayConfig { dir, mode }
+                OverlayConfig {
+                    dir,
+                    mode,
+                    preprocess: PreprocessSpec::from_overlay_value(v),
+                }
             } else {
                 // Fallback for invalid format
                 OverlayConfig {
                     dir: "overlay".to_string(),
                     mode: OverlayMode::Merge,
+                    preprocess: PreprocessSpec::None,
                 }
             }
         });
@@ -615,6 +630,65 @@ impl ExtBuildCommand {
             }
             ExtensionLocation::Local { .. } => "/opt/src".to_string(),
         };
+
+        // Opt-in overlay preprocessing (local extensions only). Materialize an
+        // interpolated copy of the overlay on the host under
+        // `.avocado/overlay-staging/` and redirect the in-container copy to it,
+        // so `{{ ... }}` in overlay files is substituted without mutating the
+        // working tree. Remote-extension overlays live in the SDK volume (not on
+        // the host), so they can't be preprocessed here — warn and skip.
+        if let Some(oc) = overlay_config.as_mut() {
+            if oc.preprocess.is_enabled() {
+                match &extension_location {
+                    ExtensionLocation::Local { .. } => {
+                        let project_root = config.project_root(&self.config_path);
+                        let mut context =
+                            crate::utils::interpolation::AvocadoContext::from_main_config(
+                                parsed,
+                                Some(target.as_str()),
+                            );
+                        // The selected runtime drives `{{ avocado.runtime }}` and
+                        // the runtime-scoped target board; from_main_config only
+                        // sees env/default, so override it when one was chosen.
+                        if let Some(rt) = self.runtime.as_deref() {
+                            context.runtime = Some(rt.to_string());
+                        }
+                        let label = format!(
+                            "ext-{}",
+                            self.extension
+                                .chars()
+                                .map(|c| if c.is_alphanumeric() || c == '-' || c == '_' {
+                                    c
+                                } else {
+                                    '_'
+                                })
+                                .collect::<String>()
+                        );
+                        if let Some(staging_rel_dir) =
+                            crate::utils::overlay_preprocess::materialize_preprocessed_overlay(
+                                &project_root,
+                                &oc.dir,
+                                &label,
+                                &oc.preprocess,
+                                parsed,
+                                &context,
+                            )?
+                        {
+                            oc.dir = staging_rel_dir;
+                        }
+                    }
+                    ExtensionLocation::Remote { name, .. } => {
+                        print_warning(
+                            &format!(
+                                "Overlay preprocessing is not supported for remote extension \
+                                 '{name}'; copying overlay verbatim."
+                            ),
+                            OutputLevel::Normal,
+                        );
+                    }
+                }
+            }
+        }
 
         // Run the post_build hook before sealing any .raw images. By this
         // point the ext sysroot has been populated by `avocado install` and
@@ -736,6 +810,8 @@ impl ExtBuildCommand {
                 parsed,
                 &self.extension,
                 &config.project_root(&self.config_path),
+                Some(target.as_str()),
+                self.runtime.as_deref(),
             )?;
             let outputs = StampOutputs::default();
             let stamp = Stamp::ext_build(&self.extension, &target, inputs, outputs);
@@ -1913,6 +1989,7 @@ mod tests {
         let overlay_config = OverlayConfig {
             dir: "peridio".to_string(),
             mode: OverlayMode::Merge,
+            preprocess: crate::utils::overlay_preprocess::PreprocessSpec::None,
         };
         let script = cmd.create_sysext_build_script(
             "1.0",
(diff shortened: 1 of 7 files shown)


src/utils/overlay_preprocess.rs
```diff
@@ -0,0 +1,421 @@
...
+        if ft.is_symlink() {
+            let target = std::fs::read_link(entry.path())
+                .map(|p| p.to_string_lossy().into_owned())
+                .unwrap_or_default();
+            hasher.update(format!("L\0{rel}\0{target}\0").as_bytes());
+        } else if ft.is_file() {
+            let raw = std::fs::read(entry.path()).ok()?;
+            let mode = file_mode(entry.path());
+            let bytes = process_file_bytes(&rel, raw, spec, root, context).ok()?;
+            let file_hash = Sha256::digest(&bytes);
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: `overlay_content_digest` now returns `Result<Option<String>>` and propagates read/preprocess errors (through `fold_overlay_content_hash` and the `compute_*_input_hash` functions, which already return `Result`). A broken overlay now fails loudly instead of silently dropping the content digest and skipping rebuild invalidation.

src/utils/overlay_preprocess.rs
```diff
@@ -0,0 +1,421 @@
...
+        return Ok(None);
+    }
+    let overlay_src = project_root.join(overlay_rel_dir);
+    if !overlay_src.is_dir() {
+        // Missing overlay dir is reported by the copy script itself; nothing to stage.
+        return Ok(None);
+    }
+
+    let rel_dir = format!("{STAGING_SUBDIR}/{label}");
+    let staging_dir = project_root.join(&rel_dir);
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: `materialize_preprocessed_overlay` now sanitizes `label` to a single safe path component (any char outside `[A-Za-z0-9_-]` becomes `_`) before it is joined into the staging path and used with `remove_dir_all`, so a label containing separators, `..`, or an absolute path cannot escape the staging root.

src/utils/overlay_preprocess.rs
```diff
@@ -0,0 +1,421 @@
...
+        let dest = staging_dir.join(&rel);
+        let ft = entry.file_type();
+
+        if ft.is_dir() {
+            std::fs::create_dir_all(&dest)
+                .with_context(|| format!("Failed to create staging dir: {}", dest.display()))?;
+        } else if ft.is_symlink() {
+            if let Some(parent) = dest.parent() {
+                std::fs::create_dir_all(parent).ok();
+            }
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: the parent `create_dir_all` before symlink creation now propagates the error with context instead of `.ok()`.

src/utils/overlay_preprocess.rs
```diff
@@ -0,0 +1,421 @@
...
+            let target = std::fs::read_link(entry.path()).with_context(|| {
+                format!("Failed to read overlay symlink: {}", entry.path().display())
+            })?;
+            recreate_symlink(&target, &dest)?;
+        } else if ft.is_file() {
+            if let Some(parent) = dest.parent() {
+                std::fs::create_dir_all(parent).ok();
+            }
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: the parent `create_dir_all` before the file write now propagates the error with context instead of `.ok()`.

src/utils/stamps.rs
```diff
@@ -892,6 +892,44 @@ fn script_hash_value(project_root: &Path, rel_path: &str) -> serde_yaml::Value {
...
+    if !spec.is_enabled() {
+        return;
+    }
+    let dir = overlay
+        .get("dir")
+        .and_then(|d| d.as_str())
+        .unwrap_or("overlay");
+    let context = crate::utils::interpolation::AvocadoContext::from_main_config(config, None);
+    if let Some(digest) = crate::utils::overlay_preprocess::overlay_content_digest(
+        project_root,
```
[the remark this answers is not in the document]
mobileoverlord: The digest uses `from_main_config(config, None)`, which resolves target from env/config. This is consistent between stamp-write and stamp-check (same function both times), and the stamp is already keyed per target (`Stamp::rootfs_install(target, …)`), so a different `--target` yields a different stamp key — it cannot cause a false cache hit or miss. `env.*` values are captured regardless of target. Threading the CLI target here would be a no-op for cache correctness and would ripple through `compute_ext_image_input_hash` and many call sites, so I've left it and documented the rationale in a comment. Happy to thread it if you'd prefer the belt-and-suspenders version.

src/utils/overlay_preprocess.rs
```diff
@@ -0,0 +1,480 @@
...
+                    PreprocessSpec::Globs(globs)
+                }
+            }
+            _ => PreprocessSpec::None,
+        }
+    }
+
+    /// Whether the file at `rel_path` (overlay-relative, forward slashes) should
+    /// be preprocessed.
+    fn matches(&self, rel_path: &str) -> bool {
```
jetm, line 71: `All` returns true for every file with no per-file skip, and `preprocess_text` errors on any unresolved `{{ ... }}` (unknown template context / missing config path). Because `overlay_content_digest` feeds `compute_*_input_hash`, which runs on every build to decide whether to rebuild, a single overlay file with a literal `{{ }}` (a systemd unit, a Go/Jinja/Handlebars template, some JSON) aborts the whole build before anything is built — with no way to exclude that file. Consider scoping preprocessing to a glob rather than defaulting to the entire tree, or making hash-time interpolation errors non-fatal.

src/utils/overlay_preprocess.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    for entry in sorted_entries(&overlay_src) {
+        let Some(rel) = rel_str(&overlay_src, entry.path()) else {
+            continue;
+        };
+        let dest = staging_dir.join(&rel);
+        let ft = entry.file_type();
+
+        if ft.is_dir() {
+            std::fs::create_dir_all(&dest)
```
jetm, line 315: Directories are recreated with bare `create_dir_all` (process umask); only regular files get `copy_mode`. The verbatim path uses `cp -a`, which preserves directory modes, so a preprocessed overlay silently widens e.g. a `0700 .ssh` to `0755`. `overlay_content_digest` also never folds directory modes into the hash, so the cache can't detect the drift. Set the dir mode from the source and include it in the digest.

src/commands/ext/build.rs
```diff
@@ -616,6 +625,58 @@ impl ExtBuildCommand {
...
+        // `.avocado/overlay-staging/` and redirect the in-container copy to it,
+        // so `{{ ... }}` in overlay files is substituted without mutating the
+        // working tree. Remote-extension overlays live in the SDK volume (not on
+        // the host), so they can't be preprocessed here — warn and skip.
+        if let Some(oc) = overlay_config.as_mut() {
+            if oc.preprocess.is_enabled() {
+                match &extension_location {
+                    ExtensionLocation::Local { .. } => {
+                        let project_root = config.project_root(&self.config_path);
+                        let context = crate::utils::interpolation::AvocadoContext::from_main_config(
```
jetm: `from_main_config(parsed, Some(target))` never receives the CLI-selected runtime (`self.runtime`), so `{{ avocado.runtime }}` and the target board render the config default regardless of which runtime the user selected. Since the stamp key and digest don't reflect the selection either, switching runtimes reuses a stale artifact. Thread the selected runtime into both the interpolation context and the stamp key.

src/utils/overlay_preprocess.rs
```diff
@@ -0,0 +1,480 @@
...
+    entries
+}
+
+fn rel_str(overlay_src: &Path, path: &Path) -> Option<String> {
+    let rel = path.strip_prefix(overlay_src).ok()?;
+    if rel.as_os_str().is_empty() {
+        return None;
+    }
+    // Forward-slash, lossy is fine (overlay paths are ASCII in practice).
+    Some(rel.to_string_lossy().replace('\\', "/"))
```
jetm: `to_string_lossy` is used for both the staging destination path and the digest key, so a non-UTF-8 filename is mangled to U+FFFD — the staged file lands under a corrupted name and two names differing only in non-UTF-8 bytes collide (and hash equal), whereas the verbatim `cp -a` preserves exact bytes. Low severity (the inline comment concedes "ASCII in practice"), but worth a guard or an explicit reject.

src/utils/overlay_preprocess.rs
```diff
@@ -0,0 +1,480 @@
...
+    use sha2::{Digest, Sha256};
+    let mut hasher = Sha256::new();
+    for entry in sorted_entries(&overlay_src) {
+        let Some(rel) = rel_str(&overlay_src, entry.path()) else {
+            continue;
+        };
+        let ft = entry.file_type();
+        if ft.is_symlink() {
+            let target = std::fs::read_link(entry.path())
+                .map(|p| p.to_string_lossy().into_owned())
```
jetm: The digest reads the symlink target with `unwrap_or_default()` (folds empty on failure) while `materialize` treats the same `read_link` failure as a hard error, so the stamp check and the build can disagree (stamp passes, build aborts) and two unreadable links hash equal. Narrow state, but the two paths should handle read failure the same way.

jetm (commented):
Reviewed the overlay preprocessing. One build-aborting footgun and a directory-permission gap stand out; details inline.

src/utils/overlay_preprocess.rs
mobileoverlord: Fixed: overlay preprocessing now runs in lenient mode — a `{{ ... }}` whose context isn't ours (Go/Jinja/Helm, e.g. `{{ .Values.x }}`/`{{ range }}`) is left untouched instead of aborting the build; only env/config/avocado templates are substituted. Genuine IO errors still propagate. Added a test covering a foreign-template file under `preprocess: true`.

src/utils/overlay_preprocess.rs
mobileoverlord: Fixed: staged directories now get their source mode via `copy_mode` (was a bare `create_dir_all` under the umask), and the digest folds directory modes (`D\0<rel>\0<mode>`), so a 0700 -> 0755 drift is both preserved on stage and invalidates the stamp.

src/commands/ext/build.rs
mobileoverlord: Fixed: the materialize context now takes the selected runtime (`self.runtime`), so `{{ avocado.runtime }}`/board render for the chosen runtime; and the ext build/image input hashes thread target+runtime into the overlay digest, so switching runtimes invalidates the (otherwise not runtime-keyed) ext-build stamp. Added a test. rootfs/initramfs overlays don't vary by runtime, so those paths are unchanged.

src/utils/overlay_preprocess.rs
mobileoverlord: Fixed: `rel_str` now rejects a non-UTF-8 path with a clear error instead of `to_string_lossy` mangling it to U+FFFD (which could collide / hash equal). Applies consistently to staging and the digest.

src/utils/overlay_preprocess.rs
mobileoverlord: Fixed: the digest now propagates `read_link` failures with context (was `unwrap_or_default`), matching `materialize` — the stamp check and the build now handle an unreadable symlink identically.

jetm (approved):
All findings addressed — lenient interpolation lets foreign `{{ }}` pass through, directory modes are preserved and folded into the digest, the selected runtime flows into the context and stamp hashes, and non-UTF-8 paths hard-error. CI green. LGTM.

PR #167, opened 2026-07-03
ext: derive `type: path` mounts from config, drop ext-paths.json

mobileoverlord:
## Summary

Makes `avocado.yaml` the single source of truth for `type: path` extension
mounts. Previously `avocado ext fetch` recorded resolved mounts in
`.avocado/ext-paths.json`, which was read both at container launch (for bindfs
mounts) and during config composition. That side-table could drift from the
config whenever an extension's `source` flipped between `package` and `path` —
a stale entry kept mounting a host path over an extension that was now a
package.

## Details

- `SdkContainer::derive_ext_path_mounts` reads + interpolates `avocado.yaml` and
  mounts every `source: { type: path }` extension (replaces the state-file read
  at container launch).
- Config composition resolves a path-based extension's config directly from its
  declared `source.path` instead of the state file.
- Removes `ExtensionPathState` entirely; `ext fetch` for a path extension is now
  validation-only (nothing is written).

Flipping an extension between `package` and `path` now takes effect immediately
with nothing to reconcile.

## Follow-up (out of scope)

A separate change will clear a previously package-populated `includes/<ext>`
when switching it to `path`, so bindfs mounts onto a clean directory.

## Testing

- Added tests for deriving mounts from config (a `path` source is mounted, a
  `package` source is ignored; no path sources → no mounts).

### DIFF
--- a/src/utils/config.rs
+++ b/src/utils/config.rs
@@ -1911,13 +1911,23 @@ impl Config {
             );
         }
 
-        // Load extension path state for path-based extensions
-        let ext_path_state = crate::utils::ext_fetch::ExtensionPathState::load_from_dir(&src_dir)
-            .ok()
-            .flatten();
-
         // For each remote extension, try to read its config
         for (ext_name, source) in remote_extensions {
+            // For a `type: path` source, resolve its host dir directly from the
+            // config declaration (relative to src_dir) — the config is the
+            // source of truth, so no separate state file is consulted.
+            let path_source_dir: Option<PathBuf> = match &source {
+                ExtensionSource::Path { path, .. } => {
+                    let p = if Path::new(path).is_absolute() {
+                        PathBuf::from(path)
+                    } else {
+                        src_dir.join(path)
+                    };
+                    Some(p.canonicalize().unwrap_or(p))
+                }
+                _ => None,
+            };
+
             // Try multiple methods to read the extension config:
             // 0. Path-based extension: read directly from source path (for source: { type: path })
             // 1. Direct container path (when running inside a container)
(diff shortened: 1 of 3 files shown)


src/utils/container.rs
```diff
@@ -528,45 +540,68 @@ impl SdkContainer {
...
+    /// interpolation is done here (no remote-extension composition) since a
+    /// `type: path` source is declared inline and needs no network.
+    pub fn derive_ext_path_mounts(&self, target: &str) -> Option<HashMap<String, PathBuf>> {
+        let config_path = self.config_path.as_ref()?;
+        let content = std::fs::read_to_string(config_path).ok()?;
+        let mut value: serde_yaml::Value = serde_yaml::from_str(&content).ok()?;
+        // Interpolate so templated extension keys (e.g.
+        // `avocado-bsp-{{ avocado.target.board }}`) and `path` values resolve.
+        crate::utils::interpolation::interpolate_config(&mut value, Some(target)).ok()?;
+
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: `derive_ext_path_mounts` now warns (when verbose) and returns None explicitly on read/parse/interpolation failure instead of silently disabling all path mounts via `.ok()?` — restoring the diagnosability the previous state-file path had.

src/utils/config.rs
```diff
@@ -2162,15 +2167,11 @@ impl Config {
...
-                    format!("/opt/_avocado/{resolved_target}/includes/{ext_name}/avocado.yaml")
-                }
+            let ext_config_path_str = if let Some(ref source_path) = path_source_dir {
+                source_path
+                    .join("avocado.yaml")
+                    .to_string_lossy()
+                    .to_string()
             } else {
                 format!("/opt/_avocado/{resolved_target}/includes/{ext_name}/avocado.yaml")
             };
```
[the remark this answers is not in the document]
mobileoverlord: Fixed: the recorded extension config path now checks for `avocado.yml` and records whichever file actually exists (falling back to `avocado.yaml`), so downstream relative-path resolution reads the real file.

src/utils/container.rs
```diff
@@ -528,45 +540,97 @@ impl SdkContainer {
...
-    /// mounted via bindfs at container runtime.
-    pub fn load_ext_path_mounts(&self) -> Option<HashMap<String, PathBuf>> {
-        use crate::utils::ext_fetch::ExtensionPathState;
+    /// The config is the single source of truth: flipping an extension's source
+    /// between `package` and `path` takes effect immediately with no separate
+    /// state file to reconcile or go stale. Only lightweight parsing +
+    /// interpolation is done here (no remote-extension composition) since a
+    /// `type: path` source is declared inline and needs no network.
+    pub fn derive_ext_path_mounts(&self, target: &str) -> Option<HashMap<String, PathBuf>> {
+        let config_path = self.config_path.as_ref()?;
```
jetm: This early-return is the crux: `derive_ext_path_mounts` returns `None` whenever `config_path` is `None`, and only `from_config()` sets it — `SdkContainer::new()` leaves it `None`. `ext install` and `runtime install` (and other `new()`-based sites) then mount zero `type: path` extensions, and nothing sets `config.ext_path_mounts` to compensate. The old `load_ext_path_mounts` read `ext-paths.json` via `src_dir`/cwd regardless of `config_path`, so this is a silent dropped-extension regression for those commands — a runtime that depends on a `type: path` extension is built against an empty includes dir. Deriving from the in-memory `Config` the callers already hold (or setting `config_path` in `new()`) would restore coverage.

src/utils/container.rs
```diff
@@ -528,45 +540,97 @@ impl SdkContainer {
...
+        let mut value: serde_yaml::Value = match serde_yaml::from_str(&content) {
+            Ok(v) => v,
+            Err(e) => {
+                warn("parse", &e);
+                return None;
+            }
+        };
+        // Interpolate so templated extension keys (e.g.
+        // `avocado-bsp-{{ avocado.target.board }}`) and `path` values resolve.
+        if let Err(e) = crate::utils::interpolation::interpolate_config(&mut value, Some(target)) {
```
jetm: Interpolating the entire config here means any interpolation or parse error anywhere returns `None` and drops all path mounts, and the warning is `verbose`-only, so it fails silently. The old `ext-paths.json` state was interpolation-independent. Interpolating only the `extensions` subtree, and surfacing the error outside verbose mode, would avoid dropping mounts because of an unrelated field.

src/utils/container.rs
```diff
@@ -528,45 +540,97 @@ impl SdkContainer {
...
             }
+            let Some(path) = source.get("path").and_then(|p| p.as_str()) else {
+                continue;
+            };
+            let resolved = if Path::new(path).is_absolute() {
+                PathBuf::from(path)
+            } else {
+                src_dir.join(path)
+            };
+            let resolved = resolved.canonicalize().unwrap_or(resolved);
```
jetm, line 641: `canonicalize().unwrap_or(resolved)` with no existence / `avocado.yaml` check means a typo'd `path:` source is bindfs-mounted against a nonexistent host dir and fails at container launch, instead of the clear pre-flight error `ext fetch` used to give. The resolution base also differs from the removed `fetch_from_path` (which fell back to the config-file dir when `src_dir` was unset) — worth confirming they match so the mount doesn't point at the wrong directory when run from a different cwd with `src_dir` unset.

jetm (commented):
Reviewed dropping `ext-paths.json` in favor of deriving mounts from config. There's a silent dropped-mount regression I'd treat as blocking (inline on `container.rs`), plus two related issues.

src/utils/container.rs
mobileoverlord: Fixed (blocking): `derive_ext_path_mounts` no longer returns None when `config_path` is unset — it falls back to `avocado.yaml`/`.yml` under `src_dir` (or cwd), mirroring the old `ext-paths.json` src_dir/cwd lookup. So `ext install` / `runtime install` (bare `SdkContainer::new()`) mount `type: path` extensions again. Added a regression test.

src/utils/container.rs
mobileoverlord: Fixed: interpolation is now scoped to the `extensions` subtree (context built from the full config, so avocado.target/board still resolve), so an error in an unrelated top-level field can't drop all mounts; and read/parse/interpolate failures now warn unconditionally rather than only under `--verbose`.

src/utils/container.rs
mobileoverlord: Fixed: a `type: path` source whose resolved dir doesn't exist is now skipped with a clear warning instead of being bindfs-mounted against a nonexistent path (cryptic launch failure) — mirroring the pre-flight check `ext fetch` had. The resolution base is now `src_dir` else the config file's directory (matching the removed `fetch_from_path`), so it no longer depends on the process cwd.

jetm (approved):
All findings addressed — `config_path` falls back to `src_dir`/cwd so `new()`-based commands still mount `type: path` extensions (with a regression test), interpolation is scoped to the extensions subtree, missing sources are skipped with a warning, and the resolution base matches the old behavior. CI green. LGTM.

PR #168, opened 2026-07-07
honor target-<name>: overrides for rootfs/initramfs/kernel sections

bsandu-peridio:
## Summary

Per-target overrides declared as `target-<name>:` inside the `rootfs:`, `initramfs:`, and `kernel:` sections are now honored by the image/build commands — most importantly for image settings such as a custom `--tag`. This matches the override behavior already used in extension image flows.

Before this change, adding a `target-<name>:` sub-key to one of these singleton sections would trip the "cannot mix singleton and named-map form" guard in the deserializer, and even where it parsed, the override was never applied when composing the image node.

## Changes

- **`src/utils/config.rs`** — `named_or_single_deserializer` strips `target-*` / `kernel-*` override sub-keys before inferring singleton-vs-named shape, so a singleton section with overrides isn't mistaken for a mixed form. The overrides are resolved later by the image commands on the raw composed value.
- **`src/commands/rootfs/image.rs`, `initramfs/image.rs`, `kernel/image.rs`** — apply `resolve_overrides_in_value(...)` to the composed section node before reading the image type/args, so per-target overrides take effect. This runs on the already-composed value, preserving path-based image sources.
- **`src/commands/runtime/build.rs`** — merge per-target overrides before extracting image type/args for runtime manifest/build script generation, matching the standalone `avocado <role> image` commands.

### Example

```yaml
rootfs:
  image:
    type: kab
    args: '-b -t kos.layer.rootfs --tag base'
  target-qemux86-64:
    image:
      args: '-b -t kos.layer.rootfs --tag qemu-x64'
  target-qemuarm64:
    image:
      args: '-b -t kos.layer.rootfs --tag qemu-arm64'
```

## Review follow-ups (Copilot)

- **Named-map form + override keys**: the deserializer no longer silently drops `target-*` / `kernel-*` keys when a section is in named-map form. Since overrides only make sense in singleton form (they override the singleton's config fields), their presence in named-map form is treated as a misplaced/mis-indented override and rejected with a clear error.
- **Test coverage**: added regression tests for `target-<name>` overrides in the `initramfs:` and `kernel:` sections (including a `kernel-<spec>:` example), plus a test asserting the named-map rejection above. Previously only `rootfs:` was covered.

## Testing

- `cargo test --lib` — all config tests pass (220 passed), including the 4 new tests.
- `cargo clippy --lib` — clean.

### DIFF
--- a/src/commands/kernel/image.rs
+++ b/src/commands/kernel/image.rs
@@ -106,7 +106,16 @@ impl KernelImageCommand {
 
         print_info("Building kernel image.", OutputLevel::Normal);
 
-        let kernel_node = composed.merged_value.get("kernel");
+        // Honor per-target `target-<name>:` overrides inside the `kernel:`
+        // section (e.g. a custom `--tag`). Apply the same override merge
+        // extensions use, but on the already-composed value so path-based
+        // sources (merge_path_based_image_sections) are preserved.
+        let kernel_merged = composed
+            .merged_value
+            .get("kernel")
+            .cloned()
+            .map(|v| config.resolve_overrides_in_value(v, &target_arch, None, "kernel"));
+        let kernel_node = kernel_merged.as_ref();
         let image_type = kernel_node
             .and_then(get_ext_image_type)
             .unwrap_or_else(|| "raw".to_string());
--- a/src/commands/kernel/image.rs
+++ b/src/commands/kernel/image.rs
@@ -107,14 +107,13 @@ impl KernelImageCommand {
         print_info("Building kernel image.", OutputLevel::Normal);
 
         // Honor per-target `target-<name>:` overrides inside the `kernel:`
-        // section (e.g. a custom `--tag`). Apply the same override merge
-        // extensions use, but on the already-composed value so path-based
-        // sources (merge_path_based_image_sections) are preserved.
-        let kernel_merged = composed
-            .merged_value
-            .get("kernel")
-            .cloned()
-            .map(|v| config.resolve_overrides_in_value(v, &target_arch, None, "kernel"));
+        // section (e.g. a custom `--tag`). Resolved on the already-composed
+        // value so path-based sources (merge_path_based_image_sections) are
+        // preserved. `kernel-<spec>:` overrides can't be applied here (the
+        // kernel version isn't known until inside the container), so the helper
+        // warns rather than silently dropping them.
+        let kernel_merged =
+            config.resolve_image_section(&composed.merged_value, "kernel", &target_arch);
         let kernel_node = kernel_merged.as_ref();
         let image_type = kernel_node
             .and_then(get_ext_image_type)
--- a/src/commands/runtime/build.rs
+++ b/src/commands/runtime/build.rs
@@ -1389,27 +1389,28 @@ fi"#
                 .join(" ")
         };
 
-        // Resolve `image:` config for rootfs / initramfs / kernel. Same
-        // dynamic accessors as extensions — the helpers don't care about
-        // the parent yaml node's identity.
-        let resolve_img = |yaml_key: &str| -> (String, Option<String>) {
-            // Honor per-target `target-<name>:` overrides (e.g. a custom
-            // `--tag`) inside the rootfs/initramfs/kernel section, matching the
-            // standalone `avocado <role> image` commands.
-            let merged = parsed
-                .get(yaml_key)
-                .cloned()
-                .map(|v| config.resolve_overrides_in_value(v, target_arch, None, yaml_key));
-            let node = merged.as_ref();
-            let t = node
+        // Resolve the rootfs / initramfs / kernel sections once, folding in
+        // per-target `target-<name>:` overrides (e.g. a custom `--tag`) to match
+        // the standalone `avocado <role> image` commands. Resolving once here
+        // means every reader below (image type/args *and* `post_install`) sees
+        // the same per-target node instead of the raw section.
+        let rootfs_section = config.resolve_image_section(parsed, "rootfs", target_arch);
+        let initramfs_section = config.resolve_image_section(parsed, "initramfs", target_arch);
+        let kernel_section = config.resolve_image_section(parsed, "kernel", target_arch);
+
+        // Extract `image:` type/args from an already-resolved section. Same
+        // dynamic accessors as extensions — the helpers don't care about the
+        // parent yaml node's identity.
+        let resolve_img = |section: Option<&serde_yaml::Value>| -> (String, Option<String>) {
+            let t = section
                 .and_then(crate::utils::config::get_ext_image_type)
                 .unwrap_or_else(|| "raw".to_string());
-            let a = node.and_then(crate::utils::config::get_ext_image_args);
+            let a = section.and_then(crate::utils::config::get_ext_image_args);
             (t, a)
         };
-        let (rootfs_image_type, rootfs_image_args) = resolve_img("rootfs");
-        let (initramfs_image_type, initramfs_image_args) = resolve_img("initramfs");
-        let (kernel_image_type, kernel_image_args) = resolve_img("kernel");
+        let (rootfs_image_type, rootfs_image_args) = resolve_img(rootfs_section.as_ref());
+        let (initramfs_image_type, initramfs_image_args) = resolve_img(initramfs_section.as_ref());
+        let (kernel_image_type, kernel_image_args) = resolve_img(kernel_section.as_ref());
 
         // Default kabtool args mirror ext/image.rs:879. Used when the
         // config provides `image: { type: kab }` without an explicit `args`.
@@ -2268,7 +2269,10 @@ echo "Docker image priming complete.""#,
             .resolve_runtime_initramfs(&self.runtime_name)
             .or_else(|| config.initramfs_default());
 
-        let rootfs_post_install = get_post_install(parsed.get("rootfs"));
+        // Read `post_install` from the per-target-resolved section (not raw
+        // `parsed`) so a `target-<name>:` override's script is honored, matching
+        // how the image tag override is applied above.
+        let rootfs_post_install = get_post_install(rootfs_section.as_ref());
         let rootfs_permissions_section = render_perms(resolved_rootfs, "$ROOTFS_WORK/etc");
         let rootfs_build_section = generate_rootfs_build_script(
             NAMESPACE_UUID,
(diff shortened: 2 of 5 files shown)


src/utils/config.rs
```diff
@@ -50,6 +50,22 @@ mod named_or_single_deserializer {
...
+            k.as_str()
+                .map(|s| s.starts_with("target-") || s.starts_with("kernel-"))
+                .unwrap_or(false)
+        };
+        let mapping: serde_yaml::Mapping = mapping
+            .iter()
+            .filter(|(k, _)| !is_override_key(k))
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect();
+
```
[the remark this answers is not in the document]
mobileoverlord: Addressed in 8380a0b: the deserializer now tracks whether any override keys were stripped and returns a hard error in the named-map branch instead of silently dropping them, since `target-*`/`kernel-*` overrides only make sense in singleton form. Added `test_override_keys_in_named_map_form_are_rejected` covering this.

src/utils/config.rs
```diff
@@ -10809,6 +10827,31 @@ rootfs:
         );
     }
 
+    #[test]
+    fn test_rootfs_target_override_keys_do_not_trip_mixed_form() {
+        // Per-target override sub-keys inside a singleton rootfs must not trip
+        // the mix-of-forms guard; they are stripped from the typed struct
+        // (the image command resolves them on the raw composed value for the tag).
```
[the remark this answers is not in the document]
mobileoverlord: Addressed in 8380a0b: added `test_initramfs_target_override_keys_do_not_trip_mixed_form` and `test_kernel_target_override_keys_do_not_trip_mixed_form` (the latter includes a `kernel-6.6:` spec example).

mobileoverlord (dismissed):

src/commands/kernel/image.rs
```diff
@@ -106,7 +106,16 @@ impl KernelImageCommand {
...
-        let kernel_node = composed.merged_value.get("kernel");
+        // Honor per-target `target-<name>:` overrides inside the `kernel:`
+        // section (e.g. a custom `--tag`). Apply the same override merge
+        // extensions use, but on the already-composed value so path-based
+        // sources (merge_path_based_image_sections) are preserved.
+        let kernel_merged = composed
+            .merged_value
+            .get("kernel")
+            .cloned()
+            .map(|v| config.resolve_overrides_in_value(v, &target_arch, None, "kernel"));
```
jetm: All four sites pass `None` as the third arg (`resolved_kver`), and in `resolve_overrides_in_value` the `kernel-<spec>:` branch pushes the key to `keys_to_remove` unconditionally but only merges the override under `if let Some(kver) = resolved_kver`. So with `None` a `kernel-6.6:` override is stripped and never applied — `avocado kernel image` builds with the base `--tag`, not the override. Since the kernel image command only learns the kernel version inside the container shell (the `Image-*` glob), it has no kver to pass here, so `kernel-<spec>:` support is effectively dead in this path even though the tests and section comments advertise it. `target-<name>:` works; `kernel-<spec>:` doesn't.

jetm (commented):
Reviewed the `target-<name>:`/`kernel-<spec>:` override support. Two correctness issues stand out: `kernel-<spec>:` overrides are silently never applied (all sites pass `None` for the kver), and `runtime build` ignores per-target `post_install` overrides while honoring the tag. Two more are behavior questions around the deserializer's prefix stripping. Details inline.

src/commands/runtime/build.rs
```diff
@@ -1393,7 +1393,14 @@ fi"#
...
         // the parent yaml node's identity.
         let resolve_img = |yaml_key: &str| -> (String, Option<String>) {
-            let node = parsed.get(yaml_key);
+            // Honor per-target `target-<name>:` overrides (e.g. a custom
+            // `--tag`) inside the rootfs/initramfs/kernel section, matching the
+            // standalone `avocado <role> image` commands.
+            let merged = parsed
+                .get(yaml_key)
+                .cloned()
+                .map(|v| config.resolve_overrides_in_value(v, target_arch, None, yaml_key));
```
jetm: `resolve_img` resolves the image tag/type override here, but `get_post_install(parsed.get("rootfs"))` and `.get("initramfs")` later in this command (~2264/2273) still read the raw `parsed` section, so a per-target `post_install` override is ignored while the tag override is honored. The standalone `avocado rootfs image` command resolves the node first and would run the override script, so the two paths diverge for the same config. Separately, this `.get(section).cloned().map(|v| resolve_overrides_in_value(v, target, None, section))` dance is now copy-pasted at four sites — resolving overrides once during composition (right after `merge_path_based_image_sections`) would close the `post_install` gap and stop every future reader of `merged_value` from having to repeat it.

src/utils/config.rs
```diff
@@ -50,6 +50,24 @@ mod named_or_single_deserializer {
             }
         };
 
+        // Per-target / per-kernel override sub-keys (`target-<name>:`,
+        // `kernel-<spec>:`) are resolved later by the image commands on the raw
+        // composed value. Drop them before shape inference and the typed
+        // deserialize below, so a `rootfs: { image: ..., target-qemux86-64: ... }`
+        // isn't mistaken for a mix of singleton + named-map forms.
+        let is_override_key = |k: &serde_yaml::Value| {
```
jetm, line 78: `is_override_key` strips any key starting `target-`/`kernel-` before `field_key_count`, regardless of form. That's right for singleton sections, but it also fires on a *named-map* section whose entry names happen to start with those prefixes — e.g. `kernel: { kernel-lts: {...}, kernel-mainline: {...} }` referenced by a runtime as `kernel: kernel-lts`. All entries get stripped, `field_key_count == 0 && had_override_keys`, and the guard below hard-errors, so a config that parsed fine on `main` fails to load. Repo examples use names like `yocto-6-6` so it may not bite today, but the prefix match is broader than the singleton-override intent.

src/utils/config.rs
```diff
@@ -58,8 +76,22 @@ mod named_or_single_deserializer {
...
         if field_key_count == 0 {
-            // Named-map form.
+            // Named-map form. `target-<name>:`/`kernel-<spec>:` overrides only
+            // make sense inside singleton form (they override the singleton's
+            // config fields), so their presence here means the section has no
+            // recognized config fields to override — most likely a mis-indented
+            // or misplaced override. Reject it rather than silently dropping it.
+            if had_override_keys {
+                return Err(de::Error::custom(format!(
+                    "{kind_label}: `target-<name>:`/`kernel-<spec>:` override keys are only valid \
```
jetm: This rejects a singleton section containing only `target-*`/`kernel-*` keys and no recognized base field. On `main` such a section parsed as a named-map (`field_key_count == 0` → `Ok`), so it's a behavior change: a per-target-only section with no shared base (e.g. `rootfs: { target-qemux86-64: {...}, target-qemuarm64: {...} }`) now hard-errors. If per-target-only sections aren't meant to be supported, worth a doc note; if they are, this locks them out.

src/utils/config.rs
```diff
@@ -50,6 +50,24 @@ mod named_or_single_deserializer {
...
+        // composed value. Drop them before shape inference and the typed
+        // deserialize below, so a `rootfs: { image: ..., target-qemux86-64: ... }`
+        // isn't mistaken for a mix of singleton + named-map forms.
+        let is_override_key = |k: &serde_yaml::Value| {
+            k.as_str()
+                .map(|s| s.starts_with("target-") || s.starts_with("kernel-"))
+                .unwrap_or(false)
+        };
+        let original_len = mapping.len();
+        let mapping: serde_yaml::Mapping = mapping
```
jetm: This rebuilds the whole section mapping — cloning every key and nested value — on every parse, even when there are no `target-*`/`kernel-*` keys (the common case for existing configs). Since `value` is already owned, `mapping.retain(|k, _| !is_override_key(k))` moves entries instead of deep-cloning, and probing `mapping.keys().any(is_override_key)` first keeps the no-override path allocation-free.

src/commands/kernel/image.rs
mobileoverlord: Good catch — fixed in 4a3b74a. Since these build paths genuinely can't know the kernel version at CLI time (it's only discovered inside the container via the `Image-*` glob), `kernel-<spec>:` can't be resolved here at all. Rather than leaving it silently stripped, the new `resolve_image_section` helper now emits a warning when a `kernel-<spec>:` key is present in an image/build section, telling the user only `target-<name>:` is honored there. `resolve_overrides_in_value`'s documented None-skip is unchanged (the ext/install path still re-applies with `Some(kver)`).

src/commands/runtime/build.rs
mobileoverlord: Both fixed in 4a3b74a. `runtime build` now resolves each section once (`config.resolve_image_section(parsed, section, target_arch)`) up front and reads both the image tag/type *and* `post_install` from that resolved node, so a per-target `post_install` override is honored and no longer diverges from `avocado <role> image`. That same helper replaces the copy-pasted `.get().cloned().map(resolve_overrides_in_value)` dance at all four sites. I kept the resolution in the commands rather than in `load_composed` because the definitive target isn't known until `resolve_target_required` runs *after* composition (load_composed may be passed `None`).

src/utils/config.rs
mobileoverlord: Fixed in 4a3b74a. The deserializer now computes `field_key_count` on the raw mapping first and only strips `target-*`/`kernel-*` keys in singleton form (when config fields are present). A named-map section like `kernel: { kernel-lts: {...}, kernel-mainline: {...} }` has no config fields, so it's treated as pure named-map and round-trips untouched. Added `test_named_map_entries_with_reserved_prefixes_are_not_stripped`.

src/utils/config.rs
mobileoverlord: Fixed in 4a3b74a. With the strip now confined to singleton form, a per-target-only section (`field_key_count == 0`) parses as a named map exactly as it did on `main` — no hard-error. The image commands still resolve the per-target override on the raw composed value. Added `test_per_target_only_section_parses`.

src/utils/config.rs
mobileoverlord: Fixed in 4a3b74a. The mapping is now owned (`Option::deserialize` gives us ownership), and stripping is guarded by a probe: `if mapping.keys().any(is_override_key) { mapping.retain(...) }`. Override-free configs (the common case) never rebuild or clone the mapping; only the singleton branch touches it, and via `retain` (move) rather than a deep clone.

src/utils/config.rs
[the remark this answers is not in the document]
mobileoverlord: Follow-up: per jetm's review below, I reversed the hard-error approach from 8380a0b. Erroring on `target-*`/`kernel-*` keys in named-map form would break legitimate named-map sections whose entry names start with those prefixes (e.g. `kernel: { kernel-lts: ... }`). As of 4a3b74a, prefix stripping only happens in singleton form; in named-map form all keys are treated as entry names and preserved (no silent drop, no false error).

jetm (approved):
All review findings addressed in 4a3b74a — `kernel-<spec>:` now warns instead of silently dropping, `runtime build` reads `post_install` from the per-target-resolved section, override-key stripping is confined to singleton form (named-map and override-only sections parse as before), and the shared `resolve_image_section` helper de-duplicates the four sites. CI green. LGTM.

PR #175, opened 2026-07-08
ext/package: emit RPM-safe Version for pre-release semver

mobileoverlord:
## Problem
Packaging an extension whose version is a valid semver **pre-release** (e.g. `1.0.0-rc.1`) fails at spec build:

```
Building for target noarch
error: line 5: Illegal char '-' (0x2d) in: Version: 1.0.0-rc.1
```

RPM forbids `-` in the `Version:` field — it's the Version/Release separator. The extension packager wrote the raw semver string straight into the spec, so any `-rc.N`/`-alpha`/`-beta` version is rejected. Surfaced by the `1.0.0-rc.1` release, but it applies to any pre-release ext version.

## Fix
Map the semver version to an RPM-compatible form for the **spec `Version:`** and the **built RPM's filename**:
- `-` → `~` (RPM pre-release: `1.0.0~rc.1` sorts *before* `1.0.0`, matching semver pre-release precedence)
- `+` → `^` (build metadata → RPM post-release)

Plain releases (`1.0.0`) are unchanged. New helper `utils::version::to_rpm_version`.

The **semver form is preserved** for the `avocado.yaml` baked into the package payload — consumers validate that field as semver, and `~` is not valid semver — so only the RPM artifacts (Version + filename) get the transformed value. This keeps the human/avocado-facing version as `1.0.0-rc.1` while the RPM/feed sees `1.0.0~rc.1`.

## Test
- New unit test `test_to_rpm_version` (plain / `-rc.1` / `-alpha.2` / `+build.5`).
- `cargo build` clean; `version::tests` (31) and `ext::package` (13) suites pass.

## Release note
The packaging is done by the *installed* avocado CLI, so this fix only takes effect once it ships in a CLI release. `1.0.0-rc.1` (already tagged) still packages with the bug — the fix will land in the next tag (rc.2). That release will also want the ext-release/release asset-upload race fix discussed on the prior PR, or its Extension Release run must be re-run after `release.yml` finishes uploading.

### DIFF
--- a/src/commands/ext/package.rs
+++ b/src/commands/ext/package.rs
@@ -525,6 +525,188 @@ impl ExtPackageCommand {
         format!("System extension package for {name}")
     }
 
+    /// Generate the `rpmbuild` shell script for this extension.
+    ///
+    /// Split out of [`Self::create_rpm_package_in_container`] so the two places
+    /// the RPM-form version lands — the spec's `Version:` line and the NVR
+    /// filename — are assertable without running a container. Mirrors
+    /// `sdk::package`'s `generate_rpm_build_script`.
+    #[allow(clippy::too_many_arguments)]
+    fn generate_rpm_build_script(
+        &self,
+        metadata: &RpmMetadata,
+        rpm_version: &str,
+        rpm_filename: &str,
+        target_provides: &str,
+        container_src_dir: &str,
+        package_files_str: &str,
+        version_bake_section: &str,
+    ) -> String {
+        format!(
+            r#"
+set -e
+
+# Extension source directory
+EXT_SRC_DIR="{container_src_dir}"
+
+# Package files patterns (may contain globs like * and **)
+PACKAGE_FILES="{package_files_str}"
+
+# Ensure output directory exists
+mkdir -p $AVOCADO_PREFIX/output/extensions
+
+# Check if extension source directory exists
+if [ ! -d "$EXT_SRC_DIR" ]; then
+echo "Extension source directory not found: $EXT_SRC_DIR"
+exit 1
+fi
+
+# Check for avocado config file
+if [ ! -f "$EXT_SRC_DIR/avocado.yaml" ] && [ ! -f "$EXT_SRC_DIR/avocado.yml" ]; then
+echo "No avocado.yaml/yml found in $EXT_SRC_DIR"
+exit 1
+fi
+
+# Create temporary directory for RPM build
+TMPDIR=$(mktemp -d)
+STAGING_DIR="$TMPDIR/staging"
+mkdir -p "$STAGING_DIR"
+cd "$TMPDIR"
+
+# Create directory structure for rpmbuild
+mkdir -p BUILD RPMS SOURCES SPECS SRPMS
+
+# Enable globstar for ** pattern support
+shopt -s globstar nullglob
+
+# Copy files matching patterns to staging directory
+cd "$EXT_SRC_DIR"
+FILE_COUNT=0
+for pattern in $PACKAGE_FILES; do
+# Expand the glob pattern
+for file in $pattern; do
+    if [ -e "$file" ]; then
+        # Create parent directory in staging and copy
+        parent_dir=$(dirname "$file")
+        if [ "$parent_dir" != "." ]; then
+            mkdir -p "$STAGING_DIR/$parent_dir"
+        fi
+        cp -r "$file" "$STAGING_DIR/$file"
+        if [ -f "$file" ]; then
+            FILE_COUNT=$((FILE_COUNT + 1))
+        elif [ -d "$file" ]; then
+            dir_files=$(find "$file" -type f | wc -l)
+            FILE_COUNT=$((FILE_COUNT + dir_files))
+        fi
+    fi
+done
+done
+cd "$TMPDIR"
+{version_bake_section}
+echo "Creating RPM with $FILE_COUNT files from source directory..."
+
+if [ "$FILE_COUNT" -eq 0 ]; then
+echo "No files matched the package_files patterns: $PACKAGE_FILES"
+exit 1
+fi
+
+# Create spec file
+# The extension's src_dir maps to a top-level /<ext_name>/ directory in the package, so
+# that installing into a SHARED includes installroot lands its content at
+# includes/<ext_name>/ without colliding with other extensions' files (and one rpmdb
+# tracks all installed extensions).
+cat > SPECS/package.spec << SPEC_EOF
+%define _buildhost reproducible
+# The payload is already-built target content, so rpm has to package it verbatim.
+# Two stock behaviours get in the way: the build-root policy scripts run the
+# container's host toolchain (strip and friends) over target binaries, and the
+# noarch guard rejects a payload carrying target ELF. These packages are noarch
+# by design - a target routes on the avocado-target provides below, not on the
+# package arch - so both checks are false positives here.
+# Comments must stay macro-free; rpm expands macros inside spec comments.
+%define __os_install_post %{{nil}}
+%define _binaries_in_noarch_packages_terminate_build 0
+AutoReqProv: no
+
+Name: {name}
+Version: {version}
+Release: {release}
+Summary: {summary}
+License: {license}
+Vendor: {vendor}
+Group: {group}{url_line}
+# Self-describe the on-disk layout so the CLI knows how to install this package: content
+# is nested under /<ext_name>/, so it installs into the SHARED includes installroot.
+# Legacy packages (content at /) lack this provide and use the per-ext installroot.
+Provides: avocado-ext-layout(nested)
+# Self-describe which targets this extension supports, derived from avocado.yaml
+# supported_targets. Surfaced in the feed's repo metadata (primary.xml) so the CLI can
+# query target compatibility without downloading the RPM, and so the feed server can route
+# it to the correct per-target -ext feed(s). "*" means all targets (cross-target).
+{target_provides}
+
+%description
+{description}
+
+%files
+/{name}
+
+%prep
+# No prep needed
+
+%build
+# No build needed
+
+%install
+# Nest the staged files under /<ext_name>/ so a shared includes installroot yields
+# includes/<ext_name>/... (collision-free, one rpmdb per includes root).
+mkdir -p %{{buildroot}}/{name}
+cp -r "$STAGING_DIR"/* %{{buildroot}}/{name}/
+
+%clean
+# Skip clean section - not needed for our use case
+
+%changelog
+SPEC_EOF
+
+# Build the RPM with custom architecture target
+rpmbuild --define "_topdir $TMPDIR" --define "_arch {arch}" --target {arch} -bb SPECS/package.spec
+
+# Move RPM to output directory
+mv RPMS/{arch}/*.rpm $AVOCADO_PREFIX/output/extensions/{rpm_filename} || {{
+mv RPMS/*/*.rpm $AVOCADO_PREFIX/output/extensions/{rpm_filename} 2>/dev/null || {{
+    echo "Failed to find built RPM"
+    exit 1
+}}
+}}
+
+echo "RPM created successfully: $AVOCADO_PREFIX/output/extensions/{rpm_filename}"
+
+# Cleanup
+rm -rf "$TMPDIR"
+"#,
+            name = metadata.name,
+            version = rpm_version,
+            release = metadata.release,
+            summary = metadata.summary,
+            license = metadata.license,
+            vendor = metadata.vendor,
+            group = metadata.group,
+            url_line = if let Some(url) = &metadata.url {
+                format!("\nURL: {url}")
+            } else {
+                String::new()
+            },
+            description = metadata.description,
+            arch = metadata.arch,
+            target_provides = target_provides,
+            rpm_filename = rpm_filename,
+            container_src_dir = container_src_dir,
+            package_files_str = package_files_str,
+            version_bake_section = version_bake_section,
+        )
+    }
+
     /// Create the RPM package containing the extension's src_dir
     ///
     /// The package root (/) maps to the extension's src_dir contents.
@@ -576,7 +758,8 @@ impl ExtPackageCommand {
         // the semver version to its RPM form (`1.0.0-rc.1` -> `1.0.0~rc.1`). Used
         // for the spec `Version:` and the built RPM's name; the semver form is kept
         // for the avocado.yaml baked into the payload (consumers validate semver).
-        let rpm_version = crate::utils::version::to_rpm_version(&metadata.version);
+        let rpm_version = crate::utils::version::to_rpm_version(&metadata.version)
+            .with_context(|| format!("Extension '{}' cannot be packaged", self.extension))?;
 
         // Create the RPM filename (matches the built RPM's NVR, so it uses the
         // RPM-form version, not the semver form).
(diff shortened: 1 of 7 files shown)


src/utils/version.rs
```diff
@@ -71,6 +71,18 @@ pub fn check_cli_requirement(requirement: &str) -> Result<()> {
...
+///
+/// RPM forbids `-` in the Version field (it is the Version/Release separator),
+/// so a semver pre-release like `1.0.0-rc.1` is illegal and `rpmbuild` rejects
+/// it. RPM uses `~` for pre-release ordering — `1.0.0~rc.1` sorts *before*
+/// `1.0.0`, matching semver pre-release precedence — and `^` for post-release,
+/// so map the pre-release `-` to `~` and the build-metadata `+` to `^`. A plain
+/// release version (no `-`/`+`) is returned unchanged.
+pub fn to_rpm_version(version: &str) -> String {
+    version.replace('-', "~").replace('+', "^")
+}
```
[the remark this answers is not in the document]
mobileoverlord: The observation is correct but the suggested fix would break the build, so keeping the global replace — now documented, with the edge cases pinned in 3611319.

RPM rejects `-` **anywhere** in `Version:`, not just as the Version/Release separator. So rewriting only the first hyphen turns `1.0.0-rc-1` into `1.0.0~rc-1`, which still carries a `-` and which `rpmbuild` refuses outright. The internal hyphens can't be left alone; they have to become something, and `~` is the only character that preserves pre-release ordering.

So the lossiness is the deliberate trade: `1.0.0-rc-1` -> `1.0.0~rc~1` doesn't round-trip back to semver. That's acceptable here because this value is only ever the spec's `Version:` and the RPM's NVR filename — the **semver** form is what gets baked into the payload's `avocado.yaml`, which is what consumers validate. Nothing reads the version back out of the RPM Version field expecting semver.

You're right about one real wart, though it's a different one: `1.0.0+build-5` -> `1.0.0^build~5` puts a `~` (pre-release) after a `^` (post-release), which is odd ordering-wise. It's theoretical for this project — versions here are `X.Y.Z` and `X.Y.Z-rc.N` — and the cleaner answer would be dropping build metadata entirely, since semver says it's ignored for precedence. Not doing that in this PR; it'd be a behavior change beyond making pre-release versions buildable.

Added a doc comment on `to_rpm_version` recording all of the above so the global replace doesn't read as an oversight to the next person.

src/utils/version.rs
```diff
@@ -127,6 +139,18 @@ mod tests {
...
+    fn test_to_rpm_version() {
+        // Plain release versions are unchanged.
+        assert_eq!(to_rpm_version("1.0.0"), "1.0.0");
+        assert_eq!(to_rpm_version("2.1.3"), "2.1.3");
+        // Pre-release `-` becomes `~` (sorts before the release in RPM).
+        assert_eq!(to_rpm_version("1.0.0-rc.1"), "1.0.0~rc.1");
+        assert_eq!(to_rpm_version("1.0.0-alpha.2"), "1.0.0~alpha.2");
+        // Build metadata `+` becomes `^`.
+        assert_eq!(to_rpm_version("1.0.0+build.5"), "1.0.0^build.5");
+    }
```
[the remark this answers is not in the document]
mobileoverlord: Fair, and these were the useful cases to ask for. Added in 3611319:

`test_to_rpm_version_rewrites_hyphens_inside_identifiers`
- `1.0.0-rc-1` -> `1.0.0~rc~1` (hyphen inside a pre-release identifier)
- `1.0.0-pre-release-2` -> `1.0.0~pre~release~2`
- `1.0.0+build-5` -> `1.0.0^build~5` (hyphen inside build metadata)
- plus the invariant that actually matters, asserted over four shapes including `2.0.0-a-b-c+d-e-f`: **no `-` survives**, because a surviving hyphen is a spec `rpmbuild` rejects

`test_to_rpm_version_handles_prerelease_and_build_together`
- `1.0.0-rc.1+build.5` -> `1.0.0~rc.1^build.5` (both separators in one pass)
- `1.0.0-rc-1+build-5` -> `1.0.0~rc~1^build~5` (both, with inner hyphens in each)

The no-hyphen-survives loop is the one I'd keep if the others were trimmed — it pins the property RPM cares about rather than a specific string mapping.

All three `to_rpm_version` tests pass locally, `cargo fmt --all -- --check` clean.

mobileoverlord:
@jetm ready for review. Both Copilot threads are addressed and resolved.

The substance: Copilot wanted only the *first* hyphen rewritten, but that would break the build — RPM rejects `-` anywhere in `Version:`, so `1.0.0-rc-1` -> `1.0.0~rc-1` still carries a hyphen and `rpmbuild` refuses it. The global replace is required, so I documented why and pinned the edge cases instead (3611319), including the invariant that no `-` survives in any shape.

One wart I flagged rather than fixed: `1.0.0+build-5` -> `1.0.0^build~5` puts a `~` after a `^`, which is odd ordering-wise. Theoretical for us (we ship `X.Y.Z` and `X.Y.Z-rc.N`), and the cleaner fix is dropping build metadata entirely since semver ignores it for precedence — but that's a behavior change beyond making pre-release versions buildable. Say the word if you'd rather it go in here.

**Heads-up on CI:** the `build (qemux86-64, 2024, edge-next)` failure showing on this PR is stale — it's from 2026-07-08 and failed with `Failed to install target-sysroot with compile dependencies`, an SDK/feed error unrelated to this diff (which only touches `to_rpm_version` and the spec's `Version:`). That same check has passed on every recent PR. The push above should re-run it.

Separately, #191 renames that job's channel from `edge-next` to `next`, which is what we should be testing against — worth landing that first, then this re-runs against the right feed.

src/utils/version.rs
```diff
@@ -71,6 +71,27 @@ pub fn check_cli_requirement(requirement: &str) -> Result<()> {
...
+/// Every `-` is rewritten, not just the first. That is deliberate and not a
+/// shortcut: RPM rejects `-` anywhere in `Version:`, so a hyphen *inside* a
+/// pre-release identifier (`1.0.0-rc-1`, legal semver) has to go too. Rewriting
+/// only the leading separator would emit `1.0.0~rc-1`, which `rpmbuild` refuses
+/// outright. The cost is that the mapping is lossy — `1.0.0-rc-1` and
+/// `1.0.0~rc~1` do not round-trip back to semver — which is acceptable because
+/// the semver form is what gets baked into the payload's `avocado.yaml`; this
+/// value exists only to satisfy RPM's own grammar and ordering.
+pub fn to_rpm_version(version: &str) -> String {
+    version.replace('-', "~").replace('+', "^")
```
jetm, line 92: **Rewriting every hyphen inverts RPM ordering for `rc1-fix` forms**

Your acceptance argument for all-hyphens is correct and I confirmed it - `Version: 1.0.0~rc-1` is rejected with `Illegal char '-'`, so rewriting only the first separator genuinely does not build. But all-hyphens is not free either, and the cost is an ordering inversion rather than anything cosmetic: `~` is a *precedence operator* in RPM and an ordinary character in semver.

Measured with the `semver` 1.0.28 crate and `rpm.labelCompare`, not reasoned:

| semver | RPM after mapping |
|---|---|
| `1.0.0-rc1 < 1.0.0-rc1-fix` | `1.0.0~rc1` **>** `1.0.0~rc1~fix` |
| `1.0.0-rc-1 > 1.0.0-rc.2` | `1.0.0~rc~1` **<** `1.0.0~rc.2` |

So dnf treats the newer hyphenated RC as older and refuses the upgrade.

No substitute character rescues it, which is why I do not think the fix is a different mapping. `_` fixes the first row (`1.0.0~rc1 < 1.0.0~rc1_fix`) but then `1.0.0~rc_1 == 1.0.0~rc.1`, so two distinct semvers compare equal. `.` collides outright.

What I would do instead: reject a hyphen *inside* a pre-release identifier at `validate_semver`/`extract_rpm_metadata` time. That makes the inversion impossible to ship silently, and it costs you only version strings nobody writes. Failing that, record it - see my note on the doc comment above.

src/utils/version.rs
```diff
@@ -71,6 +71,27 @@ pub fn check_cli_requirement(requirement: &str) -> Result<()> {
...
+/// it. RPM uses `~` for pre-release ordering — `1.0.0~rc.1` sorts *before*
+/// `1.0.0`, matching semver pre-release precedence — and `^` for post-release,
+/// so map the pre-release `-` to `~` and the build-metadata `+` to `^`. A plain
+/// release version (no `-`/`+`) is returned unchanged.
+///
+/// Every `-` is rewritten, not just the first. That is deliberate and not a
+/// shortcut: RPM rejects `-` anywhere in `Version:`, so a hyphen *inside* a
+/// pre-release identifier (`1.0.0-rc-1`, legal semver) has to go too. Rewriting
+/// only the leading separator would emit `1.0.0~rc-1`, which `rpmbuild` refuses
+/// outright. The cost is that the mapping is lossy — `1.0.0-rc-1` and
```
jetm, line 87: **The doc records a cost that isn't real, and omits the one that is**

The comment says `1.0.0-rc-1` and `1.0.0~rc~1` "do not round-trip back to semver". They do. Semver's grammar allows only `[0-9A-Za-z-.+]`, so no `~` or `^` can appear in the input; the map is injective, and the inverse (`~` to `-`, `^` to `+`) recovers `1.0.0-rc-1` exactly.

That matters beyond pedantry in two ways. A maintainer who believes the mapping is lossy will not write the reverse mapping - and the reverse mapping is exactly what the publish path needs (see my note in the review body about `connect ext publish` recording `1.0.0~rc.1` as the platform version). Meanwhile the real cost, the ordering inversion on the line below, is recorded nowhere: not here, and not in a test. `test_to_rpm_version_rewrites_hyphens_inside_identifiers` asserts the no-surviving-hyphen property and says nothing about ordering.

Swapping those two claims round would leave the next reader with an accurate picture of what this function does and does not guarantee.

src/commands/ext/package.rs
```diff
@@ -560,10 +560,17 @@ impl ExtPackageCommand {
...
-        // Create the RPM filename
+        // RPM forbids `-` in Version (it is the Version/Release separator), so map
+        // the semver version to its RPM form (`1.0.0-rc.1` -> `1.0.0~rc.1`). Used
+        // for the spec `Version:` and the built RPM's name; the semver form is kept
+        // for the avocado.yaml baked into the payload (consumers validate semver).
+        let rpm_version = crate::utils::version::to_rpm_version(&metadata.version);
+
+        // Create the RPM filename (matches the built RPM's NVR, so it uses the
+        // RPM-form version, not the semver form).
         let rpm_filename = format!(
```
jetm, line 571: **Neither call site of the new helper has a test**

The added tests cover the pure helper well. `create_rpm_package_in_container` - where `rpm_version` reaches the filename here and the spec `Version:` at 782 - has none; all 13 tests in this file target `extract_rpm_metadata` and `get_package_files`.

So a later refactor reverting either interpolation back to `metadata.version` reintroduces the `rpmbuild` abort with a fully green suite. That is the same shape of gap the helper's own doc comment is guarding against, one layer up.

The pattern for this already exists in the repo - `src/commands/ext/build.rs:1722` and `src/commands/ext/clean.rs:546` both assert on generated script text, so a test that pins the generated `Version:` line and filename does not need new machinery.

jetm (commented):
The core mechanic holds, and I checked it against real tooling rather than from memory: `rpmspec` (rpm 6.0.1) rejects `Version: 1.0.0-rc.1` with `Illegal char '-'` and accepts `1.0.0~rc.1`, `1.0.0^build.5` and `1.0.0~rc.1^build.5`, so the fix is needed. `rpm.labelCompare` confirms the direction is right - `1.0.0~rc.1 < 1.0.0`, and `1.0.0~rc.1^build.5 < 1.0.0`, because `rpmvercmp` evaluates the `~` branch before the `^` branch, so build metadata cannot lift a pre-release above its release. `+` to `^` also matches the `semver` crate's own `Ord`. And there are no collisions: valid semver contains no `~` or `^`, so the map is injective.

Three inline comments below. The important one is the ordering inversion for hyphens inside a pre-release identifier - your reasoning for rewriting every hyphen is correct as far as it goes, but the consequence is measurable and currently unrecorded.

The bigger issue is not in this diff at all, which is why it is here instead of inline. **This fixes the publish side and leaves the consumers asking for the old name.** Four call sites, none of them lines this PR touches:

1. **`src/utils/ext_fetch.rs:173` - the install side still asks dnf for the semver form.** `ext package` on `version: "1.0.0-rc.1"` now succeeds and lands `myext-1.0.0~rc.1-r0.noarch.rpm` in a feed. A consumer declaring that extension with `version: "1.0.0-rc.1"` and no lockfile entry builds `format!("{package_name}-{version}")` and runs `dnf install myext-1.0.0-rc.1`, which matches nothing - both the `repoquery --provides` layout probe at `:196` and the install at `:223` miss, so it fails or silently falls through to the legacy installroot. The lockfile branch (`src/utils/lockfile.rs:1596`) is fine because it reuses the rpm-derived NEVRA; only the config branch (`lockfile.rs:1603`, `ext_fetch.rs:173`) is wrong. That is the fresh-checkout path, so it is the common case rather than the corner.

2. **`src/commands/connect/ext.rs:105` - publish records `1.0.0~rc.1` as the platform version.** `connect ext publish` derives the published version by parsing the RPM filename, so the platform now stores the RPM-mangled string. `.github/workflows/ext-release.yml:16` triggers on `'[0-9]+.[0-9]+.[0-9]+-*'` and `Cargo.toml:3` is `1.0.0-rc.1` today, so this fires on the next RC tag: `parse_nevra` yields `version = "1.0.0~rc.1"`, POSTed at `:171`, and `--version` at `:107` is a manual override CI does not pass. Any client comparing that against the `avocado.yaml` baked into the payload - deliberately kept as `1.0.0-rc.1` at `ext/package.rs:611` - sees a mismatch. Worth saying that this diff *improves* that call: pre-change the filename `myext-1.0.0-rc.1-r0.noarch.rpm` made `parse_nevra` return `name="myext-1.0.0", version="rc.1"`. It just stops one step short of inverting the mapping, which is straightforward once the round-trip note on `to_rpm_version` is corrected.

3. **`src/commands/sdk/package.rs:482` - the sibling packaging command is unfixed.** It has its own spec template and filename builder and never calls the new helper. `:217` runs `validate_semver`, which accepts `1.0.0-rc.1`, then `:482` interpolates it raw into `Version:` - which `rpmspec` rejects outright. So `avocado sdk package` still aborts inside the container on any pre-release, and `:397`'s `main_rpm` and `:407`'s sub-package names carry the same wrong form that `ext/package.rs:571` just fixed. `to_rpm_version` living in `src/utils/version.rs` rather than next to its one caller reads like it was written for exactly this.

4. **`src/commands/runtime/build.rs:1247` - lower confidence, container path I cannot exercise locally.** For an extension with `version: "*"`, `resolve_extension_version` (`:2708`) falls back to the rpmdb `%{VERSION}`, now `1.0.0~rc.1`. The heuristic at `:1247` splits on the last `-` and requires a digit-leading remainder: `1.0.0~rc.1` passes where `rc.1` previously did not, which switches a tolerant `ls {ext_name}-*.raw` glob (`:1322`) into the strict exact-name branch at `:1311` that emits `exit 1`. `src/commands/ext/image.rs:565` builds the `.raw` name from the *config* version, so the exact name the strict branch demands may not exist.

Also not a finding, but `docs/features/extension-versioning-redesign.md:5` says the design exists to eliminate "mismatches between config and RPM versions", and this reintroduces one for pre-releases. Worth a line there.

Nothing I checked contradicts the change being the right direction - items 1 and 3 are the ones I would want landed before this ships, since together they mean a published pre-release extension cannot be installed from config.

src/utils/version.rs
mobileoverlord: Fixed in e4ceca7 — `to_rpm_version` now rejects a hyphen inside a pre-release or build identifier rather than rewriting it, which is the option you recommended.

Your measurement is what settled it. I had reasoned about the mapping as lossy-but-harmless; an ordering inversion is a different category of wrong, because nothing downstream can detect it — dnf just declines the upgrade and the user sees no error at all. And since neither leaving the hyphen (rpmbuild refuses it) nor mapping it (inverts) is correct, and no substitute character works for the reasons you gave, there is no mapping to pick. Refusing the input is the only branch that can't emit a wrong answer.

The error names the version, says ordering would invert, and points at `.` as the separator. `test_to_rpm_version_rejects_hyphen_inside_identifier` covers all six shapes including `1.0.0-rc1-fix` from your table.

src/utils/version.rs
mobileoverlord: Correct on both counts, and fixed in e4ceca7.

The round-trip claim was just wrong — as you say, semver admits only `[0-9A-Za-z.+-]`, so no `~` or `^` can appear in the input and the map is injective. I have replaced that paragraph with the ordering-inversion table from your other comment, and added `from_rpm_version` as the actual inverse.

Your point about the reverse mapping being what the publish path needs was the useful consequence: once it exists, review-body items 2 and 4 are both one call. `connect ext publish` now inverts the NVR-derived version before recording it, and `query_rpm_version` inverts `%{VERSION}` — see the review-body reply.

src/commands/ext/package.rs
mobileoverlord: Fixed in e4ceca7. I extracted the `format!` out of `create_rpm_package_in_container` into `generate_rpm_build_script`, following the `build.rs:1722` / `clean.rs:546` pattern you pointed at, so the generated text is assertable without a container. It also lines the two packaging commands up structurally — `sdk::package` already had a function by that name.

Two tests: `test_generate_rpm_build_script_uses_rpm_form_version` asserts `Version: 1.0.0~rc.1` and `my-ext-1.0.0~rc.1-r0.noarch.rpm` are present *and* that neither semver-hyphen form is, so reverting either interpolation to `metadata.version` fails rather than passing green. `test_generate_rpm_build_script_release_version_unchanged` pins that a plain release version is untouched.

mobileoverlord:
@jetm this was a much better review than the PR deserved — the framing that the diff *fixes the publish side and leaves the consumers asking for the old name* is exactly right, and it is the thing I had missed. All four items are addressed in e4ceca7, along with the three inline comments.

**The core change first:** `to_rpm_version` now **rejects** a `-` inside a pre-release or build identifier instead of rewriting it. Your measurement is what moved me — I had it filed as lossy-but-harmless, and an ordering inversion is a different category, because nothing downstream can detect it. dnf just declines the upgrade silently. Since neither leaving the hyphen nor mapping it is correct, and no substitute character works, there is no mapping to choose; refusing the input is the only branch that can't be quietly wrong.

That also unblocked everything else: with hyphenated identifiers gone the map is injective, so `from_rpm_version` is an exact inverse, and items 2 and 4 became one call each.

**1. `ext_fetch.rs` — install side.** Fixed. The package spec now asks for the RPM form, since dnf matches against the RPM's own `Version:`. Agreed this is the common case rather than the corner — it is the fresh-checkout path, and the failure mode (layout probe and install both matching nothing) is silent-ish. The conversion is idempotent for an already-RPM-form version, so the lockfile branch you identified as already-correct is unaffected.

**2. `connect/ext.rs` — publish.** Fixed via the inverse: the NVR-derived version is mapped back to semver before it is POSTed, so the platform records `1.0.0-rc.1` and matches the payload's `avocado.yaml`. An explicit `--version` is still taken as given. Your note that this diff already *improves* that call is right, and it is why this was one line rather than a redesign.

**3. `sdk/package.rs` — sibling command.** Fixed. Every use of the version inside `generate_rpm_build_script` is an RPM filename or the spec's `Version:`, so it is the RPM form throughout: main NVR, sub-package NVRs, and the spec field. You were right that `to_rpm_version` living in `utils/version.rs` rather than beside its one caller was written for this — it needed no new plumbing beyond making the function fallible.

**4. `runtime/build.rs` — the heuristic.** Confirmed real, and I would raise your confidence: it is a **regression introduced by this PR**, not a pre-existing gap. Before, a pre-release RPM could not be built at all, so the rpmdb never held a `~` version and this path was unreachable. Making packaging work is what reaches it. Traced end to end:

- `query_rpm_version` returns `%{VERSION}` = `1.0.0~rc.1`
- `1.0.0~rc.1` has no `-`, so `rfind('-')` at :1247 lands on the name/version boundary and the remainder starts with a digit — where `rc.1` previously did not — so the extension is inserted into `ext_version_map`
- that selects the strict exact-name branch at :1311 with its `exit 1`, instead of the tolerant `ls {ext_name}-*.raw` glob at :1322
- and `ext/image.rs:566` names the `.raw` from the **config** version, so the artifact the strict branch demands does not exist

Fixed at the source rather than in the heuristic: `query_rpm_version` inverts back to semver, so its callers compose a name that matches what `ext image` actually produces. The heuristic is left alone — it was never wrong, it was being fed an RPM-form string.

**Doc note.** Added a subsection to `extension-versioning-redesign.md` recording the RPM `Version:` form as one confined exception: where it may appear, that it inverts exactly, and why hyphenated identifiers are rejected. Config stays the source of truth; the RPM form is a representation at one boundary, not a second version.

**On the `1.0.0+build-5` -> `1.0.0^build~5` wart** I had flagged earlier: it is gone as a side effect, since that input is now rejected. Dropping build metadata entirely is still arguably cleaner, but it is no longer a correctness question, so I have left it.

**CI.** The two red `Analyze` jobs were `Failed to resolve action download info. Error: Service Unavailable` from the Aug 6 Actions outage, and the real test workflows had been stuck `queued` for ~98h from the same outage. Cancelled those and pushed; fresh runs are going now. Locally: `cargo fmt --all --check`, `clippy --all-targets --all-features -D warnings`, and the full suite are clean. I also merged `main` in — the branch predated clippy fixes that had already landed there, so its own lint run would have failed on unrelated code.

jetm (approved):

PR #178, opened 2026-07-08
Reap abandoned build volumes and clarify stale-volume errors

jetm:
## Problem

Each project gets a per-project Docker volume (`avo-<uuid>`) recorded in its `.avocado-state` file. Deleting a project directory by hand drops the folder but not the volume, so `avo-*` volumes accumulate unbounded. Worse, a stale or half-populated volume makes `avocado build` fail during `runtime build dev` with a cryptic `grep: .../rootfs-work/etc/passwd: No such file`, naming neither the cause nor the fix.

## Solution

Reclaim provably-abandoned `avo-*` volumes at the moment a fresh volume is minted, and guard the rootfs build so a half-populated sysroot fails fast with an actionable message instead of the grep noise.

## Key changes

- Auto-reap abandoned `avo-*` volumes in `get_or_create_volume` (the natural GC point, and the only path already holding a container-tool handle, so `init` and the pure-filesystem commands stay docker-free). Best-effort: a daemon-down or held-volume failure is swallowed and never blocks a build.
- Share one classifier, `avo_abandonment_reason`, between the reap and `avocado prune` so the two paths cannot drift on what "abandoned" means.
- Guard the rootfs build: if `rootfs-work` lacks `/etc/passwd` after the sysroot copy, exit with a message pointing at `avocado clean` and `avocado prune`.
- Tests: 5 classifier cases plus generated-script assertions (guard present and ordered before user creation).

## Reviewer notes

- Behavior change: `install`/`build` now delete provably-dead `avo-*` volumes (source dir gone, no `.avocado-state`, or state pointing at a different volume). Active project volumes are never touched.
- Verified against the QEMU getting-started flow: a clean build still succeeds, and the guard fires with the actionable message on a simulated half-populated sysroot.
- Companion docs troubleshooting note is a separate peridio/docs PR.

### DIFF
--- a/src/utils/volume.rs
+++ b/src/utils/volume.rs
@@ -70,6 +70,41 @@ impl VolumeState {
     }
 }
 
+/// Decide whether an `avo-<uuid>` volume is abandoned, based solely on
+/// on-disk state (no container calls), given the `avocado.source_path`
+/// label recorded on the volume when it was created.
+///
+/// Returns `Some(reason)` when the volume is a stale leftover safe to
+/// reap, or `None` when it is still bound to a live project. The checks
+/// mirror the ones `avocado prune` applies so both paths agree on what
+/// "abandoned" means.
+pub fn avo_abandonment_reason(volume_name: &str, source_path: Option<&str>) -> Option<String> {
+    let source_path = match source_path {
+        Some(path) => path,
+        None => return Some("no source_path label found".to_string()),
+    };
+
+    let source_dir = Path::new(source_path);
+    if !source_dir.exists() {
+        return Some(format!("source directory '{source_path}' does not exist"));
+    }
+
+    let state_file = source_dir.join(".avocado-state");
+    if !state_file.exists() {
+        return Some(format!("no .avocado-state file in '{source_path}'"));
+    }
+
+    match VolumeState::load_from_dir(source_dir) {
+        Ok(Some(state)) if state.volume_name == volume_name => None,
+        Ok(Some(state)) => Some(format!(
+            ".avocado-state links to '{}', not this volume",
+            state.volume_name
+        )),
+        Ok(None) => Some(format!("could not read .avocado-state in '{source_path}'")),
+        Err(e) => Some(format!("error reading .avocado-state: {e}")),
+    }
+}
+
 /// Docker volume manager for Avocado operations
 pub struct VolumeManager {
     container_tool: String,
@@ -109,6 +144,20 @@ impl VolumeManager {
             }
         }
 
+        // We're about to mint a fresh volume for this project. That's the
+        // natural moment to sweep volumes left behind by projects deleted
+        // from disk without `avocado clean`: a manual `rm -rf` drops the
+        // project folder but not its per-project `avo-<uuid>` volume, so
+        // they otherwise accumulate forever. Best-effort by design — a
+        // container-tool hiccup must never block volume creation.
+        let reaped = self.reap_abandoned_avo_volumes().await;
+        if !reaped.is_empty() && self.verbose {
+            print_info(
+                &format!("Reaped {} abandoned build volume(s).", reaped.len()),
+                OutputLevel::Normal,
+            );
+        }
+
         // Create new volume state
         let state = VolumeState::new(source_dir.to_path_buf(), self.container_tool.clone());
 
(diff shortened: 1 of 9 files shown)


jetm:
@lee-reinhardt could you take a look at this one when you get a chance?

src/utils/volume.rs
```diff
@@ -128,6 +177,72 @@ impl VolumeManager {
...
+            .lines()
+            .filter(|line| line.starts_with("avo-"))
+            .map(|s| s.to_string())
+            .collect())
+    }
+
+    /// Read the `avocado.source_path` label off a volume. Returns `None`
+    /// when the volume is missing, unlabeled, or the inspect output
+    /// cannot be parsed.
+    async fn inspect_source_path_label(&self, volume_name: &str) -> Option<String> {
```
lee-reinhardt: error: `inspect_source_path_label` returns `None` both when the volume genuinely lacks the `avocado.source_path` label and when `docker volume inspect` itself fails (daemon hiccup, timeout, unparseable output). `avo_abandonment_reason(name, None)` classifies that as abandoned, so a transient inspect failure mid-loop deletes an active project's volume. `prune` does not have this hazard — `classify_avo_volume` propagates inspect errors with `?`. Best-effort should mean skip-on-failure, not reap-on-failure: return `Result<Option<String>>` (or similar tri-state) and `continue` the loop on `Err`, reserving the no-label reason for a successful inspect that genuinely has no label.

src/utils/volume.rs
```diff
@@ -109,6 +144,20 @@ impl VolumeManager {
             }
         }
 
+        // We're about to mint a fresh volume for this project. That's the
+        // natural moment to sweep volumes left behind by projects deleted
+        // from disk without `avocado clean`: a manual `rm -rf` drops the
+        // project folder but not its per-project `avo-<uuid>` volume, so
+        // they otherwise accumulate forever. Best-effort by design — a
+        // container-tool hiccup must never block volume creation.
+        let reaped = self.reap_abandoned_avo_volumes().await;
```
lee-reinhardt, line 153: problem: Cross-process race: another `avocado` process in a different project can be inside its own mint window — `create_volume` done, `save_to_dir` not yet — when this reap runs. That volume's label points at a source dir that exists but has no `.avocado-state` yet, so it classifies as abandoned and gets deleted out from under the concurrent first-time install. The window is small but real for CI matrices or parallel agent-driven builds. Two cheap mitigations: age-gate on the volume's `CreatedAt` from inspect (skip volumes younger than ~1h), or write the state file before `docker volume create` (the existing `volume_exists` check already tolerates a state file pointing at a not-yet-created volume).

src/utils/volume.rs
```diff
@@ -128,6 +177,72 @@ impl VolumeManager {
...
+        let names = match self.list_avo_volumes().await {
+            Ok(names) => names,
+            Err(_) => return Vec::new(),
+        };
+
+        let mut removed = Vec::new();
+        for name in names {
+            let source_path = self.inspect_source_path_label(&name).await;
+            if avo_abandonment_reason(&name, source_path.as_deref()).is_some()
+                && self.remove_volume(&name).await.is_ok()
```
lee-reinhardt: nit: Reaped volumes leave no trail — nothing is printed by default, and verbose prints only a count at the call site. `prune` prints `Removing '<name>': <reason>` per volume. At minimum verbose should print each reaped name and its abandonment reason so a surprising deletion can be diagnosed after the fact.

src/utils/volume.rs
```diff
@@ -128,6 +177,72 @@ impl VolumeManager {
         Ok(state)
     }
 
+    /// Best-effort removal of abandoned `avo-*` volumes left behind by
+    /// projects deleted from disk without `avocado clean`. Each `avo-*`
+    /// volume is classified with [`avo_abandonment_reason`] and removed
+    /// when abandoned. Any container-tool failure — daemon down, volume
+    /// still held by a container — is swallowed so this never blocks the
+    /// caller. Returns the names of the volumes actually removed.
+    pub async fn reap_abandoned_avo_volumes(&self) -> Vec<String> {
```
lee-reinhardt, line 194: doc: Suggestion/reminder — if changelog entries are expected alongside PRs rather than batched at release time, this deserves one: `install`/`build` now auto-delete provably-abandoned `avo-*` volumes, which is a user-visible behavior change. Ignore if the changelog is compiled at release.

lee-reinhardt (changes requested):

src/utils/volume.rs
jetm: Fixed in `e3acc63`. `inspect_source_path_label` now returns `Result<Option<String>>` — `Err` on an inspect/parse failure, `Ok(None)` only for a genuine no-label — and the reap `continue`s on `Err`. So a transient docker hiccup skips the volume instead of reaping an active one. Good catch, this was the real bug.

src/utils/volume.rs
jetm: Fixed in `e3acc63` via the write-state-before-create option. `get_or_create_volume` now persists `.avocado-state` before `docker volume create`, so a concurrent reap never sees the volume without its state pointer. `volume_exists` already tolerates a state file pointing at a not-yet-created volume, so a failed create is just superseded next call.

src/utils/volume.rs
jetm: Done in `e3acc63`. Verbose now prints `Reaped abandoned build volume '<name>': <reason>` per volume, matching `prune`. Verified it shows on `install --verbose`.

src/utils/volume.rs
jetm: Added in `e3acc63` under Changed — the auto-delete of abandoned `avo-*` volumes is flagged as a user-visible change. The repo does per-PR entries (#177 added one), so it belongs here.

jetm:
Pushed two commits: `e3acc63` addresses all four review points, and `2a11f58` fixes clippy lints that rust 1.97 started flagging repo-wide (redundant format refs, a `for (_, v)` map iter, a `match`→`?`). That second one is unrelated to this PR's feature, but `cargo clippy -D warnings` was failing the pre-commit hook and CI on untouched files, so I folded the mechanical fix in here to unblock. Happy to split it into its own PR if you'd rather keep this one focused.

jetm:
All four review points are addressed in `e3acc63` (per-thread replies above), and CI is green across the board — Test Suite, qemux86-64 build, Security Audit, Windows check, and CodeQL. Ready for another look whenever you have a moment.

lee-reinhardt (approved):

PR #179, opened 2026-07-17
Add --target-board flag for board interpolation

jetm:
## Problem

`{{ avocado.target.board }}` could only be set through the
`AVOCADO_TARGET_BOARD` env var or `avocado.yaml`. There was no CLI flag for
it, unlike `--target`.

## Solution

Add a per-subcommand `--target-board` flag that overrides the board when
resolving `{{ avocado.target.board }}`, giving it parity with `--target`. The
resolved value is parked once at the CLI boundary in a process-scoped override
that the interpolation resolver reads as its top tier, so every
config-composition path honors the flag without threading a parameter through
each site.

## Key changes

- `--target-board` on build, provision, runtime provision, ext build, rootfs
  install, initramfs install, sdk install, and install (per-subcommand, not a
  root global).
- Override read as the top tier of `resolve_target_board_value`, the single
  point all `AvocadoContext` construction funnels through, so the flag reaches
  extension discovery and the composed-config first pass, not just the
  directly-threaded callers.
- Precedence: `--target-board` > `AVOCADO_TARGET_BOARD` > runtime
  `target_board` > `default_target_board` > fallback to the resolved target.
- Unit tests for the precedence chain plus an end-to-end interpolation test.

## Scope

The flag is intentionally on the 8 user-facing build/install entry points
listed above. Sub-steps that `avocado build` orchestrates (`ext image`,
`runtime build`, `sdk run/compile/dnf`) inherit the board through the
process-scoped override, so their stamp hashes stay consistent in the
orchestrated flow; run standalone they resolve the board from env/config as
before. Extending the flag to those sub-commands for standalone parity is a
possible follow-up.

## Reviewer notes

Verified end to end on a real target:
`avocado sdk install --target imx8mp-var-dart --target-board variscite-sonata`
installs `avocado-bsp-variscite-sonata` from the `2024/next` feed (exit 0),
matching what `default_target_board: variscite-sonata` in `avocado.yaml`
produces. `cargo fmt --check`, `cargo clippy --all-targets --all-features --
-D warnings`, and the full `cargo test` suite are green.


src/commands/sdk/install.rs
```diff
@@ -96,6 +99,12 @@ impl SdkInstallCommand {
     }
 
     /// Set the no_stamps flag
+    /// Set the CLI target board override
```
nicksinas, line 101: Doc comment slip: `with_target_board` was inserted between `with_no_stamps`'s `/// Set the no_stamps flag` doc and its body. As a result `with_target_board` now carries a stale "Set the no_stamps flag" line above its own doc, and `with_no_stamps` (below) is left undocumented. This is the only command file where the doc landed in the wrong place — the others put it correctly. Suggest moving the `/// Set the no_stamps flag` line back down onto `with_no_stamps`.

src/utils/interpolation/avocado.rs
```diff
@@ -20,6 +20,38 @@
...
+/// config composition re-derives the board at roughly a dozen call sites
+/// (`parse_config_value_with_interpolation`, `discover_remote_extensions`,
+/// `load_composed`'s first pass, and siblings), several of which pass `None`.
+/// Threading a parameter to each is exactly the discipline that already failed —
+/// the flag was silently dropped on the `sdk install` / `build` extension paths.
+/// Instead the resolved flag is stored here and read as the top tier of
+/// [`AvocadoContext::resolve_target_board_value`], through which every
+/// production context is built. Env (`AVOCADO_TARGET_BOARD`) already reaches
+/// every site the same way; this mirrors that channel for the flag.
+static CLI_TARGET_BOARD_OVERRIDE: RwLock<Option<String>> = RwLock::new(None);
```
nicksinas, line 54: Architectural question (DRY): this process-global override and the explicitly-threaded `cli_target_board` parameter carry the *same* `--target-board` value. In production they're always equal — `park_target_board` sets this global to exactly the value it also threads. Since the global is read as tier 2 of `resolve_target_board_value`, it already reaches every interpolation site, including the untouched `load_composed` / `interpolate_config(..., None)` paths (the `None`-threaded calls fall through tier 1 to the global). That makes the new `*_with_board` variants, the `target_board` struct fields, and the `with_target_board`/`with_cli_target_board` builders largely redundant in production — the only place threading is genuinely load-bearing is the stamp-hash tests, which inject a board without touching the global.

Net effect: every future config path now has to keep *both* channels in sync. Could we either (a) collapse to global-only and keep threaded params solely where a test needs hermetic injection, or (b) add a short comment stating the rule for which channel to reach for and why both exist? Right now it reads as belt-and-suspenders without a documented boundary.

src/commands/ext/image.rs
```diff
@@ -21,6 +21,9 @@ pub struct ExtImageCommand {
     config_path: String,
     verbose: bool,
     target: Option<String>,
+    /// Target board override for `{{ avocado.target.board }}` (feeds the
+    /// stamp input hash so a board switch invalidates the image).
+    target_board: Option<String>,
```
nicksinas, line 26: Surface inconsistency: `ExtImageCommand` gains a `target_board` field that feeds the stamp input hash, but standalone `avocado ext image` has no `--target-board` flag (unlike its sibling `ext build`). Via `avocado build` this field gets a real value; via the standalone entry point it is always `None`, so this field is unreachable from its own CLI and the two commands' stamps diverge on board-dependent overlays. `runtime build` and the `sdk run/compile/dnf` family are in the same boat. If the 8-command scope is intentional, could you confirm it in the PR description? Otherwise `ext image` seems like it should match `ext build`.

src/utils/interpolation/avocado.rs
```diff
@@ -20,6 +20,38 @@
...
+
+/// Park (or clear with `None`) the process-scoped `--target-board` override.
+/// Called once at the CLI boundary before command dispatch; tests set and clear
+/// it under `#[serial]`.
+pub fn set_cli_target_board_override(board: Option<String>) {
+    if let Ok(mut guard) = CLI_TARGET_BOARD_OVERRIDE.write() {
```
nicksinas, line 60: Silent-failure smell: `set_cli_target_board_override` swallows a poisoned lock with `if let Ok(...)`, and `cli_target_board_override` does the same via `.read().ok()`. If the lock were ever poisoned, the flag would be silently dropped and the board would silently fall back to the target — the exact "flag silently ignored" failure this PR set out to fix. In practice the global is parked once, single-threaded, before any concurrency, so poisoning is effectively impossible — but I'd either `expect(...)` to fail loud, or add a one-line comment noting the park-once invariant so the silent swallow is a deliberate, documented choice rather than an accident.

src/utils/interpolation/avocado.rs
```diff
@@ -788,10 +845,92 @@ runtimes:
...
+"#,
+        );
+        let ctx = AvocadoContext::from_main_config(&config, None, None);
+        assert_eq!(ctx.target_board, Some("env-board".to_string()));
+        env::remove_var("AVOCADO_TARGET_BOARD");
+    }
+
+    #[test]
+    #[serial]
+    fn test_cli_override_reaches_none_threaded_interpolation() {
```
nicksinas, line 904: Latent test fragility: this global is shared across the whole unit-test binary, guarded only by the `#[serial]` convention. `#[serial]` only serializes tagged tests, so a serial test holding `CLI_TARGET_BOARD_OVERRIDE = "flag-board"` runs concurrently with the untagged `interpolate_config` tests. Today none of those untagged tests interpolate `{{ avocado.target.board }}`, so it's latent rather than live — but a future board-referencing test added without `#[serial]` could flake. Worth a comment on the static (or a test-helper that always sets+clears under serial) so the invariant is explicit. The set/clear-before-assert ordering in the existing serial tests is correct — nice.

src/main.rs
```diff
@@ -1855,6 +1870,19 @@ fn resolve_runtime_at_path(config_path: &str, cli_runtime: Option<&str>) -> Resu
...
 
+/// Park the CLI `--target-board` value in the process-scoped interpolation
+/// override, then return it unchanged for the usual per-command threading.
+///
+/// Wrapping each command's `--target-board` at the dispatch boundary guarantees
+/// the flag reaches the interpolation sites that compose config with a `None`
+/// board parameter (extension discovery, composed-config first pass), which the
+/// threaded parameter alone does not cover. Passing `None` clears any prior
+/// value, keeping resolution on the env/config chain.
+fn park_target_board(target_board: Option<String>) -> Option<String> {
```
nicksinas, line 1891: Maintainability note: correctness now hinges on every relevant command arm remembering to wrap its flag in `park_target_board(...)`. A future command that adds `--target-board` but constructs its command struct with a plain `target_board` (forgetting the wrapper) would thread the value yet leave the global unset — and the `None`-threaded interpolation sites would silently drop the flag again, reintroducing exactly the bug this PR fixes. There's no compile-time guard against that. Minor, but it's the flip side of the global design (see the DRY comment on `avocado.rs`). Also, `park_target_board` both mutates global state and returns its input unchanged — the dual role is a little surprising; the name and doc help, but worth a second look.

nicksinas (commented):
Review focused on DRY, maintainability/portability, correctness, and code smells. Overall the feature works and is well-tested; the notes below are a mix of one clean fix, one architectural question, and several smells worth a look.

src/commands/sdk/install.rs
jetm: Fixed in 6cc6584 - removed the stray `/// Set the no_stamps flag` line that had slipped above `with_target_board`, so its rustdoc no longer inherits the wrong label.

src/utils/interpolation/avocado.rs
jetm: Good catch, and you're right on the mechanics: in production the global and the threaded `cli_target_board` always carry the same value, and the global alone reaches every site including the `None`-threaded ones. I went with option (b) in a95f6a8 and documented the boundary on the override static: the global is the load-bearing production channel, and the threaded params are retained only so tests can inject a board hermetically (the stamp-hash tests) without mutating the global. Collapsing to global-only is a clean follow-up I'd rather land on its own so this PR stays scoped, but the rule for which channel to reach for is now stated in the code.

src/commands/ext/image.rs
jetm: The 8-command scope is intentional, and I've added a Scope note to the PR description to say so explicitly. In the normal flow the stamp divergence doesn't occur: `avocado build` parks the board in the process override before dispatch, so the orchestrated `ext image`, `runtime build`, and `sdk` steps all resolve the same board and hash consistently. Standalone invocation of those sub-steps resolves the board from env/config as before. Giving `ext image` its own `--target-board` for parity is a small follow-up if we want per-carrier stamps when driving it directly, but I didn't want to widen the surface here.

src/utils/interpolation/avocado.rs
jetm: Documented in a95f6a8. The override is parked exactly once, single-threaded, before any async work, so the write lock is never contended and cannot be poisoned; I added a comment on the static noting that invariant so the `Ok`/`ok()` guards read as a deliberate consequence of park-once rather than a silent hole. I kept the swallow over `expect` so a hypothetical future misuse degrades to env/config resolution instead of panicking mid-build, with the safety invariant now written down.

src/utils/interpolation/avocado.rs
jetm: Fair. I added a comment on the static spelling out the invariant: any test that sets the override must carry `#[serial]` and set+clear around its assertion, since the static is shared across the whole test binary. It's latent today as you noted, no untagged test interpolates the board; the comment is there so a future board-referencing test gets `#[serial]` instead of flaking. And thanks for confirming the set/clear-before-assert ordering.

src/main.rs
jetm: Right, this is the flip side of the global design. I added an INVARIANT note on `park_target_board` in a95f6a8: every arm that accepts `--target-board` must route its value through it, and an arm that assigns `target_board` directly would leave the global unset and silently drop the flag again. There's no compile-time guard, so the doc carries that contract for now; the collapse-to-global-only follow-up from the DRY thread would remove the failure mode entirely. On the dual role, I kept mutate-and-return so every arm goes through a single call and the doc flags the surprise, but I'm happy to split it into set-then-pass if you'd prefer it explicit.

nicksinas (approved):

PR #183, opened 2026-07-23
Provision and build declared device-tree overlays

jetm:
## Problem

The CLI had no support for the device-tree-overlay feature: it neither provisioned the SDK overlay tooling nor compiled and delivered overlays during a runtime build.

## Solution

When an extension declares `device_tree_overlays`, provision the SDK wrapper and per-BSP delivery hook during `sdk install`, then compile and deliver each declared overlay during `avocado build`.

## Key changes

- `sdk/install`: provision device-tree-overlay tooling (`nativesdk-avocado-dtc-overlay` plus the delivery-hook package) from extension declarations
- `runtime/build`: compile and deliver declared device-tree overlays into the OS bundle
- `runtime/build`: search the SDK stone dir for BSP boot artifacts - see the note below, this commit's motivation changed after it was written

## Reviewer notes

Top of the device-tree-overlay stack (ENG-2134). Both dependencies are now merged - stone#28 and meta-avocado#245 - so this no longer waits on anything to be reviewable.

### Why `aa2d49a` is in this PR

It was written for a Raspberry Pi failure, `File 'u-boot.bin' not found in any input directory for FAT image`, on the theory that nothing staged u-boot into stone's input dir. **That theory was wrong.** A real rpi5 build showed `avocado-img-bootfiles` already ships `u-boot.bin` and `bootfiles/*`, reachable through `avocado-runtime`'s RDEPENDS; the project that failed simply had not declared `avocado-runtime`. meta-avocado#274 chased the same wrong premise from the BSP side and is closed for that reason.

The commit stays because a different consumer turned up and this time the gap is real. On Tegra the base DTB genuinely is absent from the runtime input dir - the `.dtb` files under `runtimes/<rt>/devicetree/` are AGX Orin parts, and only `tegraflash-bsp` carries the Orin Nano one - and the overlay delivery hook has to read it to merge overlays into it. meta-avocado#292 stages that BSP into the SDK stone dir in its first commit (`92eb079`); this commit is the half that reads it. The `runtime/build.rs` comment describing itself as "the consumer half of a two-part change" is accurate, and #292 is the other half.

Merge order is not a constraint: without #292 the directory holds nothing new, so the search is inert rather than broken.

### Verification

- **raspberrypi5**: build path end to end - wrapper compiles the `.dtbo`, hook delivers and claims it, `os-bundle.aos` finalizes. Re-run green after the Jetson work as a regression gate.
- **Jetson Orin Nano Developer Kit (P3767-0005), on-device**: paired positive and negative. With the overlay declared, `/proc/device-tree/hello-overlay/avocado,marker` reads the declared value on the running board; with the declaration removed and nothing else changed, the node is absent. Requires meta-avocado#292.

The Jetson result is worth one line of emphasis for a reviewer of this PR specifically: every build-path assertion this code emits passed for two full rounds while the board booted with no overlay applied. The unclaimed-overlay guard here is what makes a *missing* delivery loud; it cannot detect a delivery that lands in the wrong place. Only reading the running kernel's device tree did.

### Known gap, not introduced here

`avocado runtime provision` cannot flash a Jetson from its container. The production path passes no `--privileged` - the occurrence in `provision.rs` is inside `#[cfg(test)]` - so libusb fails at `could not retrieve chip ID`. With `--privileged` added by hand, signing and RCM boot report success but the target never boots the pushed initrd or exports its LUN. The identical sequence run host-side succeeds every time, and every Jetson flash behind the verification above was host-side.

This is pre-existing on `main` and out of scope here, but it means the documented CLI provisioning path is not the path the Jetson result was obtained through. Tracked separately.

### DIFF
--- a/src/commands/runtime/build.rs
+++ b/src/commands/runtime/build.rs
@@ -1232,6 +1232,14 @@ cp /opt/src/.tuf-staging-tmp/delegations/runtime-{runtime_uuid}.json \
                 )
             })?;
 
+        // Device-tree overlays declared by this runtime's extensions (ENG-2134).
+        // Empty section - and no build-time work - unless at least one is declared.
+        let device_tree_overlay_section = {
+            let overlays =
+                crate::utils::device_tree_overlay::collect_for_runtime(&merged_runtime, parsed)?;
+            crate::utils::device_tree_overlay::render_build_section(&overlays)?
+        };
+
         // Extract extension names from the `extensions` array
         let mut required_extensions = HashSet::new();
         let _extension_type_overrides: HashMap<String, Vec<String>> = HashMap::new();
@@ -2445,6 +2453,8 @@ if [ -n "${{AVOCADO_STONE_INCLUDE_PATHS:-}}" ]; then
 fi
 STONE_INCLUDE_FLAGS="$STONE_INCLUDE_FLAGS -i $STONE_INPUT_DIR"
 
+STONE_OVERLAY_FLAG=""
+{device_tree_overlay_section}
 echo -e "\033[94m[INFO]\033[0m Running stone bundle."
 echo -e "  Manifest:  $STONE_MANIFEST"
 echo -e "  Output:    $STONE_AOS_OUTPUT"
@@ -2461,6 +2471,7 @@ stone bundle \
     $STONE_INITRD_FLAG \
     -m "$STONE_MANIFEST" \
     $STONE_INCLUDE_FLAGS \
+    $STONE_OVERLAY_FLAG \
     --partition-size "var=$FINAL_SIZE" \
     -o "$STONE_AOS_OUTPUT" \
     --build-dir "$STONE_BUILD_DIR"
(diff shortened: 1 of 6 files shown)


jetm:
Producer half is up: avocado-linux/meta-avocado#274 stages `u-boot.bin` and `bootfiles/` into the SDK stone dir, which is the directory this PR adds as a stone `-i` input.

Either half alone is inert and neither regresses a current build (stone resolves first-match-wins across `-i` dirs, so until both land the added directory carries nothing new), so they can merge in either order.

mobileoverlord (approved):
Verification is the model kind: rpi5 end-to-end plus paired positive/negative on-device Jetson reads of /proc/device-tree, with the limits of the build-path assertions stated plainly. The aa2d49a retention rationale holds now that meta-avocado#292 (merged today) is the producing half. Rebased onto current main (the stale CI failures were infra-level; 3361 tests pass locally) — merging on fresh green.

PR #185, opened 2026-07-27
Stop reinstalling rootfs and initramfs on every sdk install

mobileoverlord:
## Problem

`avocado sdk install` redid the rootfs and initramfs installs in full on
every invocation, even when nothing had changed. Two independent defects
caused it.

**1. False "packages removed" detection wiped both sysroots every run.**
`detect_sysroot_package_removals` compared *config-declared* package
names against *locked* names and returned true — meaning `rm -rf
$AVOCADO_PREFIX/<sysroot>` plus a from-scratch reinstall — whenever the
lock held names the config didn't.

But the install auto-appends two packages the config never declares, then
records them in the lock:

- `packagegroup-avocado-{rootfs,initramfs}-modules-<kver>`
- `kernel-image-<kver>` (rootfs only)

From the second run onward those always read as removed. Because
`get_rootfs_packages` / `get_initramfs_packages` default to the single
meta-package, this fired on *every* project, not only ones overriding
`rootfs.packages`. Confirmed against every real lockfile in the tree —
for example `references/jetson-trt/avocado.lock`:

| | locked names | config names |
|---|---|---|
| rootfs | `avocado-pkg-rootfs`, `kernel-image-6.8.12-…`, `packagegroup-avocado-rootfs-modules-6.8.12-…` | `avocado-pkg-rootfs` |
| initramfs | `avocado-pkg-initramfs`, `packagegroup-avocado-initramfs-modules-6.8.12-…` | `avocado-pkg-initramfs` |

**Second-order effect:** the same path called
`remove_packages_from_sysroot`, dropping those packages' version pins
from the in-memory lock. `build_package_spec_with_lock` then fell back to
a bare package name, so dnf resolved the kernel image and module
packagegroup to *newest available* instead of the pinned NVR.
`avocado.lock` looked pinned on disk (it is rewritten post-install) but
the pin never applied.

**2. The install stamps were written but never read.** `install_sysroot`
wrote `rootfs/install.stamp` and `initramfs/install.stamp`, but nothing
consulted them to skip work — across the whole CLI, stamps were only
prerequisite gates. So even with defect 1 fixed, each run still paid a
kernel repoquery, off-kernel exclude computation, a repoquery existence
probe, the dnf transaction, `query_installed_packages`, kernel-sysroot
staging, and the stamp write.

## Solution

An unchanged config now does zero container work for rootfs/initramfs; a
changed one reinstalls exactly the sysroot(s) affected; `avocado.lock`
pins actually bind.

## Key changes

- **Compare removals against the effective set.** `install_sysroot`
resolves the kernel and computes the auto-appended names *before* removal
detection, and `detect_sysroot_package_removals` now takes the effective
name set (config packages ∪ auto-appends) and returns the removed names
rather than a bool. That set is built once and reused as the post-install
lock query list, replacing a second, duplicate computation. A kver change
falls out of the same mechanism, so the dedicated kernel-pin-change block
became the *reason* for the clean rather than a second code path, and the
two verbatim inline `rm -rf` RunConfig blocks collapse into one
`clean_sysroot` helper built from the existing `clean_sysroot_command`.

- **Skip the install when the stamp is current.** New
`read_stamps_batch` / `StampBatch` in `utils/prerequisites.rs` wraps
`generate_batch_read_stamps_script` + one container run + parsing. `sdk
install` reads both sysroot stamps in a single invocation before the
parallel phase and passes them in via `SysrootInstallParams`; the
standalone `rootfs install` / `initramfs install` read their own through
the same helper. `install_sysroot` short-circuits before any container
call. Stamps live in the `/opt/_avocado` named volume rather than a host
bind mount, so a container run is unavoidable — hence batching.
`check_prerequisites` and five sites that hand-rolled the same RunConfig
(`build.rs`, `sdk/{clean,package}.rs`, `ext/{clean,checkout}.rs`) are
ported to the helper.

- **Make the hash trustworthy enough to skip on.** The two per-sysroot
hash functions collapse into one core that now hashes the *effective*
package map (so an absent `rootfs.packages` and an explicit default hash
identically), plus `sdk.repo_url`, `sdk.repo_release` — the hook that
makes a snapshot bump or `avocado update` land — and
`sdk.disable_weak_dependencies`. `StampInputs.package_list_hash` is
populated with a digest of that sysroot's lockfile pins, so a hand-edited
`avocado.lock` or an `avocado unlock` goes stale automatically. An empty
pin set hashes to its own value rather than to `None`, because
`Stamp::is_current` only compares that field when both sides carry one —
otherwise a cleared lock would compare equal by omission.
`STAMP_VERSION` is deliberately *not* bumped: that would invalidate every
SDK/ext/runtime stamp. Changing only these two hash bodies makes existing
rootfs/initramfs stamps mismatch naturally, costing one reinstall.

- **Make `clean` invalidate its stamp.** `clean_sysroot_command` also
removes `$AVOCADO_PREFIX/.stamps/<sysroot>`, mirroring `ext clean`.
Without it, `rootfs clean && sdk install` would skip against a current
stamp and leave an empty sysroot.

- **Drop the mid-flight lockfile saves.** `install_sysroot` and
`stage_kernel_sysroot_from_rootfs` saved their lockfile *clone* while the
rootfs and initramfs tasks ran concurrently on separate clones —
last-writer-wins, healed only by the caller's final save. Saving is now
the caller's job throughout, which removes the race.

Escape hatches are unchanged in spirit: `--no-stamps` disables the read
along with the write, so it always reinstalls, and `avocado rootfs clean`
remains the hard reset.

## Reviewer notes

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the
full `cargo test` suite are green (1124 lib tests plus the integration
suites, no failures).

New unit tests:

- `detect_sysroot_package_removals` reports nothing for the exact
`references/jetson-trt/avocado.lock` shape — a default config with both
auto-appended kernel packages locked. This is the regression test for the
bug.
- It still reports a genuinely config-dropped package, and reports the
previous kver's auto-appends as stale after a repin.
- The rootfs hash is stable across absent-vs-explicit-default
`rootfs.packages`, and moves on an added package, on `repo_url`,
`repo_release` and `disable_weak_dependencies`, and on a lock re-pin.
- `package_list_hash` is order-independent (pins come out of a `HashMap`),
and an `avocado unlock` invalidates `is_current`.
- A rootfs-only config edit does not invalidate the initramfs hash.

End-to-end verification against a live feed is in progress separately.

### DIFF
--- a/src/commands/initramfs/install.rs
+++ b/src/commands/initramfs/install.rs
@@ -145,37 +145,48 @@ impl InitramfsInstallCommand {
             },
             runs_on_context.as_ref(),
         )
-        .await?;
-
-        let result = install_sysroot(&mut SysrootInstallParams {
-            sysroot_type: SysrootType::Initramfs,
-            config,
-            lock_file: &mut lock_file,
-            src_dir,
-            container_helper: &container_helper,
-            container_image,
-            target: &target,
-            target_board: self.target_board.as_deref(),
-            repo_url: repo_url.as_deref(),
-            repo_release: repo_release.as_deref(),
-            merged_container_args: merged_container_args.clone(),
-            dnf_args: self.dnf_args.clone(),
-            verbose: self.verbose,
-            force: self.force,
-            runs_on_context: runs_on_context.as_ref(),
-            sdk_arch: self.sdk_arch.as_ref(),
-            no_stamps: self.no_stamps,
-            parsed: Some(&composed.merged_value),
-            prefetched_stamp,
-            tui_context: None,
-        })
         .await;
 
+        // Not `?`: the teardown block below is the only thing that tears down the
+        // `runs_on` NFS server and remote mount, and an early return here would
+        // skip it — the next `--runs-on` run picks a fresh nfs_port and stacks
+        // another one on top. Carried into `result` and returned past teardown.
+        let result = match prefetched_stamp {
+            Err(e) => Err(e),
+            Ok(prefetched_stamp) => {
+                install_sysroot(&mut SysrootInstallParams {
+                    sysroot_type: SysrootType::Initramfs,
+                    config,
+                    lock_file: &mut lock_file,
+                    src_dir,
+                    container_helper: &container_helper,
+                    container_image,
+                    target: &target,
+                    target_board: self.target_board.as_deref(),
+                    repo_url: repo_url.as_deref(),
+                    repo_release: repo_release.as_deref(),
+                    merged_container_args: merged_container_args.clone(),
+                    dnf_args: self.dnf_args.clone(),
+                    verbose: self.verbose,
+                    force: self.force,
+                    runs_on_context: runs_on_context.as_ref(),
+                    sdk_arch: self.sdk_arch.as_ref(),
+                    no_stamps: self.no_stamps,
+                    parsed: Some(&composed.merged_value),
+                    prefetched_stamp,
+                    tui_context: None,
+                })
+                .await
+            }
+        };
+
         // Persist the lockfile the install updated — `install_sysroot` leaves
-        // saving to its caller.
-        if result.is_ok() {
-            lock_file.save(src_dir)?;
-        }
+        // saving to its caller. Folded into `result` rather than `?` so a save
+        // failure still reaches the teardown below.
+        let result = match result {
+            Ok(()) => lock_file.save(src_dir),
+            Err(e) => Err(e),
+        };
 
         if let Some(ref mut context) = runs_on_context {
             if let Err(e) = context.teardown().await {
--- a/src/commands/initramfs/install.rs
+++ b/src/commands/initramfs/install.rs
@@ -5,14 +5,16 @@ use std::sync::Arc;
 
 use crate::utils::{
     config::{ComposedConfig, Config},
-    container::SdkContainer,
+    container::{RunConfig, SdkContainer},
     lockfile::{LockFile, SysrootType},
     output::{print_error, OutputLevel},
     runs_on::RunsOnContext,
     target::validate_and_log_target,
 };
 
-use crate::commands::rootfs::install::{install_sysroot, SysrootInstallParams};
+use crate::commands::rootfs::install::{
+    install_sysroot, read_sysroot_install_stamp, SysrootInstallParams,
+};
 
 /// Implementation of the 'initramfs install' command.
 pub struct InitramfsInstallCommand {
@@ -128,6 +130,24 @@ impl InitramfsInstallCommand {
             .unwrap_or(std::path::Path::new("."));
         let mut lock_file = LockFile::load(src_dir)?;
 
+        let prefetched_stamp = read_sysroot_install_stamp(
+            &SysrootType::Initramfs,
+            self.no_stamps,
+            &container_helper,
+            RunConfig {
+                container_image: container_image.to_string(),
+                target: target.to_string(),
+                verbose: self.verbose,
+                repo_url: repo_url.clone(),
+                repo_release: repo_release.clone(),
+                container_args: merged_container_args.clone(),
+                sdk_arch: self.sdk_arch.clone(),
+                ..Default::default()
+            },
+            runs_on_context.as_ref(),
+        )
+        .await?;
+
         let result = install_sysroot(&mut SysrootInstallParams {
             sysroot_type: SysrootType::Initramfs,
             config,
@@ -147,9 +167,16 @@ impl InitramfsInstallCommand {
             sdk_arch: self.sdk_arch.as_ref(),
             no_stamps: self.no_stamps,
             parsed: Some(&composed.merged_value),
+            prefetched_stamp,
             tui_context: None,
         })
         .await;
+
+        // Persist the lockfile the install updated — `install_sysroot` leaves
+        // saving to its caller.
+        if result.is_ok() {
+            lock_file.save(src_dir)?;
+        }
 
         if let Some(ref mut context) = runs_on_context {
             if let Err(e) = context.teardown().await {
--- a/src/commands/initramfs/install.rs
+++ b/src/commands/initramfs/install.rs
@@ -137,7 +137,6 @@ impl InitramfsInstallCommand {
             RunConfig {
                 container_image: container_image.to_string(),
                 target: target.to_string(),
-                verbose: self.verbose,
                 repo_url: repo_url.clone(),
                 repo_release: repo_release.clone(),
                 container_args: merged_container_args.clone(),
--- a/src/commands/rootfs/install.rs
+++ b/src/commands/rootfs/install.rs
@@ -380,6 +380,7 @@ fn compute_install_stamp_inputs(
         repo_url: params.repo_url,
         repo_release: params.repo_release,
         disable_weak_dependencies: params.config.get_sdk_disable_weak_dependencies(),
+        dnf_args: params.dnf_args.as_deref(),
         locked_packages: params
             .lock_file
             .get_sysroot_versions(params.target, &params.sysroot_type),
@@ -1132,38 +1133,49 @@ impl RootfsInstallCommand {
             },
             runs_on_context.as_ref(),
         )
-        .await?;
-
-        let result = install_sysroot(&mut SysrootInstallParams {
-            sysroot_type: SysrootType::Rootfs,
-            config,
-            lock_file: &mut lock_file,
-            src_dir,
-            container_helper: &container_helper,
-            container_image,
-            target: &target,
-            target_board: self.target_board.as_deref(),
-            repo_url: repo_url.as_deref(),
-            repo_release: repo_release.as_deref(),
-            merged_container_args: merged_container_args.clone(),
-            dnf_args: self.dnf_args.clone(),
-            verbose: self.verbose,
-            force: self.force,
-            runs_on_context: runs_on_context.as_ref(),
-            sdk_arch: self.sdk_arch.as_ref(),
-            no_stamps: self.no_stamps,
-            parsed: Some(&composed.merged_value),
-            prefetched_stamp,
-            tui_context: None,
-        })
         .await;
+
+        // Not `?`: the teardown block below is the only thing that tears down the
+        // `runs_on` NFS server and remote mount, and an early return here would
+        // skip it — the next `--runs-on` run picks a fresh nfs_port and stacks
+        // another one on top. Carried into `result` and returned past teardown.
+        let result = match prefetched_stamp {
+            Err(e) => Err(e),
+            Ok(prefetched_stamp) => {
+                install_sysroot(&mut SysrootInstallParams {
+                    sysroot_type: SysrootType::Rootfs,
+                    config,
+                    lock_file: &mut lock_file,
+                    src_dir,
+                    container_helper: &container_helper,
+                    container_image,
+                    target: &target,
+                    target_board: self.target_board.as_deref(),
+                    repo_url: repo_url.as_deref(),
+                    repo_release: repo_release.as_deref(),
+                    merged_container_args: merged_container_args.clone(),
+                    dnf_args: self.dnf_args.clone(),
+                    verbose: self.verbose,
+                    force: self.force,
+                    runs_on_context: runs_on_context.as_ref(),
+                    sdk_arch: self.sdk_arch.as_ref(),
+                    no_stamps: self.no_stamps,
+                    parsed: Some(&composed.merged_value),
+                    prefetched_stamp,
+                    tui_context: None,
+                })
+                .await
+            }
+        };
 
         // Persist the lockfile the install updated. `install_sysroot` no
         // longer saves for itself — under `avocado sdk install` it runs on a
-        // clone that the caller merges and saves once.
-        if result.is_ok() {
-            lock_file.save(src_dir)?;
-        }
+        // clone that the caller merges and saves once. Folded into `result`
+        // rather than `?` so a save failure still reaches the teardown below.
+        let result = match result {
+            Ok(()) => lock_file.save(src_dir),
+            Err(e) => Err(e),
+        };
 
         // Always teardown runs_on context
         if let Some(ref mut context) = runs_on_context {
@@ -1312,6 +1324,32 @@ mod tests {
         assert!(
             detect_sysroot_package_removals(&effective, &SysrootType::Rootfs, TARGET, &lock)
                 .is_empty()
+        );
+    }
+
+    /// The empty-lock case above passes for the wrong reason — it early-returns
+    /// before comparing anything, so it survives inverting the `difference`
+    /// operands. This one seeds the lock so the comparison actually runs, and
+    /// pins the direction: locked-minus-effective, never the reverse.
+    #[test]
+    fn removal_detection_compares_locked_against_effective_not_the_reverse() {
+        let lock = lock_with(&SysrootType::Rootfs, &["avocado-pkg-rootfs", "vim"]);
+
+        // `vim` dropped from config: locked but no longer effective => removed.
+        let effective = name_set(&["avocado-pkg-rootfs"]);
+        assert_eq!(
+            detect_sysroot_package_removals(&effective, &SysrootType::Rootfs, TARGET, &lock),
+            vec!["vim".to_string()],
+        );
+
+        // The mirror image: a package that is effective but not yet locked is an
+        // *addition*, and must never be reported. Inverting the operands makes
+        // this arm return `curl` and fail.
+        let effective = name_set(&["avocado-pkg-rootfs", "vim", "curl"]);
+        assert!(
+            detect_sysroot_package_removals(&effective, &SysrootType::Rootfs, TARGET, &lock)
+                .is_empty(),
+            "a newly added package must not read as a removal",
         );
     }
 
--- a/src/commands/rootfs/install.rs
+++ b/src/commands/rootfs/install.rs
@@ -325,7 +325,11 @@ fn detect_sysroot_package_removals(
 /// Best-effort by design, matching the inline copies this replaced: an
 /// absent sysroot is the normal case on a first install and must not fail
 /// the run.
-async fn clean_sysroot(params: &SysrootInstallParams<'_>, sysroot_dir: &str) {
+/// Returns whether the wipe actually succeeded. The result used to be discarded
+/// outright; it is reported now because a failed wipe means the reinstall lands
+/// *alongside* the old contents (dnf install is additive), and writing a current
+/// stamp over that latches the mixed state forever.
+async fn clean_sysroot(params: &SysrootInstallParams<'_>, sysroot_dir: &str) -> bool {
     let clean_config = RunConfig {
         container_image: params.container_image.to_string(),
         target: params.target.to_string(),
@@ -341,18 +345,31 @@ async fn clean_sysroot(params: &SysrootInstallParams<'_>, sysroot_dir: &str) {
         ..Default::default()
     };
 
-    if let Some(context) = params.runs_on_context {
+    let outcome = if let Some(context) = params.runs_on_context {
         params
             .container_helper
             .run_in_container_with_context(&clean_config, context)
             .await
-            .ok();
     } else {
-        params
-            .container_helper
-            .run_in_container(clean_config)
-            .await
-            .ok();
+        params.container_helper.run_in_container(clean_config).await
+    };
+
+    match outcome {
+        Ok(true) => true,
+        Ok(false) => {
+            print_error(
+                &format!("Failed to clean the {sysroot_dir} sysroot before reinstalling."),
+                OutputLevel::Normal,
+            );
+            false
+        }
+        Err(e) => {
+            print_error(
+                &format!("Failed to clean the {sysroot_dir} sysroot before reinstalling: {e}"),
+                OutputLevel::Normal,
+            );
+            false
+        }
     }
 }
 
@@ -466,6 +483,67 @@ fn parse_probe_output(text: &str) -> Option<bool> {
     let idx = text.lines().position(|l| l.trim() == "AVOCADO_PROBE_OK")?;
     let names = text.lines().skip(idx + 1).collect::<Vec<_>>().join("\n");
     Some(!names.trim().is_empty())
+}
+
+/// Write this sysroot's install stamp.
+///
+/// Split out of [`install_sysroot`] so its failures can be reported without
+/// `?`-ing out of the caller: the install has already landed and its lock pins
+/// are recorded in memory by this point, and the caller only persists them when
+/// `install_sysroot` returns `Ok`.
+async fn write_install_stamp(
+    params: &SysrootInstallParams<'_>,
+    packages: &HashMap<String, serde_yaml::Value>,
+    label: &str,
+) -> Result<()> {
+    let Some(inputs) = compute_install_stamp_inputs(params, packages)? else {
+        return Ok(());
+    };
+
+    let stamp = match params.sysroot_type {
+        SysrootType::Rootfs => {
+            Stamp::rootfs_install(params.target, inputs, StampOutputs::default())
+        }
+        SysrootType::Initramfs => {
+            Stamp::initramfs_install(params.target, inputs, StampOutputs::default())
+        }
+        _ => unreachable!("sysroot type was validated at entry"),
+    };
+
+    let stamp_config = RunConfig {
+        container_image: params.container_image.to_string(),
+        target: params.target.to_string(),
+        command: generate_write_stamp_script(&stamp)?,
+        verbose: params.verbose,
+        source_environment: true,
+        interactive: false,
+        repo_url: params.repo_url.map(|s| s.to_string()),
+        repo_release: params.repo_release.map(|s| s.to_string()),
+        container_args: params.merged_container_args.clone(),
+        sdk_arch: params.sdk_arch.cloned(),
+        tui_context: params.tui_context.clone(),
+        ..Default::default()
+    };
+
+    if let Some(context) = params.runs_on_context {
+        params
+            .container_helper
+            .run_in_container_with_context(&stamp_config, context)
+            .await?;
+    } else {
+        params
+            .container_helper
+            .run_in_container(stamp_config)
+            .await?;
+    }
+
+    if params.verbose {
+        print_info(
+            &format!("Wrote install stamp for {label}."),
+            OutputLevel::Normal,
+        );
+    }
+    Ok(())
 }
 
 /// Install a sysroot (rootfs or initramfs) via DNF into the SDK container volume.
@@ -746,9 +824,14 @@ pub async fn install_sysroot(params: &mut SysrootInstallParams<'_>) -> Result<()
         );
     }
 
-    if needs_clean_reinstall {
-        clean_sysroot(params, sysroot_dir).await;
-    }
+    // A failed wipe is not fatal — the reinstall below still repairs the common
+    // case — but it must not be latched by a current stamp, since dnf installs
+    // additively and the old contents (an old kernel's modules, say) survive.
+    let clean_ok = if needs_clean_reinstall {
+        clean_sysroot(params, sysroot_dir).await
+    } else {
+        true
+    };
 
     let mut pkg_specs: Vec<String> = if packages.is_empty() {
         vec![build_package_spec_with_lock(
@@ -928,6 +1011,7 @@ $DNF_SDK_HOST $DNF_SDK_TARGET_REPO_CONF \
         // date" and the user is wedged until they guess --no-stamps.
         let versions_recorded = !installed_versions.is_empty();
         let mut kernel_staging_ok = true;
+        let install_is_clean = clean_ok;
 
         if versions_recorded {
             params.lock_file.update_sysroot_versions(
@@ -1001,12 +1085,16 @@ $DNF_SDK_HOST $DNF_SDK_TARGET_REPO_CONF \
         // derived: the install just re-pinned this sysroot's packages, and the
         // stamp has to record the lock state the *next* run will compare
         // against.
-        if !params.no_stamps && !(versions_recorded && kernel_staging_ok) {
+        let stamp_is_trustworthy = versions_recorded && kernel_staging_ok && install_is_clean;
+
+        if !params.no_stamps && !stamp_is_trustworthy {
             print_warning(
                 &format!(
                     "Not recording an install stamp for {label}: {}. \
                      The next run will reinstall rather than report it up to date.",
-                    if !versions_recorded {
+                    if !install_is_clean {
+                        "the sysroot could not be cleaned first, so stale contents may remain"
+                    } else if !versions_recorded {
                         "the installed package versions could not be read"
                     } else {
                         "kernel sysroot staging failed"
@@ -1016,51 +1104,21 @@ $DNF_SDK_HOST $DNF_SDK_TARGET_REPO_CONF \
             );
         }
 
-        if !params.no_stamps && versions_recorded && kernel_staging_ok {
-            if let Some(inputs) = compute_install_stamp_inputs(params, &packages)? {
-                let stamp = match params.sysroot_type {
-                    SysrootType::Rootfs => {
-                        Stamp::rootfs_install(params.target, inputs, StampOutputs::default())
-                    }
-                    SysrootType::Initramfs => {
-                        Stamp::initramfs_install(params.target, inputs, StampOutputs::default())
-                    }
-                    _ => unreachable!("sysroot type was validated at entry"),
-                };
-
-                let stamp_config = RunConfig {
-                    container_image: params.container_image.to_string(),
-                    target: params.target.to_string(),
-                    command: generate_write_stamp_script(&stamp)?,
-                    verbose: params.verbose,
-                    source_environment: true,
-                    interactive: false,
-                    repo_url: params.repo_url.map(|s| s.to_string()),
-                    repo_release: params.repo_release.map(|s| s.to_string()),
-                    container_args: params.merged_container_args.clone(),
-                    sdk_arch: params.sdk_arch.cloned(),
-                    tui_context: params.tui_context.clone(),
-                    ..Default::default()
-                };
-
-                if let Some(context) = params.runs_on_context {
-                    params
-                        .container_helper
-                        .run_in_container_with_context(&stamp_config, context)
-                        .await?;
-                } else {
-                    params
-                        .container_helper
-                        .run_in_container(stamp_config)
-                        .await?;
-                }
-
-                if params.verbose {
-                    print_info(
-                        &format!("Wrote install stamp for {label}."),
-                        OutputLevel::Normal,
-                    );
-                }
+        // Deliberately not `?` anywhere below. The pins recorded above are only
+        // persisted by the caller when this function returns Ok, so propagating a
+        // stamp-write failure would discard the pins for an install that already
+        // landed — the next run would re-resolve the kernel against feed head with
+        // no prev_pinned_kver to compare against and install additively on top.
+        // A missing stamp is the benign outcome: the next run reinstalls.
+        if !params.no_stamps && stamp_is_trustworthy {
+            if let Err(e) = write_install_stamp(params, &packages, label).await {
+                print_warning(
+                    &format!(
+                        "Installed {label} sysroot but could not record its install stamp: {e}. \
+                         The next run will reinstall rather than report it up to date."
+                    ),
+                    OutputLevel::Normal,
+                );
             }
         }
     } else {
--- a/src/commands/rootfs/install.rs
+++ b/src/commands/rootfs/install.rs
@@ -1,7 +1,7 @@
 //! Rootfs sysroot install command and shared install logic for rootfs/initramfs.
 
 use anyhow::{Context, Result};
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::path::Path;
 use std::sync::Arc;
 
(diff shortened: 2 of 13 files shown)


src/commands/sdk/install.rs
```diff
@@ -414,6 +415,40 @@ $DNF_SDK_HOST $DNF_NO_SCRIPTS $DNF_SDK_TARGET_REPO_CONF \
...
+        } else {
+            let rootfs_req = StampRequirement::rootfs_install();
+            let initramfs_req = StampRequirement::initramfs_install();
+            let batch = read_stamps_batch(
+                &[rootfs_req.clone(), initramfs_req.clone()],
+                container_helper,
+                RunConfig {
+                    container_image: container_image.to_string(),
+                    target: target.to_string(),
+                    verbose: self.verbose,
```
nicksinas: **Nit (consistency):** this stamp-read `RunConfig` sets `verbose: self.verbose`, which flips `AVOCADO_VERBOSE=1` for what is just an `echo … | tr` script. It's harmless — the output is captured (not echoed on success) and `parse_batch_stamps_output` ignores any non-`:::` lines — but it's inconsistent with the other reads now routed through `read_stamps_batch` (`check_prerequisites`, `sdk clean`, `sdk package`, `ext/*`), which leave `verbose` at its `false` default. Same applies to the `read_sysroot_install_stamp` config in the standalone `rootfs`/`initramfs` paths. Suggest dropping it here for consistency.

src/commands/rootfs/install.rs
```diff
@@ -347,57 +460,38 @@ pub async fn install_sysroot(params: &mut SysrootInstallParams<'_>) -> Result<()
...
 
+    // Short-circuit: nothing to do when the stamp on record still matches the
+    // current inputs. This is checked before any container call, so an
+    // unchanged project pays nothing here — no kernel repoquery, no dnf
+    // transaction, no lock rewrite.
+    //
+    // Escape hatches: `--no-stamps` skips the read (and the write) so it
+    // always reinstalls, and `avocado {rootfs,initramfs} clean` removes the
+    // stamp along with the sysroot, so a cleaned sysroot never skips.
+    if !params.no_stamps {
```
nicksinas, line 479: **Nit (cosmetic):** the skip short-circuit nests three `if`/`if let`. A let-chain (`if !params.no_stamps && let Some(stamp) = params.prefetched_stamp.as_ref() && let Some(inputs) = compute_install_stamp_inputs(params, &packages)?`) or a small extracted helper would read a touch flatter. Purely stylistic — feel free to ignore.

src/commands/rootfs/install.rs
```diff
@@ -1040,10 +1154,18 @@ impl RootfsInstallCommand {
...
             tui_context: None,
         })
         .await;
 
+        // Persist the lockfile the install updated. `install_sysroot` no
+        // longer saves for itself — under `avocado sdk install` it runs on a
+        // clone that the caller merges and saves once.
+        if result.is_ok() {
+            lock_file.save(src_dir)?;
+        }
```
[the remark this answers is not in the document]
mobileoverlord: Traced this, and the new ordering is actually tighter than what it replaced — the gate is safe.

The load-bearing fact: the stamp write is the **last fallible step** in `install_sysroot` before `Ok(())`. Nothing between the stamp container run and the return can fail. So there are only three outcomes:

- fail *before* the stamp write → `Err` → lock not saved **and** no stamp written
- fail *during* the stamp write → `Err` → lock not saved **and** no stamp written
- reach `Ok(())` → lock saved **and** stamp written

The lock and the stamp are now persisted under the same condition, so they cannot disagree. The "sysroot volume has changed but `avocado.lock` is not updated" state you describe is real but benign: because no stamp was written either, the next run does **not** skip — it re-runs the install, re-resolves, and saves. The hazard you're pointing at (a later install trusting pins that don't match the sysroot) requires a stamp present *without* a matching lock, and this ordering makes that unreachable.

Worth contrasting with what it replaced. On `main`, `install_sysroot` saved the lock at `rootfs/install.rs:791` — *before* kernel-sysroot staging and *before* the stamp write. A failure in between left the lock persisted from a run that did not complete, which is the inconsistency you're describing, in the direction that actually bites.

The reason the inner save had to go is separate from the failure semantics: under `avocado sdk install`, `install_sysroot` now runs against a **clone** of the lockfile that the caller merges and saves once (`sdk/install.rs`). Saving from inside would clobber the parent lock with a partial, single-sysroot view.

src/commands/initramfs/install.rs
```diff
@@ -147,10 +167,17 @@ impl InitramfsInstallCommand {
...
+            prefetched_stamp,
             tui_context: None,
         })
         .await;
 
+        // Persist the lockfile the install updated — `install_sysroot` leaves
+        // saving to its caller.
+        if result.is_ok() {
+            lock_file.save(src_dir)?;
+        }
```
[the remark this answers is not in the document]
mobileoverlord: Traced this, and the new ordering is actually tighter than what it replaced — the gate is safe.

The load-bearing fact: the stamp write is the **last fallible step** in `install_sysroot` before `Ok(())`. Nothing between the stamp container run and the return can fail. So there are only three outcomes:

- fail *before* the stamp write → `Err` → lock not saved **and** no stamp written
- fail *during* the stamp write → `Err` → lock not saved **and** no stamp written
- reach `Ok(())` → lock saved **and** stamp written

The lock and the stamp are now persisted under the same condition, so they cannot disagree. The "sysroot volume has changed but `avocado.lock` is not updated" state you describe is real but benign: because no stamp was written either, the next run does **not** skip — it re-runs the install, re-resolves, and saves. The hazard you're pointing at (a later install trusting pins that don't match the sysroot) requires a stamp present *without* a matching lock, and this ordering makes that unreachable.

Worth contrasting with what it replaced. On `main`, `install_sysroot` saved the lock at `rootfs/install.rs:791` — *before* kernel-sysroot staging and *before* the stamp write. A failure in between left the lock persisted from a run that did not complete, which is the inconsistency you're describing, in the direction that actually bites.

The reason the inner save had to go is separate from the failure semantics: under `avocado sdk install`, `install_sysroot` now runs against a **clone** of the lockfile that the caller merges and saves once (`sdk/install.rs`). Saving from inside would clobber the parent lock with a partial, single-sysroot view.

src/commands/sdk/install.rs
mobileoverlord: Good catch — fixed. Dropped `verbose: self.verbose` from all three of the new stamp-read `RunConfig`s, so they now match the five pre-existing `read_stamps_batch` callers (`build.rs`, `ext/checkout.rs`, `ext/clean.rs`, `sdk/clean.rs`, `sdk/package.rs`), which all leave it at its `false` default:

- `sdk/install.rs` — the batched two-sysroot read here
- `rootfs/install.rs` — the standalone `read_sysroot_install_stamp` config
- `initramfs/install.rs` — same

You're right that it was harmless (output is captured and `parse_batch_stamps_output` drops non-`:::` lines), but there's no reason for an `echo … | tr` to run with `AVOCADO_VERBOSE=1`. `fmt`, `clippy -D warnings`, and the full suite are green after the change.

src/commands/rootfs/install.rs
mobileoverlord: Leaving this one as-is, because the let-chain form can't compile here — this crate is `edition = "2021"`, and let-chains are gated to edition 2024. Confirmed directly:

```
$ rustc --edition 2021 letchain.rs
error: let chains are only allowed in Rust 2024 or later
 --> letchain.rs:2:14
  |
2 |     if !a && let Some(x) = b && let Some(y) = c { return x == y; }
  |              ^^^^^^^^^^^^^^^

$ rustc --edition 2024 letchain.rs      # compiles
```

(rustc 1.94.0, so the feature is stable — it's the edition that gates it.)

That leaves the extracted helper as the only viable option, and at three levels guarding a single early return I'd rather keep the nesting than add an indirection whose body is one `if`. Happy to revisit as part of an edition-2024 bump, where the let-chain becomes a genuine one-line simplification and the whole file can move at once rather than this one spot.

mobileoverlord:
@jetm this is ready for review.

All four review threads are addressed and resolved:

- **Both Copilot lockfile threads** — replied with the trace rather than a change. The stamp write is the last fallible step before `Ok(())`, so the lock and the stamp are now persisted under the same condition and cannot disagree. The state Copilot describes is reachable but benign, because no stamp is written either, so the next run reinstalls instead of trusting stale pins. This is strictly tighter than `main`, which saved the lock at `rootfs/install.rs:791` — before kernel staging and before the stamp write — leaving a persisted lock from a run that never completed.
- **@nicksinas's verbose nit** — fixed in b9dc588. Dropped `verbose: self.verbose` from all three new stamp-read `RunConfig`s so they match the five pre-existing `read_stamps_batch` callers.
- **@nicksinas's let-chain nit** — declined with a reason: this crate is `edition = "2021"` and let-chains are gated to edition 2024, so the suggested form doesn't compile (verified against rustc 1.94). Worth revisiting on an edition bump.

Verified locally on the pushed tip: `cargo fmt --all -- --check` clean, `cargo clippy --all-targets --all-features -- -D warnings` clean, and `cargo test --all` green at 1124 + 1132 + 126 + 20 + 5 + 49 + 6 + 7 + 6, zero failures.

One thing worth a second opinion, since it's the actual behavioral heart of the PR: the skip short-circuit means an unchanged project does **no** container work for rootfs/initramfs. The escape hatches are `--no-stamps` and `avocado {rootfs,initramfs} clean`. If you can think of an input that should invalidate the stamp but isn't in `compute_install_stamp_inputs`, that's the bug worth finding here.

jetm (commented):
Reviewed this at b9dc588 and built it in a scratch worktree - `cargo test --no-run` exits 0. Twelve findings inline, most severe first. The findings themselves are code-traced rather than build-derived, except where a comment says otherwise.

The one that matters most is the first: the `sdk install` lock merge copies `entry.rootfs` but not `cleared_sections`, and because this PR also removed the `save()` that used to run on the clone (base `install.rs:791`), a kernel-pin-change clean is no longer authoritative on disk. The result is that `avocado.lock` permanently regains the old kver's keys, so every later run reads stale and wipes both sysroots - which is the exact loop this PR exists to remove, on the path it is named for.

A pattern worth noting across findings 3, 4, 7 and 9: moving the lock save after the stamp block introduced several windows where an install fully succeeds on disk but its pins or its teardown are skipped. Individually each is a narrow failure; together they mean a partial failure now tends to leave the stamp claiming success while the state behind it is wrong, and the recovery the error text names ("run rootfs install") is the command that then answers "up to date" and exits 0.

Two of the twelve I marked lower-confidence rather than dropping: the latched failed wipe (I could not stage a forced non-zero exit from the clean invocation) and the `sdk.image` omission from the sysroot stamp (I could not name two images that yield different sysroot contents for identical feed plus config, so treat it as a question about which inputs belong in that hash).

Things I checked and decided are NOT problems, so nobody re-derives them: `avocado build --runs-on` reading stamps from the local volume is byte-identical to pre-PR behaviour and purely pre-existing; the `rm -rf "$AVOCADO_PREFIX/..."` shell is not an injection risk because `sysroot_dir` is a compile-time literal and the prefix is exported unconditionally by the entrypoint; and `STAMP_VERSION` not being bumped is safe in this direction, since an old stamp cannot match the wider hash and so reads stale.

src/commands/sdk/install.rs
```diff
@@ -502,6 +537,7 @@ $DNF_SDK_HOST $DNF_NO_SCRIPTS $DNF_SDK_TARGET_REPO_CONF \
             sdk_arch: self.sdk_arch.as_ref(),
             no_stamps: self.no_stamps,
             parsed: Some(&composed.merged_value),
+            prefetched_stamp: initramfs_stamp,
             tui_context: initramfs_tui,
         };
```
jetm, line 543: **The lock merge drops `cleared_sections`, so a kernel repin never sticks**

This defeats the PR's stated purpose on exactly the path it is named for, so it is the one I would fix first. (Anchored on the last changed line of this file - the merge itself is at `sdk/install.rs:651`, outside the diff, which is part of why the interaction is easy to miss.)

The merge does `entry.rootfs = target_locks.rootfs.clone()` but never copies the clone's `cleared_sections`, which is `#[serde(skip)]` and therefore in-memory only. The PR also deleted `params.lock_file.save(params.src_dir)?` from inside `install_sysroot` (base `install.rs:791`). That save ran on the clone, which *did* carry `clear_rootfs` set, so a kernel-pin-change clean was authoritative on disk. Now the only save is `final_lock.save(src_dir)`, and `final_lock` has an empty flag set - so `merge_with`'s `entry(k).or_insert(v)` (`lockfile.rs:934-943`) resurrects the old kver's keys.

After a kernel repin under `avocado sdk install`, `avocado.lock` permanently regains `kernel-image-<oldkver>` and `packagegroup-avocado-rootfs-modules-<oldkver>`. Every later run then sees the disk `package_list_hash` differ from the stamp written off the clone, reads stale, and `detect_sysroot_package_removals` reports those two - so both sysroots get wiped and reinstalled on every invocation, forever.

Copying `cleared_sections` across in the merge is the narrow fix.

src/commands/sdk/clean.rs
```diff
@@ -111,28 +112,23 @@ impl SdkCleanCommand {
...
+                    sdk_arch: self.sdk_arch.clone(),
+                    ..Default::default()
+                },
+                None,
+            )
+            .await?
+            .validate(&requirements, &[]);
 
             if !validation.is_satisfied() {
                 validation
```
jetm, line 134: **`sdk clean` leaves the install stamps behind**

(Anchored on the end of this hunk; the cleanup itself is at `clean.rs:159`, just outside the diff.)

The cleanup removes the SDK prefix and the rootfs/initramfs sysroots, but never touches `$AVOCADO_PREFIX/.stamps/{rootfs,initramfs}`, and `avocado.lock` is host-side and untouched.

What makes this worth raising now is that the PR fixed precisely this hazard for the siblings - `rootfs/clean.rs:21` now removes `.stamps/{sysroot_dir}`, with a doc comment naming the failure - and this third site was missed.

So `avocado sdk clean` followed by `avocado sdk install` on an unchanged project: both stamps survive, both hashes still match, both installs print "rootfs/initramfs sysroot is up to date." and do nothing. Exit 0 with two empty sysroots, and the following `build` or `rootfs image` consumes an empty tree. Recovery needs `avocado clean --stamps` or `--no-stamps`, and no message names either.

src/commands/rootfs/install.rs
```diff
@@ -839,63 +942,56 @@ $DNF_SDK_HOST $DNF_SDK_TARGET_REPO_CONF \
...
+        // Write install stamp (unless --no-stamps or no parsed config available).
+        //
+        // Computed fresh rather than reusing the value the skip check above
+        // derived: the install just re-pinned this sysroot's packages, and the
+        // stamp has to record the lock state the *next* run will compare
+        // against.
         if !params.no_stamps {
-            if let Some(parsed) = params.parsed {
-                let stamp_result = match params.sysroot_type {
+            if let Some(inputs) = compute_install_stamp_inputs(params, &packages)? {
```
jetm, line 952: **A stamp-write failure now discards the install's lock pins**

`update_sysroot_versions` records the NVRs, and then four `?` sites follow before any save: `compute_install_stamp_inputs(...)?` here at 952, `generate_write_stamp_script(&stamp)?` at 966, and the two stamp-write `.await?` runs at 982 and 987. Callers save only `if result.is_ok()`. Pre-PR the save sat at old line 791, *before* the stamp block, so this window did not exist.

On `avocado rootfs install --runs-on builder`: dnf succeeds, the pins and the kernel re-pin are recorded in memory, then the SSH channel for the stamp-write run drops. `install_sysroot` returns `Err`, the caller skips the save, and `avocado.lock` keeps its pre-install state while the remote sysroot is fully populated. The next run re-resolves the kernel against feed head with no `prev_pinned_kver` to compare against, so it installs additively on top and skips the pin-change clean.

src/commands/rootfs/install.rs
```diff
@@ -839,63 +942,56 @@ $DNF_SDK_HOST $DNF_SDK_TARGET_REPO_CONF \
...
         }
 
-        // Write install stamp (unless --no-stamps or no parsed config available)
+        // Write install stamp (unless --no-stamps or no parsed config available).
+        //
+        // Computed fresh rather than reusing the value the skip check above
+        // derived: the install just re-pinned this sysroot's packages, and the
+        // stamp has to record the lock state the *next* run will compare
+        // against.
         if !params.no_stamps {
```
jetm, line 951: **The stamp is written even when kernel staging silently failed**

The lock update is gated on `!installed_versions.is_empty()` at 880, but the stamp write is not - and `stage_kernel_sysroot_from_rootfs`'s error is only `print_error`ed at 934 with no `return`. `query_installed_packages` also returns `Ok(HashMap::new())` on failure (`container.rs:1721`), so an empty result is indistinguishable from a real one.

One failed kernel-staging container run therefore yields: install reports success, a current stamp is written, and `$AVOCADO_PREFIX/kernel/<kver>/` is never created. `avocado kernel image` then hard-fails with `ERROR: no staged kernel image at $AVOCADO_PREFIX/kernel/*/Image - run 'avocado rootfs install' first` (`kernel/image.rs:172`) - and that named recovery answers "rootfs sysroot is up to date." and exits 0. The user is wedged until they guess `--no-stamps` or `rootfs clean`.

Gating the stamp write on the same condition as the lock update would keep the two consistent.

src/commands/rootfs/install.rs
```diff
@@ -593,6 +616,97 @@ pub async fn install_sysroot(params: &mut SysrootInstallParams<'_>) -> Result<()
         _ => None,
     };
 
+    // The set of names a completed install of this sysroot is expected to
+    // have recorded in the lockfile: config-declared packages plus whatever
+    // was auto-appended above. Used as the reference set for removal
+    // detection and, after the install, as the lock query list.
+    let mut effective_names: Vec<String> = if packages.is_empty() {
```
jetm, line 623: **A transient repoquery failure wipes the sysroot and drops the modules packagegroup**

`package_exists_in_target_repo` ends with a `|| true` at 423 and maps empty stdout to `Ok(false)` at 453, so a feed or metadata blip is indistinguishable from "not in feed" - and it prints the misleading "feed predates per-kernel module packagegroups" while it does so. `auto_module_pkg` becomes `None` and drops out of `effective_name_set`, so the locked packagegroup reads as a removal.

One `sdk install` that hits transient feed trouble during the probe: removal detection fires, `clean_sysroot` wipes the sysroot, and the reinstall proceeds *without* the module packagegroup. Because `packages_for_hash` hashes only config packages and auto-appends are not in it, the next healthy run's skip check at 482 fires before the repoquery runs. The image boots a kernel with no modules and never self-heals.

Distinguishing "probe failed" from "absent from feed" is what stops a blip becoming a permanent state.

src/commands/rootfs/install.rs
```diff
@@ -593,6 +616,97 @@ pub async fn install_sysroot(params: &mut SysrootInstallParams<'_>) -> Result<()
...
+        print_info(
+            &format!(
+                "Packages removed from {label}: {}. Cleaning sysroot for fresh install.",
+                removed_packages.join(", ")
+            ),
+            OutputLevel::Normal,
+        );
+        // Drop only the stale entries, preserving version pins for the
+        // packages that remain.
+        params.lock_file.remove_packages_from_sysroot(
```
jetm, line 699: **A removed pin is resurrected, so `package_list_hash` never converges**

`remove_packages_from_sysroot` (`lockfile.rs:1432`) never inserts into `cleared_sections`, unlike `clear_rootfs` at 1358. `save()` merges disk back in, so a deliberately dropped pin returns on the very next save. Note this one bites the standalone path too, not only the `sdk install` merge site.

Drop `vim` from `rootfs.packages`: run 1 detects the removal, wipes, reinstalls, and writes a stamp from the pruned in-memory lock - then `save()` re-adds `vim` from disk. Run 2 loads the resurrected lock, the hash mismatches, it reads stale, the removal is detected again, and it wipes. Full reinstall on every invocation forever, with `avocado.lock` permanently listing a package that is not installed.

src/commands/rootfs/install.rs
```diff
@@ -1021,6 +1117,23 @@ impl RootfsInstallCommand {
...
+                target: target.to_string(),
+                repo_url: repo_url.clone(),
+                repo_release: repo_release.clone(),
+                container_args: merged_container_args.clone(),
+                sdk_arch: self.sdk_arch.clone(),
+                ..Default::default()
+            },
+            runs_on_context.as_ref(),
+        )
+        .await?;
```
jetm, line 1135: **Two new `?` sites skip the "always teardown" `runs_on` block**

Pre-PR this body was `let result = install_sysroot(...).await;` then `// Always teardown runs_on context` then `result` - the capture was deliberate, and there was no `?` above the teardown. This PR inserts two: `read_sysroot_install_stamp(...).await?` here at 1135 and `lock_file.save(src_dir)?` at 1165. `src/commands/initramfs/install.rs:133` has the same shape.

On `avocado rootfs install --runs-on user@host`, if the remote SDK container invocation fails (image pull, remote podman down) or the project dir is read-only, the `?` returns early, `context.teardown()` never runs, and the local NFS server plus the remote mount stay up. The next `--runs-on` run picks a different `nfs_port` and stacks another one on top.

Capturing these into `result` the way the original did preserves the invariant the comment still claims.

src/commands/rootfs/install.rs
jetm, line 479: **`--dnf-args` is silently ignored on an up-to-date sysroot**

`dnf_args` is interpolated straight into the transaction at 826 but is absent from `SysrootStampInputs`, so this short-circuit fires before the flag is ever used.

After a successful install, `avocado rootfs install --dnf-args '--enablerepo=extra'` returns at line 487 having done nothing, prints "rootfs sysroot is up to date.", and exits 0 with no hint the flag was discarded.

For contrast, `--force` is *not* affected - it only selects `-y` and interactivity, not content - so folding `dnf_args` into the stamp inputs is the difference between a flag that changes the transaction and one that does not.

src/commands/rootfs/install.rs
```diff
@@ -593,6 +616,97 @@ pub async fn install_sysroot(params: &mut SysrootInstallParams<'_>) -> Result<()
...
+        // packages that remain.
+        params.lock_file.remove_packages_from_sysroot(
+            params.target,
+            &params.sysroot_type,
+            &removed_packages,
+        );
+    }
+
+    if needs_clean_reinstall {
+        clean_sysroot(params, sysroot_dir).await;
```
jetm, line 707: **A failed mid-install wipe is latched by the stamp write** (lower confidence)

`clean_sysroot` discards both the `Result` and the returned success bool - `.await.ok()` at 349 and 355 - which is preserved from the pre-PR inline copies, so that part is not new. What is new is that the run now continues to the stamp write regardless, so a failed wipe is latched instead of re-attempted on the next run.

Kernel pin changes K to K' and the cleanup container run fails (podman hiccup, or a `--runs-on` blip). dnf then installs K' alongside K, leaving `/lib/modules/K/` and K's kernel image behind. A current stamp is written, so every later run reports "up to date" and the two-kernel sysroot is never repaired.

Marked lower confidence because confirming it needs a forced non-zero exit from the clean invocation, which I did not stage.

src/utils/stamps.rs
```diff
@@ -1282,61 +1345,66 @@ pub fn compute_rootfs_input_hash(
         );
     }
 
+    // Feed identity and resolver flags. A snapshot bump, a feed switch, or a
```
jetm, line 1348: **The sysroot stamp omits `sdk.image`, unlike its sibling hash** (lower confidence)

`compute_sdk_input_hash` folds `sdk.image` (`stamps.rs:979`), but the new `compute_sysroot_install_input_hash` folds only `repo_url`, `repo_release` and `disable_weak_dependencies`. The SDK image is what actually runs the install - its dnf/rpm config and scriptlet machinery - as are `sdk.container_args`.

A project on `sdk.image: .../sdk:dev` that pins to a fixed tag with different dnf/rpm config: feed and all config sections are byte-identical, so both stamps stay current and the sysroots keep whatever the old image produced.

Uncertain because package NVRs come from the feed and the repo URL is always passed explicitly, so I could not point at a concrete pair of images that yield different sysroot contents for identical feed plus config. Raising it as a question about which inputs belong in that hash rather than as a confirmed bug.

src/commands/rootfs/install.rs
```diff
@@ -1060,7 +1181,139 @@ impl RootfsInstallCommand {
...
+            detect_sysroot_package_removals(&effective, &SysrootType::Rootfs, TARGET, &lock),
+            vec![
+                format!("kernel-image-{KVER}"),
+                format!("packagegroup-avocado-rootfs-modules-{KVER}"),
+            ],
+        );
+    }
+
+    #[test]
+    fn removal_detection_is_noop_on_first_install() {
```
jetm, line 1306: **`removal_detection_is_noop_on_first_install` cannot fail**

`LockFile::new()` means `get_locked_package_names` is empty, so the function early-returns `Vec::new()` at 311-313 before any comparison logic runs. I inverted the `difference` operands and the test still passes.

That runs into this repo's own `CLAUDE.md` test-integrity rule - "Every test must fail when the implementation is wrong (flip the logic and check)" - and the cost is that the test constrains nothing, so a future regression in the comparison direction ships undetected.

Seeding the lock with at least one package name is what would make it bite.

(`removal_detection_flags_stale_kernel_auto_appends` at 1277 is fine by comparison - I initially read it as vacuous, but the state it tests *is* reachable when `prev_pinned_kver` is `None`; only its comment is wrong.)

src/utils/prerequisites.rs
```diff
@@ -9,10 +9,85 @@
...
+    }
+
+    /// Validate `requirements` against freshly computed inputs. See
+    /// [`validate_stamps_batch`] for how requirements are matched to inputs.
+    pub fn validate(
+        &self,
+        requirements: &[StampRequirement],
+        current_inputs: &[CurrentInput<'_>],
+    ) -> StampValidationResult {
+        validate_stamps_batch(requirements, &self.raw, current_inputs)
```
jetm, line 46: **`StampBatch` holds both raw and parsed, and re-parses on validate**

`read_stamps_batch` parses into `self.stamps`, then keeps `raw` alive solely so `validate()` can call `validate_stamps_batch(requirements, &self.raw, ...)` - whose first line is `let stamp_data = parse_batch_stamps_output(batch_output);` (`stamps.rs:1659`), producing the same map it already holds.

So there are two representations of one value, one redundant parse per validate, and a field comment that documents the redundancy rather than removing it. `check_stamp_requirement` takes `Option<&str>`, which the stored map supplies directly, and all 6 new call sites go through `StampBatch` - so a parsed-map variant is mechanical and behavior-preserving.

src/commands/sdk/install.rs
mobileoverlord: Fixed in e052539. `cleared_sections` is now copied across in both the rootfs and initramfs merge arms.

You were right that this was the one to fix first, and right about why it was easy to miss — the flag is `#[serde(skip)]`, so it is invisible in every lockfile you could inspect, and the merge is outside the diff. The chain you traced holds exactly: the flag lived only on the clone, the only surviving save was `final_lock.save`, and `or_insert` then resurrected the old kver's keys on the very next write.

Worth noting the second half of it, which your other comment on `remove_packages_from_sysroot` turned out to be: even with the flag propagated, the *standalone* path had the same non-convergence for a different reason. Both are fixed, and they are the same bug wearing two hats — a section that is authoritative in memory losing that fact before the disk merge.

src/commands/sdk/clean.rs
mobileoverlord: Fixed in e052539 — `sdk clean` now removes `.stamps/{rootfs,initramfs}` along with the sysroots.

You put your finger on why this one was worse than a plain omission: the PR taught `rootfs clean` and `initramfs clean` to do exactly this, and documented the hazard in `clean_sysroot_command`'s doc comment, then left the third site on the old behavior. So the codebase asserted the invariant in prose while violating it one file over.

`exit 0` with two empty sysroots and an "up to date" message is about the worst shape a failure can take, and the recovery (`--no-stamps`, `avocado clean --stamps`) is unguessable from the output.

src/commands/rootfs/install.rs
mobileoverlord: Fixed in e052539, though not by moving the save back.

The stamp write is now a helper (`write_install_stamp`) whose failure is warned about rather than propagated, so none of those four `?` sites can return `Err` after `update_sysroot_versions` has recorded the pins. Nothing between the pin recording and the `Ok` can fail now, which is the invariant the caller-saves design needs.

Restoring the old save position would have reintroduced what it was removed for — under `sdk install`, `install_sysroot` runs on a clone, and saving from inside clobbers the parent lock with a partial single-sysroot view. Making the tail infallible gets the same guarantee without that.

The resulting semantics line up with the stamp gating in your other comments: a stamp that could not be written is benign, because its absence makes the next run reinstall. A lock that was not saved is not benign, because the sysroot moved without it. So the failure modes now fall on the harmless side.

src/commands/rootfs/install.rs
mobileoverlord: Fixed in e052539, using the gating you suggested plus one more condition.

The stamp write is now gated on `versions_recorded && kernel_staging_ok && install_is_clean`, and when it is skipped it says so and names which condition failed, so the next run reinstalls instead of reporting "up to date".

Your point about `query_installed_packages` returning `Ok(HashMap::new())` on failure is what makes the `!installed_versions.is_empty()` gate load-bearing rather than cosmetic — an empty result really is indistinguishable from a real one at that call site, so the only safe reading is "do not claim this install is recorded".

The wedge you describe was the sharpest part of the report: `avocado kernel image` telling the user to run `rootfs install`, and `rootfs install` answering "up to date" and exiting 0. A recovery instruction that points at a command which refuses to act is worse than no instruction.

src/commands/rootfs/install.rs
mobileoverlord: Fixed in e052539. `package_exists_in_target_repo` now returns `Result<Option<bool>>` — `Some` when the probe answered, `None` when it failed.

The `|| true` is replaced by a sentinel: the command prints `AVOCADO_PROBE_OK` before repoquery's output on the success branch and `AVOCADO_PROBE_FAILED` otherwise, so an exit code that the shell pipeline used to swallow now survives to the caller. A failed probe keeps the packagegroup in the install set rather than assuming it absent — if it genuinely is missing, dnf fails loudly instead of quietly shipping a module-less kernel.

The detail that made this worth fixing rather than tolerating is the one you added at the end: `packages_for_hash` covers only config packages, so the next healthy run's skip check fires before the probe ever runs. That is what turns a transient blip into a permanent state — there is no self-heal path. Parsing is now a pure `parse_probe_output` with a test covering the failed, empty, populated, and banner-prefixed cases.

src/commands/rootfs/install.rs
mobileoverlord: Fixed in e052539. Both new `?` sites are captured into `result`, in `rootfs/install.rs` and `initramfs/install.rs`.

The stamp read now lands in a `match` that carries an `Err` through to the teardown, and the lockfile save is folded in as `Ok(()) => lock_file.save(src_dir)` so it still only runs on success but can no longer return early. The teardown block's comment claims an invariant, and it is true again.

The stacking detail is what makes this more than a leaked process: because each run picks a different `nfs_port`, repeated failures pile up rather than colliding and being noticed.

src/commands/rootfs/install.rs
mobileoverlord: Fixed in e052539 by marking the section cleared inside `remove_packages_from_sysroot`, the same way `clear_rootfs` does.

Your observation that this bites the standalone path too is what made it a separate bug from the merge one rather than the same bug seen twice — propagating `cleared_sections` across the `sdk install` merge does nothing for `avocado rootfs install`, which never goes through that merge.

On the flag's meaning: "cleared" is still accurate here even though only some keys are dropped, because the caller goes on to reinstall and record the full installed set, so by save time the in-memory section is the complete intended state. I noted that in the code so the next reader does not have to re-derive it. Only set when a key was actually removed, so a no-op removal list does not suppress the merge.

src/commands/rootfs/install.rs
mobileoverlord: Fixed in e052539 — `dnf_args` is now part of `SysrootStampInputs` and folded into the hash.

Your contrast with `--force` is what made the boundary obvious: `--force` selects `-y` and interactivity, `--dnf-args` changes what the transaction resolves. One belongs in the hash and the other does not, and the test asserts both directions (a non-empty list invalidates, an empty one does not, so `--dnf-args ""` will not force pointless rebuilds).

Hashing the sequence as given rather than a normalized set, so a reorder invalidates. Wrong-but-safe beats missing a change.

src/commands/rootfs/install.rs
mobileoverlord: Confirmed and fixed in e052539 — you were right to raise it even at lower confidence.

`clean_sysroot` now returns whether the wipe succeeded instead of discarding it through `.ok()`, and that feeds the same stamp gate as the other two integrity conditions. Your read of what is new versus preserved was exactly right: discarding the result came from the pre-PR inline copies, and what this PR added was the stamp write that latches the consequence.

I did not stage a forced non-zero exit either, so this is fixed on the reasoning rather than on a reproduction — but the reasoning does not have much room in it. dnf install is additive, a failed wipe means the old kernel's modules and image survive, and a current stamp over that is never re-attempted.

src/utils/stamps.rs
mobileoverlord: Fixed in e052539, and I think your uncertainty was about the wrong thing.

`sdk.image` is now folded into `compute_sysroot_install_input_hash`, matching `compute_sdk_input_hash`. You could not point at a concrete pair of images yielding different sysroots for identical feed plus config, and I cannot either — but the defect does not need one. The two hashes disagreeing about what counts as an input is the bug: the image is what runs dnf, so either both stamps should track it or neither should, and picking per-hash means the answer depends on which stamp you are looking at.

Folded the image only, not `sdk.container_args`. Those are mounts, devices and env for the run rather than the transaction, and I would rather not invalidate both sysroots on a `-v` change without a concrete reason. Happy to add them if you have one.

src/commands/rootfs/install.rs
mobileoverlord: Confirmed and fixed in e052539 — and I verified your claim the same way you did rather than taking it on faith. Inverting the `difference` operands leaves `removal_detection_is_noop_on_first_install` green, because the empty lock early-returns at 311-313 before any comparison.

Added `removal_detection_compares_locked_against_effective_not_the_reverse`, which seeds the lock so the comparison actually runs and pins the direction from both sides: a dropped package reports as removed, and a newly added one must not. Under the inverted implementation the new test fails on the second arm while the old one still passes — which is the check the `CLAUDE.md` rule is asking for.

Kept the original test. It is vacuous as a logic check but does pin the early return, which is a real behavior. Thanks also for re-checking `removal_detection_flags_stale_kernel_auto_appends` and correcting your own initial read — I fixed its comment rather than the test.

src/utils/prerequisites.rs
mobileoverlord: Fixed in e052539. `StampBatch` holds only the parsed map now.

`validate_stamps_batch` keeps its `&str` signature — it has around twenty callers passing raw container output — and delegates to a new `validate_stamps_parsed` that takes the map. `StampBatch::validate` calls that directly, so the `raw` field and the per-validate re-parse are both gone. Behavior-preserving, as you predicted.

The field comment documenting the redundancy instead of removing it was the part worth calling out. A comment explaining why a structure is awkward is usually a sign the structure should change.

mobileoverlord:
@jetm all twelve are addressed in e052539 (three commits), with per-thread replies inline. Short version:

**Fixed as reported (9):** the `cleared_sections` merge drop, `sdk clean` leaving the install stamps, the resurrected pin in `remove_packages_from_sysroot`, the two `?` sites skipping teardown, the stamp surviving a failed kernel staging, the stamp-write window discarding lock pins, the transient repoquery wipe, `--dnf-args` missing from the stamp inputs, and the vacuous first-install test.

**Both lower-confidence ones were real (2).** The latched failed wipe I fixed on reasoning rather than a reproduction, same as you — `clean_sysroot` now reports its outcome instead of `.ok()`-ing it. On `sdk.image`: I think your uncertainty was aimed at the wrong thing. You couldn't name a pair of images producing different sysroots for identical feed plus config, and neither can I, but the defect doesn't need one — the two hashes *disagreeing* about what counts as an input is the bug. Folded it in to match `compute_sdk_input_hash`. I left `sdk.container_args` out; those are mounts and env for the run rather than the transaction, and I'd want a concrete reason before invalidating both sysroots on a `-v` change.

**Cleanup (1):** `StampBatch` holds only the parsed map; `validate_stamps_batch` keeps its `&str` signature for its ~20 callers and delegates to a new `validate_stamps_parsed`.

Two things worth pulling out of the threads:

**The two `cleared_sections` findings were one bug wearing two hats** — a section that is authoritative in memory losing that fact before the disk merge. Propagating across the `sdk install` merge does nothing for `avocado rootfs install`, which never goes through it, so both sites genuinely needed fixing. Neither is visible in a lockfile, since the flag is `#[serde(skip)]`.

**The failure modes now fall on the harmless side, consistently.** A stamp that couldn't be written is benign — its absence makes the next run reinstall. A lock that wasn't saved is not benign — the sysroot moved without it. So the stamp write is now a helper whose failure warns rather than propagating, and the stamp is gated on all three integrity conditions (`versions_recorded && kernel_staging_ok && install_is_clean`), naming which one failed when it skips. I didn't move the lock save back to its pre-PR position, since that's what the clone design removed it for; making the tail infallible gets the same guarantee without clobbering the parent lock.

**One thing found while fixing yours:** `install_sysroot`'s doc comment was sitting above `package_exists_in_target_repo`, so `install_sysroot` shipped undocumented and the probe carried prose describing a different function. Pre-existing on this branch, not from your review, but the same shape as the misattachment you caught elsewhere. Both reattached.

**New tests:** removal direction (verified it fails under inverted operands, per the `CLAUDE.md` rule), probe-output parsing across failed/empty/populated/banner-prefixed, `--dnf-args` invalidating while an empty list doesn't, and `sdk.image` invalidating.

**On CI:** nothing was actually wrong with it. The two red `Analyze` rows were CodeQL jobs with conclusion `cancelled`, not `failure` — killed at 15m02s each in the Aug 6 Actions outage, which `gh pr checks` renders as `fail`. The `Extension Test` workflow had also been stuck `queued` for 101 hours from the same outage, which is why the board build and version check were missing from the list entirely. Cancelled it and pushed; the full set is running now. Locally `fmt --check`, `clippy --all-targets --all-features -D warnings` and the suite are clean. I also merged `main` in.

jetm (approved):

PR #186, opened 2026-07-31
Report the build commit in avocado --version

jetm:
## Problem

`avocado --version` reports only `avocado 1.0.0-rc.1`, which does not identify which build a user is running. Release candidates move faster than the version number, so one string covers many different binaries and a bug report cannot be tied to a commit.

`build.rs` already computed a short SHA into `AVOCADO_CLI_VERSION`, but nothing consumed it: the CLI declared a bare `#[command(version)]`, which clap fills from `CARGO_PKG_VERSION`.

## Solution

Report the commit in rustc's shape:

```
avocado 1.0.0-rc.1 (e566baa 2026-07-31)
```

Keeping the version in its own field leaves it a bare semver, so version comparisons need no special handling for an appended commit. The date is the **commit** date, not the build date, so rebuilding the same commit yields the same string.

This does move the version out of the last field, which two consumers were reading:

- `utils::remote::check_cli_version` (this repo) — fixed here.
- avocado-desktop's `parse_cli_version` — fixed in avocado-linux/avocado-desktop#73. That one accepts both the old and new shapes, so it is safe to merge first and is correct against today's released CLI too.

## Key changes

- `main.rs`: `#[command(version = env!("AVOCADO_CLI_VERSION"))]`, so the value `build.rs` computes actually reaches `--version`.
- `build.rs` emits `<version> (<short-sha> <commit-date>)`, and is hardened: it unwrapped the git invocation, so building outside a checkout (source tarball, vendored crate) aborted the build instead of falling back to the bare version; it never trimmed the trailing newline off `rev-parse`, which would have put a line break inside the version string; and it declared no `rerun-if-changed`, so cargo cached the script output and kept reporting a stale commit. The rerun hint follows the **branch ref**, not `HEAD` alone, because `HEAD` is a symref whose contents do not change on commit.
- `check_cli_version` reads the version field via a `parse_reported_version` helper instead of `split_whitespace().last()`, which now yields the commit date. This matters beyond cosmetics: an unparseable version makes that check **fail open** and silently accept a remote running an older CLI. The helper still tolerates a bare version from an older remote.
- `tests/version.rs`: the version field is a bare semver equal to the crate version, and the line carries the current commit and date (skips outside a git checkout).
- Unit tests for `parse_reported_version` across the new shape, the old bare shape, and a version with no prefix.

## Reviewer notes

- Verified end to end: prints `avocado 1.0.0-rc.1 (e566baa 2026-07-31)`; committing without touching `build.rs` rebuilds and reports the new commit, which is the check that the stale-commit fix actually works.
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`, and the full `cargo test` suite are clean.
- No CI change needed: `actions/checkout@v5` leaves `.git` present, so release and PR builds resolve a commit; `cross` musl builds and the Windows `cargo check` fall back to the bare version if git is unavailable rather than failing.
- `env!("TARGET")` from the same build script is unchanged (still used by `avocado upgrade`).
- A `.dirty` marker for uncommitted builds was deliberately left out to keep this focused; easy follow-up if wanted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Build commit in `avocado --version`.** The version line now reports the
+  commit the binary was built from and its date, following rustc's shape:
+  `avocado 1.0.0-rc.1 (abc1234 2026-03-05)`. A bug report now identifies the
+  exact build. Builds made outside a git checkout - or inside an unrelated
+  one, such as a vendored copy in another repository - keep reporting the bare
+  version rather than embedding a commit that is not this binary's. Anything
+  parsing this line must read the **second** whitespace-separated field, since
+  the last field is now a date.
+
+  **Rollout note.** That instruction only reaches parsers written from here
+  on. Every already-released `avocado` reads the remote's version with
+  `.last()` over the whole `--version` output, so during the window where an
+  older local CLI talks to a newer remote it takes `2026-03-05)` as the remote
+  version, fails to parse it, and - on those releases - falls open rather than
+  refusing: the minimum-remote-version check is skipped and a date is printed
+  as the version. Nothing can be shipped that fixes those binaries
+  retroactively, so upgrade the local side first. Moving the build detail to a
+  second output line would not have avoided this: the shipped parser splits the
+  entire stdout blob, not its first line, so `.last()` still reaches it.
+  `avocado-desktop` is not affected - its parser already reads the first line's
+  second field.
+
 ### Fixed
 - **`--connect-sign` guidance.** The deploy help text and the Level 2 setup
   messages now reference `avocado connect trust promote-root --key <KEY>` with
--- a/build.rs
+++ b/build.rs
@@ -1,17 +1,117 @@
 use std::env;
+use std::path::Path;
 use std::process::Command;
 
+/// Emit the version string reported by `avocado --version`, plus the target
+/// triple `avocado upgrade` uses to pick its release asset.
+///
+/// The build detail follows rustc's shape, `1.2.3 (abc1234 2026-03-05)`, so the
+/// version stays a bare semver in its own field and the commit and its date sit
+/// in a trailing parenthetical. Anything parsing this line has to read the
+/// SECOND whitespace-separated field: the last one is now a date.
+/// `utils::remote::check_cli_version` does, and avocado-desktop's
+/// `parse_cli_version` has to match.
+///
+/// The date is the commit date, not the build date, so the string stays
+/// reproducible across rebuilds of the same commit.
 fn main() {
-    // This runs only during build
-    let output = Command::new("git")
-        .args(["rev-parse", "--short", "HEAD"])
-        .output()
-        .unwrap();
-    let git_hash = String::from_utf8(output.stdout).unwrap();
-    println!(
-        "cargo:rustc-env=AVOCADO_CLI_VERSION={} {}",
-        env!("CARGO_PKG_VERSION"),
-        git_hash
-    );
+    println!("cargo:rerun-if-changed=build.rs");
+    emit_git_rerun_hints();
+
+    let version = match (
+        git(&["rev-parse", "--short", "HEAD"]),
+        git(&["show", "-s", "--format=%cs", "HEAD"]),
+    ) {
+        (Some(sha), Some(date)) if in_own_repository() => {
+            format!("{} ({sha} {date})", env!("CARGO_PKG_VERSION"))
+        }
+        // Building outside a git checkout (source tarball, vendored crate) is
+        // legitimate, so report the bare crate version rather than failing the
+        // build the way an unwrap here would. `in_own_repository` routes the
+        // more dangerous case here too: git succeeded, but against somebody
+        // else's repository.
+        _ => env!("CARGO_PKG_VERSION").to_string(),
+    };
+    println!("cargo:rustc-env=AVOCADO_CLI_VERSION={version}");
+
     println!("cargo:rustc-env=TARGET={}", env::var("TARGET").unwrap());
+}
+
+/// Whether the repository git just answered from is this crate's own.
+///
+/// `git` runs with the build script's cwd and will happily answer from whatever
+/// repository encloses it. A crate unpacked inside an unrelated one - a vendored
+/// copy in a monorepo, a source tree under a Yocto `WORKDIR` that is itself
+/// version-controlled, a docker build context inside a repo - would otherwise
+/// embed that repository's sha, and the string looks authoritative enough that a
+/// bug report citing an unrelated commit goes unquestioned. Reporting the bare
+/// version, as if git were absent, is the honest answer there.
+///
+/// This crate lives at its repository root, so an exact match is the check;
+/// anything else means git found a different tree.
+fn in_own_repository() -> bool {
+    let (Some(toplevel), Ok(manifest_dir)) = (
+        git(&["rev-parse", "--show-toplevel"]),
+        env::var("CARGO_MANIFEST_DIR"),
+    ) else {
+        return false;
+    };
+    // Compare canonicalized: either side can arrive with a symlinked path.
+    match (
+        std::fs::canonicalize(&toplevel),
+        std::fs::canonicalize(&manifest_dir),
+    ) {
+        (Ok(a), Ok(b)) => a == b,
+        _ => toplevel == manifest_dir,
+    }
+}
+
+/// Rebuild when HEAD moves. Without these, cargo caches the build-script
+/// output and the reported commit goes stale after the next commit or checkout.
+fn emit_git_rerun_hints() {
+    let mut watched = Vec::new();
+    if let Some(head) = git(&["rev-parse", "--git-path", "HEAD"]) {
+        watched.push(head);
+    }
+    // The reflog is the only one of these that changes on EVERY commit and is
+    // never packed away. `.git/HEAD` is a symref whose contents do not change
+    // when you commit, and the loose branch ref that does is deleted by
+    // `git gc` / `pack-refs` / `maintenance` - after which nothing watched here
+    // would move again and the embedded sha would stay stale indefinitely,
+    // without self-healing. `packed-refs` is no substitute: its mtime does not
+    // change on commit either.
+    if let Some(reflog) = git(&["rev-parse", "--git-path", "logs/HEAD"]) {
+        watched.push(reflog);
+    }
+    // Follow the branch ref too, for the case the reflog is disabled
+    // (`core.logAllRefUpdates=false`, the default in a bare repository).
+    // `--symbolic-full-name` prints `HEAD` on a detached checkout, which
+    // resolves right back to `.git/HEAD`; a detached checkout needs no extra
+    // watch anyway, since `.git/HEAD` then holds the sha directly and changes
+    // with it. Dedup below drops the repeat.
+    if let Some(branch) = git(&["rev-parse", "--symbolic-full-name", "HEAD"]) {
+        if let Some(path) = git(&["rev-parse", "--git-path", &branch]) {
+            watched.push(path);
+        }
+    }
+
+    watched.sort();
+    watched.dedup();
+    for path in watched {
+        if Path::new(&path).exists() {
+            println!("cargo:rerun-if-changed={path}");
+        }
+    }
+}
+
+/// Run a git command, returning its trimmed stdout when it succeeds with
+/// output. `None` covers git being absent as well as a failed invocation.
+fn git(args: &[&str]) -> Option<String> {
+    let output = Command::new("git").args(args).output().ok()?;
+    if !output.status.success() {
+        return None;
+    }
+
+    let stdout = String::from_utf8(output.stdout).ok()?.trim().to_string();
+    (!stdout.is_empty()).then_some(stdout)
 }
--- a/src/utils/remote.rs
+++ b/src/utils/remote.rs
@@ -242,11 +242,7 @@ impl SshClient {
             );
         }
 
-        // Parse version from output like "avocado 0.20.0"
-        let remote_version = version_output
-            .split_whitespace()
-            .last()
-            .unwrap_or(&version_output);
+        let remote_version = parse_reported_version(&version_output);
 
         // Compare versions
         if !is_version_compatible(local_version, remote_version) {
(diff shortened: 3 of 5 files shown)


jetm:
Companion PR: avocado-linux/avocado-desktop#73 fixes the other consumer of this line (`parse_cli_version`), which otherwise reads the commit date as the installed CLI version.

Suggested order: merge the desktop one first. It accepts both the old bare `avocado 0.40.0` and the new `avocado 1.0.0-rc.1 (abc1234 2026-03-05)` shapes, so it is correct against today's released CLI too, and landing it first avoids a window where the Updates pane misreports for anyone running a CLI built from this branch.

build.rs
```diff
@@ -1,17 +1,69 @@
...
+        _ => env!("CARGO_PKG_VERSION").to_string(),
+    };
+    println!("cargo:rustc-env=AVOCADO_CLI_VERSION={version}");
+
     println!("cargo:rustc-env=TARGET={}", env::var("TARGET").unwrap());
 }
+
+/// Rebuild when HEAD moves. Without these, cargo caches the build-script
+/// output and the reported commit goes stale after the next commit or checkout.
+fn emit_git_rerun_hints() {
```
jetm, line 71: **A packed branch ref makes the embedded sha permanently stale.**

The watch set is `.git/HEAD` plus the *loose* branch-ref file, and the loose one is pushed only `if Path::new(&path).exists()`. `git gc`, `git pack-refs`, or `git maintenance` moves the ref into `packed-refs` and deletes the loose file - after which the only watched path is `.git/HEAD`, whose contents a commit never changes. Cargo then never re-runs this script and the binary keeps reporting an older commit indefinitely.

Reproduced end to end in a scratch crate with this `build.rs` copied verbatim: `git pack-refs --all`, build gives `0.1.0 (4f9e121 ...)`; commit; rebuild still says `4f9e121` while `git rev-parse --short HEAD` is `4387475`; a third build does not self-heal.

The comment just below already names the trigger ("while the ref stays packed") but stops short of the consequence, which is why it reads as a caveat rather than a bug. `packed-refs` is not the fix either - its mtime does not change on commit. Watching `git rev-parse --git-path logs/HEAD` does: every commit appends to the reflog. Adding it made the next build report the right sha.

Secondary cost: `tests/version.rs` asserts the exact version line, so after a routine `git gc` a developer gets a version-mismatch test failure that reads like a broken assertion rather than a stale build cache.

src/utils/remote.rs
```diff
@@ -963,6 +959,16 @@ pub async fn get_local_ip_for_remote(remote_host: &str) -> Result<IpAddr> {
...
 }
 
+/// Extract the version from an `avocado --version` line.
+///
+/// The line reads `avocado <version> (<short-sha> <commit-date>)`, so the
+/// version is the second field. Reading the last field would pick up the commit
+/// date instead. Falls back to the whole string so a remote that prints a bare
+/// version still resolves.
+fn parse_reported_version(line: &str) -> &str {
+    line.split_whitespace().nth(1).unwrap_or(line)
```
jetm: **Positional `nth(1)` misreads banner-prefixed remote output, and fails open.**

`version_output` is the entire stdout blob of `test -f ~/.profile && . ~/.profile; test -f ~/.bashrc && . ~/.bashrc; avocado --version` - not one line. Anything the remote's profile echoes to stdout shifts the field index.

Executed with both parsers plus `is_version_compatible` copied from `origin/main`: input `"Welcome to the board!\navocado 1.0.0-rc.1 (abc1234 2026-03-05)"` gives `nth(1)` = `"to"`, and `is_version_compatible("1.0.0", "to")` returns `true` because it fails open on unparseable input. So the older-remote guard is silently skipped and `runs_on.rs:123` prints `to` as "Remote avocado version". `"bash: avocado: command not found"` yields `"avocado:"` and also passes.

The part that makes this worth fixing here rather than filing separately: for a remote still on a bare-version CLI, the old `.last()` handled the banner case correctly (`"0.41.2"`, gate fires). So this is a regression for that combination, not a pre-existing weakness carried forward.

Anchor on the line beginning `avocado `, or on the first token that parses as a version, instead of a fixed field index.

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
```
jetm, line 10: **The entry addresses future parsers but not the already-shipped ones, which is where the real exposure is.**

"Anything parsing this line must read the second whitespace-separated field" is right for anything written from here on. Every *already released* `avocado-cli` parses with `.last()` (the pre-change `remote.rs:238`) and cannot be updated retroactively.

Executed with the pre-change parser: `.last()` on `avocado 1.0.0-rc.1 (abc1234 2026-03-05)` gives `"2026-03-05)"`, and `is_version_compatible("1.0.0", "2026-03-05)")` returns `true`. So during the rollout window an older local CLI talking to a newer remote both prints a date as the remote version *and* stops enforcing the minimum remote version - it fails open rather than loudly.

The same applies to avocado-desktop's `parse_cli_version`, which the doc comment at `build.rs:12` names but this entry does not.

Two ways out: say so plainly in the entry so the rollout order is a conscious decision, or sidestep it entirely by putting the build detail on a second output line, which no `.last()` parser would ever reach.

src/utils/remote.rs
```diff
@@ -1065,6 +1071,20 @@ mod tests {
         assert!(!is_version_compatible("0.20.1", "0.20.0"));
     }
 
+    #[test]
```
jetm, line 1115: **The parser test covers three well-formed inputs and no negative path.**

The reversion check does work - with `.last()` restored, the 4-field case yields `2026-03-05)` and the test fails - so the happy path is genuinely pinned. What is missing is exactly where the positional parser breaks: multi-line or banner-prefixed output, output from a different binary, and any assertion that a non-version return value is *rejected* rather than fail-open accepted.

That matters because the banner defect above ships with a fully green suite.

Related gap in the new test file: `tests/version.rs:52` still passes if `build.rs` reverted to the older `"<version> <sha>"` shape, so the only test pinning the new format is `tests/version.rs:76` - and that one `return`s silently when git is unavailable, which turns the format assertion into a no-op that reports success on a runner without git.

`CLAUDE.md` is explicit here: "Edge cases are not optional: empty, single element, max size, malformed input" and "Test negative paths with equal rigor as happy paths".

build.rs
```diff
@@ -1,17 +1,69 @@
...
-    let git_hash = String::from_utf8(output.stdout).unwrap();
-    println!(
-        "cargo:rustc-env=AVOCADO_CLI_VERSION={} {}",
-        env!("CARGO_PKG_VERSION"),
-        git_hash
-    );
+    println!("cargo:rerun-if-changed=build.rs");
+    emit_git_rerun_hints();
+
+    let version = match (
```
jetm, line 21: **A vendored or unpacked crate can embed an unrelated repository's commit.** (PLAUSIBLE - reasoned, not executed against our release paths.)

`git` runs with the build script's cwd and nothing checks that the repository it finds is this crate's. A crate unpacked inside an unrelated git repo - a Yocto `WORKDIR` under version control, a docker build context inside a repo, a vendored copy in a monorepo - silently embeds that repo's sha and commit date.

That is precisely the case the doc comment presents as falling through to the bare version ("source tarball, vendored crate"). Instead of the intended fallback, a bug report cites a commit with nothing to do with the binary, and the string looks authoritative enough that nobody questions it.

Comparing `git rev-parse --show-toplevel` against `CARGO_MANIFEST_DIR` before trusting the output closes it.

jetm (commented):
Cold second read on my own PR. Six findings inline, two of them things I would want fixed before this ships.

The first is a stale-version bug rather than a style point: the rerun-if-changed set is `.git/HEAD` plus the loose branch ref, so once `git gc` or `git pack-refs` packs the ref, the only watched file is one that never changes on commit. Cargo stops re-running the build script and the binary reports an old commit permanently. Reproduced in a scratch crate. The comment already names the trigger and stops at the caveat; the reflog (`--git-path logs/HEAD`) is the path that actually changes on every commit.

The second is that `parse_reported_version` reads a fixed field index out of what is actually the whole stdout blob from a login shell. A remote whose profile echoes anything shifts the index, and `is_version_compatible` fails open, so the older-remote guard silently disappears. For a bare-version remote the old `.last()` got that case right, which makes it a regression rather than an inherited weakness.

One finding is out of the diff and so not anchorable: **`tests/runs_on_integration.rs:475` still carries its own copy of the old `.last()` logic**, then asserts the result `contains('.')`. Against the new line the token is `2026-03-05)`, which has no dot, so the assertion fails. It is `#[ignore]`d on SSH, so CI never runs it and this will surface as a confusing environment failure to whoever next runs `--ignored` on a box with a git-built avocado. It is also the last `split_whitespace` version parser left in the tree, so fixing it closes the set.

Sequencing note, since this cuts across repos: this PR is what introduces the `(<sha> <date>)` format, and `avocado-desktop#73` is the fix for the desktop side reading it. Per the CHANGELOG finding, the exposure is wider than that one app - every already-released CLI parses with `.last()` and fails open on the result, so the rollout order is a real decision rather than a detail. Worth settling before either side merges. Related: #186 and desktop#73 now contain two independent parsers for the same wire format in two repos, and this one has the better edge-case coverage of the two.

Checked and deliberately not reported: empty and whitespace-only output cannot reach the parser (the `is_empty()` / `"not-installed"` bail at `remote.rs:236` guards it); a `v` prefix fails open exactly as before and no avocado build emits one; `#[command(version = env!(...))]` disturbs no other consumer (every other site reads `CARGO_PKG_VERSION` directly); shallow and detached CI clones both produce a correct sha and date; and a dirty tree embeds HEAD's sha with no `-dirty` marker, which matches rustc.

build.rs
```diff
@@ -1,17 +1,69 @@
...
+/// Rebuild when HEAD moves. Without these, cargo caches the build-script
+/// output and the reported commit goes stale after the next commit or checkout.
+fn emit_git_rerun_hints() {
+    let mut watched = Vec::new();
+    if let Some(head) = git(&["rev-parse", "--git-path", "HEAD"]) {
+        watched.push(head);
+    }
+    // Follow the branch ref as well. HEAD itself is a symref whose contents do
+    // not change when you commit, so watching it alone catches a checkout but
+    // never a commit. Absent on a detached HEAD, and while the ref stays packed.
```
jetm: **The detached-HEAD half of this comment is wrong, and the entry duplicates.**

`git rev-parse --symbolic-full-name HEAD` prints `HEAD` on a detached checkout and exits 0 - it is not absent. So `--git-path HEAD` resolves and `.git/HEAD` gets pushed onto `watched` twice. Verified in a scratch repo after `git checkout --detach HEAD`.

The duplicate `rerun-if-changed` is harmless. The comment is not: a detached checkout actually works *correctly*, because `.git/HEAD` holds the sha directly and changes with it. Anyone debugging the stale-version behaviour above will read this and rule out the CI path for the wrong reason.

build.rs
jetm: Fixed in 3e9f3a0 - `logs/HEAD` is now in the watch set, exactly as you found. The reflog is appended to on every commit and is never packed away, which neither `.git/HEAD` nor the loose branch ref can claim.

Reproduced both directions in a scratch crate with refs packed (`git pack-refs --all`, zero loose refs):

```
OLD build.rs   build1: a7d032b  real=a7d032b
               build2: a7d032b  real=6d83a8e
               build3: a7d032b  real=cfe4ef5   <- never self-heals
NEW build.rs   build1: 9de937d  real=9de937d
               build2: 7bc2762  real=7bc2762
```

The comment now names the consequence rather than stopping at the caveat, and says why `packed-refs` is not the fix. The branch ref stays in the set for the case the reflog is disabled (`core.logAllRefUpdates=false`), with the list deduped.

src/utils/remote.rs
jetm: Fixed in 3e9f3a0. `parse_reported_version` now anchors on the line whose first field is `avocado` and takes its second field, falling back to the first version-shaped token anywhere in the blob.

That fallback is specifically to restore the case you identified as a regression: a banner in front of a bare version, which the old `.last()` got right and a positional read does not.

Your falsifier is now a test. Reverting to `output.split_whitespace().nth(1)` fails `test_parse_reported_version_skips_a_login_banner` with `left: "to"` - the exact value from your report - and `test_parse_reported_version_does_not_invent_a_version_from_prose` with `"avocado:"`.

I did **not** change the fail-open half here. `is_version_compatible` returning `true` for unparseable input is fixed on #187, which makes it `Option<bool>` and has the caller act on the three cases; duplicating that here would collide on merge. So this PR fixes the misparse and #187 fixes what happens when a parse fails.

CHANGELOG.md
jetm: Fixed in 3e9f3a0 - the entry now carries an explicit rollout note, so the ordering is a stated decision rather than an implicit one.

Two corrections to my own finding while writing it, both of which narrow it:

**avocado-desktop is not exposed.** I said the same applied to its `parse_cli_version`. It does not - on `origin/main` that parser already reads `stdout.lines().next()` then `nth(1)`, so it takes the second field of the first line and was never a `.last()` parser. Only avocado-cli's own shipped parser is affected. The entry says so.

**The second-line sidestep does not work.** I offered it as the clean way out. It is not: the shipped parser builds `version_output` from the *entire* stdout blob and then calls `.last()` on it, so a build detail on line two is still the last token it sees. It would have helped a first-line-then-`.last()` parser, which is not what shipped.

That leaves documenting it as the only real option, which is what landed: upgrade the local side first, and the note says plainly that during the window an older local CLI takes a date as the remote version and skips the minimum-version check rather than refusing.

src/utils/remote.rs
jetm: Fixed in 3e9f3a0, both gaps.

Added the negative paths: a banner-prefixed multi-line blob, prose with no version anywhere (asserted to return something the compatibility check will still reject, not a word from position 1), output from a different binary, and a direct table for `is_version_shaped` over version-like and word-like tokens.

On `tests/version.rs`: `version_field_is_the_bare_crate_version` now also requires any build detail to be parenthesized. It previously passed for the older `<version> <sha>` shape - the one shape where a `.last()` parser reads a sha as the version - so a revert to it went uncaught. Reverting `build.rs` to that shape now fails **both** tests in the file.

And `:76` no longer `return`s when git is unavailable. It asserts the bare form on that branch instead, so the no-git path is a real assertion rather than a no-op reporting success. Simulated it by pointing the helper at a nonexistent binary while the built binary still carried its build detail: the test fails with `with no git available the version line must be bare`, where before it would have passed.

build.rs
jetm: Fixed in 3e9f3a0 - and I executed it rather than leaving it reasoned, since I had marked it PLAUSIBLE.

Built the case: an unrelated git repo with the crate unpacked at `vendor/scratchver/` and no `.git` of its own.

```
outer repo HEAD: b56c95f
OLD build.rs -> 0.1.0 (b56c95f 2026-08-03)   <- another project's commit
NEW build.rs -> 0.1.0
```

So it is confirmed, not plausible. `in_own_repository()` compares `git rev-parse --show-toplevel` against `CARGO_MANIFEST_DIR` (canonicalized, since either can arrive symlinked); this crate sits at its repository root, so an exact match is the right test. A mismatch routes to the same arm as "git absent" and reports the bare version.

Verified the real build is unaffected: the binary still reports `avocado 1.0.0-rc.1 (730d38b 2026-07-31)`, matching `git rev-parse --short HEAD`.

build.rs
jetm: Fixed in 3e9f3a0. You were right on both counts and I have rewritten the comment: `--symbolic-full-name HEAD` prints `HEAD` on a detached checkout and exits 0, so it resolves back to `.git/HEAD` and the entry duplicated.

The comment now says the detached case works correctly and why - `.git/HEAD` holds the sha directly there, so it changes with every commit and needs no extra watch. That was the actively misleading part, since the previous wording would have sent anyone debugging the staleness above down the wrong path.

Deduped the list too (`sort` + `dedup`), which is cosmetic but stops the repeat from reappearing in the emitted directives.

mobileoverlord (approved):
Reviewed and tested locally. Approving.

Verified on the PR head merged with current `main`:
- `cargo test --all` green: 1120 + 1128 + 126 + 20 + 5 + 49 + 6 + 7 + 2 + 7, zero failures
- `--version` prints `avocado 1.0.0-rc.1 (e3e47e4 2026-08-06)`, where `e3e47e4` is the local merge commit rather than a cached one — which is the real check that the `rerun-if-changed` fix works, since the stale-sha bug would have reported the branch tip instead.

I confirmed the blast radius independently: `AVOCADO_CLI_VERSION` has exactly one consumer, `main.rs`'s `#[command(version = ...)]`. Everything that compares or records a version — `upgrade.rs`, `update_check.rs`, `utils/version.rs`, `stamps.rs`, `save.rs`/`load.rs`, `snapshot.rs`, `vm_update_check.rs`, and `remote.rs`'s `local_version` — reads `CARGO_PKG_VERSION` directly and is unaffected. avocado-desktop#73 is merged, so the desktop parser already reads the first line's second field.

`in_own_repository()` is a nice catch — a vendored copy under an unrelated repo (or a Yocto `WORKDIR` that is itself version-controlled) would otherwise embed a foreign sha behind an authoritative-looking string.

One thing for release notes, already documented in the CHANGELOG's rollout note and not a blocker: every already-released CLI parses the remote's version with `.last()` over the whole stdout blob, so an **older local** CLI talking to a **newer remote** will read `2026-08-06)`, fail to parse, and fall open — skipping the minimum-remote-version check silently. Nothing shippable fixes those binaries retroactively, so the "upgrade the local side first" guidance needs to be loud in the RC notes. #187 is what makes that skip visible going forward.

PR #187, opened 2026-08-03
fix(remote): stop a skipped version check from looking like a passed one

jetm:
## Problem

`is_version_compatible` returned `true` when either version failed to parse, so
"the remote is new enough" and "I could not tell" were the same answer. The
version gate in `RemoteHost`'s runs-on path then proceeded without printing
anything.

The case the gate exists to catch - a remote CLI too old for the command about
to run - is also the case most likely to report a version this parser cannot
read, so the check was most likely to be skipped exactly when it mattered. It
was skipped silently: nothing in the output distinguished a verified-compatible
remote from one that was never checked.

## Approach

Return `Option<bool>` so the caller can tell the two apart, and have it warn and
carry on rather than bail.

Proceeding on `None` is deliberate. Refusing to run because a remote printed an
unusual version string would break working setups to guard against a
hypothetical, and the complaint here is the silence, not the permissiveness. The
warning names the string that could not be read and the version the remote
needs, so an odd later failure has a breadcrumb.

If you would rather this fail closed, that is a one-line change at the call site
(`None` arm bails instead of warning) - say so and I will switch it. I went with
warn-and-proceed because it cannot break a working deploy.

`test_version_compatible_unparseable` asserted the old behaviour with the
comment "should fail open (assume compatible)". Fail-open was a deliberate,
documented, tested choice rather than an oversight, so that test is **flipped
rather than deleted** - the replacement records that the decision was
reconsidered, not that it never existed. A paired test pins `Some(true)` against
`Some(false)` so a caller matching on the result cannot collapse unknown into
either.

## Scope note

The desktop half of this failure mode is
avocado-linux/avocado-desktop#75 - `parse_cli_version` there accepted any
non-empty token, so a different program named `avocado` cleared the desktop's
`min_cli_version` floor. The two repos need different answers (the desktop can
fail closed without breaking anything; this side should not), which is why they
are separate changes rather than one.

This does not touch `parse_reported_version`, which is only on #186's branch.

## Verification

```
cargo fmt --all -- --check
cargo clippy --all-targets --all-features -- -D warnings
cargo test
```

All green at the tip: 1115 lib + 1123 + 126 + 20 + 5 + 49 across the test
binaries.

Falsified: restoring the `_ => Some(true)` fail-open makes
`test_version_unparseable_is_reported_as_unknown_not_compatible` fail, and only
that test.

### DIFF
--- a/src/utils/output.rs
+++ b/src/utils/output.rs
@@ -92,7 +92,7 @@ pub fn print_warning_above(message: &str) {
 /// Emit `formatted` where neither an active renderer nor the NDJSON stream can
 /// swallow it.
 ///
-/// The three-way routing had been open-coded four times (here,
+/// The three-way routing had been open-coded three times (here,
 /// [`crate::utils::container::print_failure_notice`], and
 /// `commands/runtime/build.rs`), each copy re-deriving the same policy. This is
 /// the one implementation; [`warning_sink`] stays separate as the pure,
--- a/src/utils/output.rs
+++ b/src/utils/output.rs
@@ -61,6 +61,58 @@ pub fn print_warning(message: &str, _level: OutputLevel) {
     }
 }
 
+/// Print a warning that an active renderer cannot swallow.
+///
+/// [`print_warning`] early-returns on [`tui_is_active`], which is also true for
+/// `--json`, so on the interactive renderer path and every JSON invocation it
+/// prints nothing at all. That is fine for progress chatter the task lines
+/// already convey, but not for a notice that a safety check was skipped: those
+/// are exactly the paths where the user would otherwise see an unqualified
+/// green run.
+///
+/// Routes through the renderer's `print_above` when one is active so the notice
+/// lands above the task list instead of being overwritten by it. With no
+/// renderer but JSON output active, goes to stderr - stdout carries the NDJSON
+/// stream and prose would corrupt it.
+pub fn print_warning_above(message: &str) {
+    let formatted = format!("\x1b[93m[WARNING]\x1b[0m {message}");
+    match warning_sink(
+        crate::utils::tui::get_active_renderer().is_some(),
+        crate::utils::output_format::is_json_output_active(),
+    ) {
+        WarningSink::Renderer => {
+            // Unwrap is sound: the sink is Renderer only when one is active, and
+            // nothing clears it mid-call.
+            crate::utils::tui::get_active_renderer()
+                .expect("renderer active")
+                .print_above(&formatted)
+        }
+        WarningSink::Stderr => eprintln!("{formatted}"),
+        WarningSink::Stdout => println!("{formatted}"),
+    }
+}
+
+/// Where a [`print_warning_above`] notice goes.
+///
+/// Split out from the emit so the routing policy is testable: the property that
+/// matters is that there is no fourth, suppressed variant, which is precisely
+/// what `print_warning` has and what made the skipped-check notice invisible on
+/// the default paths.
+#[derive(Debug, PartialEq, Eq)]
+pub enum WarningSink {
+    Renderer,
+    Stderr,
+    Stdout,
+}
+
+pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
+    match (renderer_active, json_active) {
+        (true, _) => WarningSink::Renderer,
+        (false, true) => WarningSink::Stderr,
+        (false, false) => WarningSink::Stdout,
+    }
+}
+
 /// Print a message without any color formatting.
 /// Suppressed when TUI is active.
 #[allow(dead_code)]
@@ -131,5 +183,40 @@ mod tests {
         print_debug("Test debug", OutputLevel::Normal);
         flush_stdout();
         flush_stderr();
+    }
+
+    #[test]
+    fn warning_above_is_never_suppressed() {
+        // `tui_is_active()` is true for BOTH an active renderer and `--json`,
+        // and those are the two states `print_warning` drops the message in -
+        // which is every interactive run and every JSON run. Each of the four
+        // combinations has to reach a real sink.
+        for (renderer, json) in [(true, true), (true, false), (false, true), (false, false)] {
+            let sink = warning_sink(renderer, json);
+            assert!(
+                matches!(
+                    sink,
+                    WarningSink::Renderer | WarningSink::Stderr | WarningSink::Stdout
+                ),
+                "renderer={renderer} json={json} produced {sink:?}"
+            );
+        }
+    }
+
+    #[test]
+    fn warning_above_goes_through_the_renderer_when_one_is_active() {
+        // A plain println would be painted over by the task list on the next
+        // frame, so an active renderer has to take precedence over the JSON
+        // check rather than the other way round.
+        assert_eq!(warning_sink(true, false), WarningSink::Renderer);
+        assert_eq!(warning_sink(true, true), WarningSink::Renderer);
+    }
+
+    #[test]
+    fn warning_above_keeps_json_stdout_clean() {
+        // stdout carries the NDJSON stream; prose on it would break consumers
+        // parsing line by line.
+        assert_eq!(warning_sink(false, true), WarningSink::Stderr);
+        assert_eq!(warning_sink(false, false), WarningSink::Stdout);
     }
 }
--- a/src/utils/output.rs
+++ b/src/utils/output.rs
@@ -75,20 +75,48 @@ pub fn print_warning(message: &str, _level: OutputLevel) {
 /// renderer but JSON output active, goes to stderr - stdout carries the NDJSON
 /// stream and prose would corrupt it.
 pub fn print_warning_above(message: &str) {
-    let formatted = format!("\x1b[93m[WARNING]\x1b[0m {message}");
-    match warning_sink(
-        crate::utils::tui::get_active_renderer().is_some(),
-        crate::utils::output_format::is_json_output_active(),
-    ) {
-        WarningSink::Renderer => {
-            // Unwrap is sound: the sink is Renderer only when one is active, and
-            // nothing clears it mid-call.
-            crate::utils::tui::get_active_renderer()
-                .expect("renderer active")
-                .print_above(&formatted)
-        }
-        WarningSink::Stderr => eprintln!("{formatted}"),
-        WarningSink::Stdout => println!("{formatted}"),
+    // stdout carries the NDJSON stream, so the human line goes to stderr under
+    // --json (see print_notice_above). Without an event too, a consumer reading
+    // only the stream sees nothing at all and renders an unqualified green run -
+    // which is the failure this whole notice exists to prevent, just moved.
+    if crate::utils::output_format::is_json_output_active() {
+        crate::utils::output_format::emit_json_event(
+            &serde_json::json!({"event": "warning", "message": message}),
+        );
+    }
+    print_notice_above(&format!("\x1b[93m[WARNING]\x1b[0m {message}"), |line| {
+        println!("{line}")
+    });
+}
+
+/// Emit `formatted` where neither an active renderer nor the NDJSON stream can
+/// swallow it.
+///
+/// The three-way routing had been open-coded four times (here,
+/// [`crate::utils::container::print_failure_notice`], and
+/// `commands/runtime/build.rs`), each copy re-deriving the same policy. This is
+/// the one implementation; [`warning_sink`] stays separate as the pure,
+/// testable statement of that policy.
+///
+/// `plain` is the emit for the neither-active case and is a parameter because
+/// severity genuinely differs there: a warning belongs on stdout, an error on
+/// stderr. Everything above that case is identical, which is the part worth
+/// sharing.
+///
+/// The renderer `Option` is bound once and matched alongside the sink rather
+/// than looked up a second time behind an `expect`. `get_active_renderer`
+/// returning `None` on the second call is reachable - `clear_active_renderer`
+/// is `pub` and is the first statement of `TaskRenderer::shutdown` - so the
+/// old shape could abort the process while printing an informational notice.
+/// Pairing the two makes that branch fall through to stderr instead of
+/// depending on the interleaving being impossible.
+pub fn print_notice_above(formatted: &str, plain: impl FnOnce(&str)) {
+    let renderer = crate::utils::tui::get_active_renderer();
+    let json_active = crate::utils::output_format::is_json_output_active();
+    match (warning_sink(renderer.is_some(), json_active), renderer) {
+        (WarningSink::Renderer, Some(active)) => active.print_above(formatted),
+        (WarningSink::Renderer, None) | (WarningSink::Stderr, _) => eprintln!("{formatted}"),
+        (WarningSink::Stdout, _) => plain(formatted),
     }
 }
 
@@ -191,16 +219,15 @@ mod tests {
         // and those are the two states `print_warning` drops the message in -
         // which is every interactive run and every JSON run. Each of the four
         // combinations has to reach a real sink.
-        for (renderer, json) in [(true, true), (true, false), (false, true), (false, false)] {
-            let sink = warning_sink(renderer, json);
-            assert!(
-                matches!(
-                    sink,
-                    WarningSink::Renderer | WarningSink::Stderr | WarningSink::Stdout
-                ),
-                "renderer={renderer} json={json} produced {sink:?}"
-            );
-        }
+        //
+        // Asserting the exact sink per combination, not merely that the result
+        // is one of the three variants: a three-variant enum makes that weaker
+        // form unfalsifiable, so it stayed green under a `warning_sink` forced
+        // to return a single arm.
+        assert_eq!(warning_sink(true, true), WarningSink::Renderer);
+        assert_eq!(warning_sink(true, false), WarningSink::Renderer);
+        assert_eq!(warning_sink(false, true), WarningSink::Stderr);
+        assert_eq!(warning_sink(false, false), WarningSink::Stdout);
     }
 
     #[test]
--- a/src/utils/remote.rs
+++ b/src/utils/remote.rs
@@ -204,14 +204,10 @@ impl SshClient {
             // printed "[SUCCESS] Remote avocado version: <local>" for a check
             // that never ran - the precise defect Skipped exists to remove.
             //
-            // The warning is empty on purpose. Same machine, same binary, so
-            // there is nothing for the user to act on; an empty warning renders
-            // as no notice at all rather than a [WARNING] line on every
+            // No warning: same machine, same binary, so there is nothing for
+            // the user to act on and nothing to print on every
             // `--runs-on localhost` run.
-            return Ok(RemoteVersionCheck::Skipped {
-                reported: local_version.to_string(),
-                warning: String::new(),
-            });
+            return Ok(RemoteVersionCheck::Skipped(None));
         }
 
         if self.verbose {
@@ -279,10 +275,7 @@ impl SshClient {
             // Proceed, but hand the caller the notice rather than swallowing it.
             // Refusing here would break a working setup whose remote merely
             // prints an unusual version string.
-            VersionGate::Unreadable(warning) => Ok(RemoteVersionCheck::Skipped {
-                reported: remote_version.to_string(),
-                warning,
-            }),
+            VersionGate::Unreadable(warning) => Ok(RemoteVersionCheck::Skipped(Some(warning))),
         }
     }
 
@@ -988,9 +981,9 @@ pub async fn get_local_ip_for_remote(remote_host: &str) -> Result<IpAddr> {
 pub enum RemoteVersionCheck {
     /// The remote reported a parseable version at least as new as the local one.
     Verified(String),
-    /// The remote's `--version` output could not be parsed, so no comparison
-    /// happened. Carries what it reported and the notice to show for it.
-    Skipped { reported: String, warning: String },
+    /// No comparison happened. Carries the notice to show for it, or `None`
+    /// when there is nothing worth saying (localhost).
+    Skipped(Option<String>),
 }
 
 /// The three ways the version gate can land, with the prose for the two that
@@ -1038,8 +1031,8 @@ pub fn version_check_notice(check: &RemoteVersionCheck) -> VersionNotice {
         RemoteVersionCheck::Verified(version) => {
             VersionNotice::Success(format!("Remote avocado version: {version}"))
         }
-        RemoteVersionCheck::Skipped { warning, .. } if warning.is_empty() => VersionNotice::Silent,
-        RemoteVersionCheck::Skipped { warning, .. } => VersionNotice::Warning(warning.clone()),
+        RemoteVersionCheck::Skipped(None) => VersionNotice::Silent,
+        RemoteVersionCheck::Skipped(Some(warning)) => VersionNotice::Warning(warning.clone()),
     }
 }
 
@@ -1393,10 +1386,8 @@ mod tests {
         // The whole point of the Skipped variant. Reverting runs_on's arm to
         // print_success left the suite green because nothing tested the
         // mapping, so the bug this PR fixes could return unobserved.
-        let skipped = RemoteVersionCheck::Skipped {
-            reported: "avocado (dev build)".to_string(),
-            warning: "the version check was skipped".to_string(),
-        };
+        let skipped =
+            RemoteVersionCheck::Skipped(Some("the version check was skipped".to_string()));
         assert_eq!(
             version_check_notice(&skipped),
             VersionNotice::Warning("the version check was skipped".to_string())
@@ -1416,10 +1407,7 @@ mod tests {
         // The localhost path. It has no version to report and nothing to warn
         // about, so it must produce neither a green line nor a [WARNING] on
         // every `--runs-on localhost` run.
-        let localhost = RemoteVersionCheck::Skipped {
-            reported: "0.21.0".to_string(),
-            warning: String::new(),
-        };
+        let localhost = RemoteVersionCheck::Skipped(None);
         assert_eq!(version_check_notice(&localhost), VersionNotice::Silent);
     }
 }
--- a/src/utils/remote.rs
+++ b/src/utils/remote.rs
@@ -174,13 +174,22 @@ impl SshClient {
         Ok(())
     }
 
-    /// Check that the remote avocado CLI version is compatible
+    /// Check that the remote avocado CLI version is compatible.
     ///
-    /// The remote version must be equal to or greater than the local version.
-    /// Returns the remote version string if compatible.
+    /// Returns [`RemoteVersionCheck::Verified`] when the remote reported a
+    /// parseable version at least as new as the local one, and
+    /// [`RemoteVersionCheck::Skipped`] when its `--version` output could not be
+    /// read, so no comparison happened. Errors when the remote is genuinely
+    /// older, or when the CLI is missing there entirely.
+    ///
+    /// The two `Ok` cases stay distinct all the way to the caller on purpose. A
+    /// `Result<String>` collapsed them back into one indistinguishable value, so
+    /// the caller printed the same green success line for a skipped check as for
+    /// a passed one - which is the failure this whole three-valued path exists
+    /// to prevent.
     ///
     /// For localhost/127.0.0.1, this check is skipped since it's the same machine.
-    pub async fn check_cli_version(&self) -> Result<String> {
+    pub async fn check_cli_version(&self) -> Result<RemoteVersionCheck> {
         let local_version = env!("CARGO_PKG_VERSION");
 
         // Skip version check for localhost - it's the same machine
@@ -191,7 +200,7 @@ impl SshClient {
                     OutputLevel::Normal,
                 );
             }
-            return Ok(local_version.to_string());
+            return Ok(RemoteVersionCheck::Verified(local_version.to_string()));
         }
 
         if self.verbose {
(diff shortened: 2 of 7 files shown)


src/utils/remote.rs
```diff
@@ -249,15 +249,34 @@ impl SshClient {
...
+                    local_version,
+                    self.remote.ssh_target(),
+                    local_version
+                );
+            }
+            // Proceed, but say so. Refusing here would break a working setup
+            // whose remote merely prints an unusual version string; staying
+            // silent is what let a skipped check look like a passed one.
+            None => {
+                print_warning(
```
jetm: **`print_warning` is swallowed whenever the TUI is active, so the default paths still say nothing.**

`print_warning` early-returns on `tui_is_active()` (`output.rs:59`), and `tui_is_active()` is also true for `--json` (`output.rs:26-28`). So in the interactive renderer path and every `--json` invocation, the skipped check produces no output at all - the exact silence this change sets out to remove.

Traced concretely for `avocado install --runs-on user@host --force` on a TTY, non-verbose: `install.rs:179-187` evaluates `should_create_renderer() && !self.verbose && self.force` to true and calls `set_active_renderer(&r)`. Only afterwards does the remote work reach `check_cli_version` (`install.rs:214` `.with_runs_on(...)`, executed at `:223`). Remote prints `avocado (dev build)`, last token `build)`, unparseable, `None` arm runs, `print_warning` hits the guard and returns without printing. The renderer paints a normal green checklist and the user sees nothing.

Same outcome for `avocado build --runs-on host --no-stamps`, which skips the pre-renderer pre-flight at `build.rs:177` so the first version check lands after `build.rs:253`. Only the `build.rs` pre-flight stamp path and verbose/CI/non-TTY runs actually show the warning today.

This needs a renderer-aware emit (or routing the notice through whatever the active renderer uses for warnings), not `print_warning`.

Second, smaller point on the same arm: this warn-and-proceed branch is the only genuinely new behaviour here and it has no test. Replace the whole body with `None => {}` and both `cargo test --lib` and `tests/runs_on_integration.rs` still pass - which is also how the finding above got through. There is no existing seam for it (no test constructs an `SshClient` or fakes the ssh boundary), so closing it means extracting the three arms into a pure `fn version_gate_outcome(..) -> Outcome` and testing that, rather than just adding a case.

src/utils/remote.rs
```diff
@@ -249,15 +249,34 @@ impl SshClient {
...
+                        "Could not read the avocado version on '{}' from '{}', so the \
+                         version check was skipped. If this command fails oddly, confirm \
+                         the remote has avocado {} or later.",
+                        self.remote.ssh_target(),
+                        remote_version,
+                        local_version
+                    ),
+                    OutputLevel::Normal,
+                );
+            }
```
jetm: **The caller prints `[SUCCESS]` for a skipped check, so even where the warning shows, the skip reads as a pass.**

After this arm falls through, `check_cli_version` returns `Ok(remote_version)` at `:289`, collapsing `Some(true)` and `None` back into one indistinguishable value. Its only caller, `RunsOnContext::setup` (`runs_on.rs:122`), then unconditionally prints a green success line with the unparsed string at `runs_on.rs:123-126`.

So for `avocado build --runs-on user@host` with `CI=1` (where the warning does print): remote reports `avocado v0.20-dev`, last token `v0.20-dev`, `None`. The output is a yellow "check was skipped" line immediately followed by `[SUCCESS] Remote avocado version: v0.20-dev` in green - directly contradictory, and the green line is the one that survives a scroll-back skim.

The three-valued information this diff works to preserve is discarded at the `Result<String>` boundary, so no present or future caller can distinguish verified from skipped. The doc comment at `:180` ("Returns the remote version string if compatible") is inaccurate for the same reason. Propagating the outcome instead of re-flattening it is what makes the `Option<bool>` refactor pay off.

src/utils/remote.rs
```diff
@@ -1047,41 +1074,55 @@ mod tests {
...
+        assert_eq!(is_version_compatible("unparseable", "0.20.0"), None);
+        assert_eq!(is_version_compatible("0.20.0", "unparseable"), None);
     }
 
     #[test]
-    fn test_version_compatible_unparseable() {
-        // Unparseable versions should fail open (assume compatible)
-        assert!(is_version_compatible("unparseable", "0.20.0"));
-        assert!(is_version_compatible("0.20.0", "unparseable"));
+    fn test_version_compatible_distinguishes_compatible_from_too_old() {
```
jetm: This test's two assertions already exist character-for-character elsewhere in the same module, so it can never fail unless an existing test fails first.

Line 1125 (`is_version_compatible("0.20.0", "0.21.0") == Some(true)`) is identical to line 1083 in `test_version_compatible_remote_newer`; line 1126 (`("0.21.0", "0.20.0") == Some(false)`) is identical to line 1090 in `test_version_incompatible_remote_older`.

Its comment claims it stops "a caller matching on the result" from collapsing the outcomes, but it exercises no caller - and the only caller, `check_cli_version`, is untested, which is where both of the findings above live. So the stated guarantee is not the one this test provides. A test that actually drove `version_gate_outcome` (or whatever the extracted seam ends up being) would earn its place; this one costs maintenance for zero added coverage.

jetm (commented):
Second-reviewer pass on my own change, cold - and the headline is that the change does not yet achieve what it sets out to do.

The refactor itself is right. `is_version_compatible` has exactly one non-test caller at this head (`check_cli_version`, `remote.rs:252`), which the diff correctly converts to a three-arm match, so nothing was left compiling against the old `bool`. `parse_version` is untouched, `Some(remote >= local)` preserves the old comparison exactly, and `local_version` is `env!("CARGO_PKG_VERSION")` so it always parses - meaning `None` really does imply "remote unparseable", which is what the warning text claims.

Where it falls down is one frame above, in both directions: the warning is emitted through `print_warning`, which is suppressed whenever a renderer or `--json` is active - i.e. the default interactive path - and the function then re-flattens `Some(true)` and `None` into the same `Ok(String)`, so its caller prints a green `[SUCCESS]` line for a check that was skipped. A skipped check still looks like a passed one in the common case; it just looks like one for a different reason than before. Two findings inline, plus a third on a test that duplicates two existing assertions verbatim.

Checked and not raising: echoing the untrusted remote version string unsanitized is not newly introduced (the `Some(false)` bail arm already interpolated it); the now-inaccurate `#[allow(dead_code)]` on `print_warning` sits on an untouched line in an untouched file; and the `bool` to `Option<bool>` signature change is technically breaking for the `pub` lib surface but this is a 0.x internal util with no external consumers.

Verification caveat: all findings are from reading and quoting the code at this head, not from a run - the clone's working tree is at a different commit and no finding depends on a test result. The two behavioural ones would need a live remote printing an unparseable version to observe end-to-end.

src/utils/remote.rs
jetm: Fixed in 9f86a02, both halves.

**The emit.** Added `print_warning_above` (`output.rs`) and routed the notice through it. It dispatches on `warning_sink(renderer_active, json_active)`: an active renderer takes precedence and gets `print_above`, so the line lands above the task list instead of being painted over on the next frame; with no renderer but `--json` active it goes to stderr, leaving the NDJSON stream on stdout parseable; otherwise stdout. There is deliberately no fourth suppressed variant - that is the whole defect - and `warning_above_is_never_suppressed` walks all four combinations of the two flags to pin it.

**The missing test.** Extracted `version_gate_outcome(local, remote, ssh_target) -> VersionGate`, pure over the three strings, exactly as you suggested. Your falsifier now bites: making the `None` arm return `Compatible` (the old silent fail-open) fails `test_version_gate_unreadable_warns_instead_of_staying_silent` and `test_version_gate_does_not_treat_unreadable_as_too_old`. It previously failed nothing.

What that test pins is that the arm *produces a notice*, not just that it proceeds - proceeding was never the problem.

src/utils/remote.rs
jetm: Fixed in 9f86a02. `check_cli_version` now returns `RemoteVersionCheck`, not `Result<String>`:

```rust
pub enum RemoteVersionCheck {
    Verified(String),
    Skipped { reported: String, warning: String },
}
```

`RunsOnContext::setup` matches on it and prints `[SUCCESS]` only for `Verified`; `Skipped` gets the warning and no success line, so the contradictory yellow-then-green pair is gone. The doc comment at the old `:180` was inaccurate for the same reason and now describes both `Ok` cases.

The warning text travels with the outcome rather than being printed inside `check_cli_version`, so the message is built where `ssh_target` and `local_version` live while the caller stays in charge of presentation - and it is emitted exactly once.

src/utils/remote.rs
jetm: Deleted in 9f86a02. You were right that it could not fail unless an existing test failed first.

Replaced with four tests that drive `version_gate_outcome`, the seam the old test's comment claimed to protect: that `Compatible` carries no prose, that `TooOld` names both versions and the host to upgrade, that `Unreadable` says "skipped" and quotes what the remote actually reported, and that `Unreadable` and `TooOld` stay distinct - a test asserting merely "not `Compatible`" would pass for either, yet one proceeds and one refuses.

mobileoverlord:
Reviewed this and it looks good — but #186 landed first and this now conflicts with `main` in `src/utils/remote.rs`. Could you rebase? I verified the resolution locally so it should be mechanical.

**The conflict is purely additive.** One region, where both branches appended new free functions at the same spot:
- your side: `RemoteVersionCheck`, `VersionGate`, `version_gate_outcome`
- `main` (from #186): `parse_reported_version`, `is_version_shaped`

Keep both blocks, each with its own closing braces — nothing needs to be reconciled semantically.

**The `check_cli_version` body auto-merges correctly**, and the composition is exactly what you'd want:

```rust
let remote_version = parse_reported_version(&version_output);
let gate = version_gate_outcome(local_version, remote_version, &self.remote.ssh_target());
```

#186's parser feeds your gate, so the `(<sha> <date>)` suffix is stripped to a bare semver before the comparison — the two changes reinforce each other rather than overlap. Worth double-checking that line survives your rebase, since a bad resolution there could silently restore the `.last()` read.

On the merged result (your branch + `main` at e3c39fa) I got:
- `cargo test --all` green: 1127 + 1135 + 126 + 20 + 5 + 49 + 6 + 7 + 2 + 7, zero failures

I'd push the resolution myself, but the ruleset has `require_last_push_approval`, so my approval wouldn't count afterward and it'd need a third person. Ping me once it's rebased and I'll approve and merge.

On the open question in your description — warn-and-proceed vs fail-closed: keep warn-and-proceed. #186's CHANGELOG rollout note makes this more than hypothetical. Every already-released CLI reads the remote version with `.last()` over the whole stdout blob, so an older local CLI against a post-#186 remote will now read `2026-08-06)` and fail to parse. Failing closed here would turn that into a hard stop for anyone who upgrades the remote first, which is the wrong failure for a version *hint*. The warning is the fix.

mobileoverlord:
I rebased, waiting for GH Actions to resume to merge.

src/utils/remote.rs
```diff
@@ -191,7 +200,7 @@ impl SshClient {
                     OutputLevel::Normal,
                 );
             }
-            return Ok(local_version.to_string());
+            return Ok(RemoteVersionCheck::Verified(local_version.to_string()));
```
jetm: **The localhost shortcut returns `Verified` with the local version**

This is the one path whose own log line above says "Skipping version check", and it returns `Verified(local_version)` - a value copied from the local binary and never observed on any remote.

`avocado install --runs-on localhost` prints `[SUCCESS] Remote avocado version: 1.0.0-rc.1` for a check that did not run. That is exactly the skipped-looks-like-passed shape this commit adds `RemoteVersionCheck::Skipped` to eliminate, and the diff touched this line and chose `Verified` anyway.

`Skipped` fits here: same machine, so there is nothing to warn about, but the output should not assert a version it never read.

src/utils/remote.rs
```diff
@@ -959,6 +968,59 @@ pub async fn get_local_ip_for_remote(remote_host: &str) -> Result<IpAddr> {
...
+    ssh_target: &str,
+) -> VersionGate {
+    match is_version_compatible(local_version, remote_version) {
+        Some(true) => VersionGate::Compatible,
+        Some(false) => VersionGate::TooOld(format!(
+            "Remote avocado version '{remote_version}' is older than local version \
+             '{local_version}'. Please upgrade avocado on '{ssh_target}' to version \
+             {local_version} or later."
+        )),
+        None => VersionGate::Unreadable(format!(
```
jetm, line 1058: **This arm is much harder to reach than the doc implies, because of the parser feeding it**

The new three-way gate is the right shape, but `Unreadable` fires only when `parse_reported_version` yields something unparseable - and that function (pre-existing, not touched here) adopts the first version-shaped token anywhere in the blob. Measured against the real functions with local `0.41.0`:

- `"Welcome to Ubuntu 22.04.3 LTS (GNU/Linux 5.15.0-1044-aws x86_64)\nnot-installed"` gives `Verified("22.04.3")` - a green `[SUCCESS] Remote avocado version: 22.04.3` against a host with **no avocado installed**.
- `"Kernel 6.6.1\n0.20.0"` gives `Verified("6.6.1")` where the remote is genuinely `0.20.0`. The pre-PR `.last()` bailed `TooOld` on that input.
- `"avocado sdk ready\navocado 0.20.0"` gives `"sdk"`, so `Unreadable` and warn-and-proceed, against a remote two dozen minors old that the old code refused.

So the enabler is the unchanged `version_output == "not-installed"` exact-equality guard at remote.rs:246: the premise that profile output precedes the version is precisely what defeats that equality.

None of that parsing is this PR's code, which is why I am raising it here rather than as a defect in the diff - but it decides whether the arm you just gave a warning to ever runs, and the new doc line "Errors ... when the CLI is missing there entirely" is not true today. Gating the anchored field on `is_version_shaped` and continuing to scan is a one-line change that makes the gate mean what it says.

src/utils/runs_on.rs
```diff
@@ -119,11 +120,17 @@ impl RunsOnContext {
...
+        match ssh.check_cli_version().await? {
+            RemoteVersionCheck::Verified(remote_version) => print_success(
+                &format!("Remote avocado version: {remote_version}"),
+                OutputLevel::Normal,
+            ),
+            // Never a green success line here: the check did not run. The notice
+            // goes through `print_warning_above` because `print_warning` is
+            // suppressed whenever a renderer or `--json` is active, which is the
+            // default for most of these commands.
+            RemoteVersionCheck::Skipped { warning, .. } => print_warning_above(&warning),
```
jetm: **The bug this commit fixes has no regression test**

I reverted this arm to the pre-PR behaviour - `Skipped { reported, .. } => print_success("Remote avocado version: {reported}")` - and the entire suite stayed green: 1127 lib tests and all 10 suites, with only an `unused import: print_warning_above` warning that a reviewer would silence by deleting the import.

`check_cli_version`'s three-way return and both new match arms have zero coverage.

This matters more than usual here because the commit's own commentary says the previous silent skip survived review for exactly this reason: deleting its body left the suites passing. The replacement inherits that property, so the next person to simplify this arm gets a green run.

jetm (commented):
Independent second read on my own PR, cold - seven findings inline. The three-way `RemoteVersionCheck` is the right shape and the `print_warning_above` routing is the right fix for an invisible warning; most of what follows is about the arms not being reachable or not being pinned.

The one I would fix before merge is the localhost shortcut returning `Verified` with the local version, because it reproduces the exact defect this commit exists to remove, on the one path whose log line already says it is skipping.

Two notes that are not inline comments.

**A test in a file this PR edits will fail once anyone runs it.** `tests/runs_on_integration.rs::test_localhost_cli_version_check` is `#[ignore]`d, still parses with `.split_whitespace().last()`, and asserts `version.contains('.')`. Against the current `--version` format the last token is `2026-08-06)`, which has no dot, so the only end-to-end test of the ssh `--version` path fails under `--ignored`. Its lines are not in a hunk, so it cannot take an inline comment.

**`--org` is still silently dropped on the `--token` login path.** The new comment and test declare the invariant "an explicit `--org` must be honored or fail loudly - never silently dropped", and that holds only for the browser path. `login_with_token` (auth.rs:213) never reads `self.org`, and `provision_org_token` takes `me.organizations.first()` (auth.rs:598). So `login --token T --org uuid-B` on an account with orgs [A, B] mints a token scoped to **A** and exits 0. Every defective line is unchanged context, so this is out of scope for the diff - but the invariant it states is what makes it worth writing down here rather than losing. Related, in the same file: `provision_org_token`'s unscoped-token fallback (auth.rs:609) hand-prefixes "Warning:" onto `print_info`, which `tui_is_active()` suppresses - the same invisibility bug this commit adds `print_warning_above` to fix, on a security-relevant fallback.

Also worth a CHANGELOG line: the remote-check behaviour change, and `is_version_compatible` going from `bool` to `Option<bool>`, which is breaking for consumers of the `avocado_cli` lib target.

Separately, and not about this PR: reviewing the rebased range surfaced findings on code already on `main` (the `--version` format, `parse_reported_version`, and build.rs's rerun hints). Those are not in this diff and I am not raising them here.

src/utils/output.rs
```diff
@@ -132,4 +184,39 @@ mod tests {
         flush_stdout();
         flush_stderr();
     }
+
+    #[test]
+    fn warning_above_is_never_suppressed() {
```
jetm, line 217: **This assertion cannot fail**

`assert!(matches!(sink, Renderer | Stderr | Stdout))` enumerates all three variants of a three-variant enum, so it is `assert!(true)` - the property is already enforced by the type system.

Demonstrated: I replaced `warning_sink`'s body with an unconditional `Stdout`, the exact regression the comment claims to guard against. The other two tests failed and this one passed (2 passed, 2 failed).

What would make it real: assert the specific sink per input combination, which is what the two passing tests beside it already do.

src/utils/output.rs
```diff
@@ -61,6 +61,58 @@ pub fn print_warning(message: &str, _level: OutputLevel) {
...
+/// prints nothing at all. That is fine for progress chatter the task lines
+/// already convey, but not for a notice that a safety check was skipped: those
+/// are exactly the paths where the user would otherwise see an unqualified
+/// green run.
+///
+/// Routes through the renderer's `print_above` when one is active so the notice
+/// lands above the task list instead of being overwritten by it. With no
+/// renderer but JSON output active, goes to stderr - stdout carries the NDJSON
+/// stream and prose would corrupt it.
+pub fn print_warning_above(message: &str) {
```
jetm, line 77: **This is the fourth copy of the same three-way router**

`crate::utils::container::print_failure_notice` (container.rs:405) already implements this identical dispatch, with the same rationale in its doc comment and 11 call sites - three of them in `src/utils/runs_on.rs` (464, 469, 481), the same file this PR adds a `print_warning_above` call to. `src/commands/runtime/build.rs:817` is a third inline copy.

Cost of the fourth: another router to keep in sync, and the `\x1b[93m[WARNING]\x1b[0m` literal duplicated from `print_warning` (output.rs:60).

Reusing the existing helper also removes the double renderer lookup and the `expect` below, so it collapses two of my other comments on this file.

src/utils/output.rs
```diff
@@ -61,6 +61,58 @@ pub fn print_warning(message: &str, _level: OutputLevel) {
...
+    let formatted = format!("\x1b[93m[WARNING]\x1b[0m {message}");
+    match warning_sink(
+        crate::utils::tui::get_active_renderer().is_some(),
+        crate::utils::output_format::is_json_output_active(),
+    ) {
+        WarningSink::Renderer => {
+            // Unwrap is sound: the sink is Renderer only when one is active, and
+            // nothing clears it mid-call.
+            crate::utils::tui::get_active_renderer()
+                .expect("renderer active")
```
jetm: **`expect("renderer active")` can fire, despite the comment**

`get_active_renderer()` is called twice - once inside `warning_sink(...)`, once here - over a global `Mutex<Option<Weak<TaskRenderer>>>`. The defending comment says nothing clears it mid-call, and that is not so: `tui::clear_active_renderer()` is `pub` and is the first statement of `TaskRenderer::shutdown()` (renderer.rs:307), reached from build.rs:388, install.rs:230, and the Ctrl-C path at scheduler/executor.rs:107.

If a shutdown lands between the two lookups, printing an informational notice aborts the process.

Marking this lower confidence because I could not construct the interleaving. Binding the `Option` once makes the panic branch unrepresentable regardless, so the fix does not depend on settling whether it is reachable.

src/utils/output.rs
```diff
@@ -61,6 +61,58 @@ pub fn print_warning(message: &str, _level: OutputLevel) {
...
+/// what `print_warning` has and what made the skipped-check notice invisible on
+/// the default paths.
+#[derive(Debug, PartialEq, Eq)]
+pub enum WarningSink {
+    Renderer,
+    Stderr,
+    Stdout,
+}
+
+pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
```
jetm, line 136: **In `--output json` the notice never reaches the NDJSON stream**

Every combination terminates on stderr - `warning_sink(true, true)` gives `Renderer`, then `print_above`, then `eprintln` - and `print_warning_above` emits no `emit_json_event`.

So `avocado install --runs-on user@host --output json` against an unreadable remote version sends zero events about the skipped check down the stream the desktop app drives its UI from. The human paths get the fix; the machine consumer named in the CHANGELOG does not, and it still renders an unqualified green run.

Lower confidence only on how much it matters - whether the desktop app would surface such an event today is a question about that repo, not this one. The asymmetry itself is certain.

jetm:
Correction to my review above. I listed "the `--version` format" among the findings my review turned up on `main`. That one was wrong, and I'd rather retract it here than leave it implying there's something to chase.

The concern was that `avocado <ver> (<sha> <date>)` breaks avocado-desktop's update check, which would have made this PR's CHANGELOG claim false. It doesn't: avocado-desktop#73 (`e7174e1`) already changed that parser from `.split_whitespace().last()` to `.nth(1)` and added the parenthetical form to its expected-formats comment. So `nth(1)` yields `1.0.0-rc.1`, the version comparison is correct, and **the CHANGELOG claim is accurate as written.**

The mistake was mine and mechanical: our local avocado-desktop clone sits on a feature branch four commits behind `main`, and I read the parser out of that working tree instead of at an explicit ref.

What does still stand on `main` from that same review, all minor and none of it cross-repo: `build.rs` emitting an explicit `rerun-if-changed` discards cargo's default whole-package watch and emits no git path when git fails or `.git` is absent, the `exists()` guard drops the branch-ref watch once `git pack-refs` runs, and `tests/version.rs` demands the parenthetical form whenever git answers, contradicting `build.rs`'s deliberate bare-version fallback for a vendored copy. Separate from this PR either way.

The seven inline findings above are unaffected.

src/utils/remote.rs
jetm: Fixed in 1eba3a8. Returns `Skipped` now, so the localhost path can no longer print a green line for a comparison that never happened.

On the open question of whether `warning` should be empty or carry a one-liner: empty, and `runs_on`'s arm treats an empty warning as silent. Same machine, same binary, so there is nothing for the user to act on, and a `[WARNING]` on every `--runs-on localhost` run is how you train people to ignore the notice on the runs where it does mean something.

`a_skipped_check_with_no_warning_says_nothing` pins that, so the localhost path cannot drift back into emitting a notice or a version line.

src/utils/runs_on.rs
jetm: Confirmed - I reproduced the revert before fixing it, and the whole suite stayed green.

The mapping now lives in `version_check_notice` (`remote.rs`) rather than inline in the arm, which is what makes it reachable from a test. Three tests over it: `a_skipped_check_never_renders_as_a_green_version_line`, `a_verified_check_renders_the_version_it_actually_read`, and `a_skipped_check_with_no_warning_says_nothing`. Also added `version_gate_reports_compatible_when_the_remote_is_new_enough` - the other two `version_gate_outcome` arms were covered and that one was not, so a gate that never returned `Compatible` would have gone unnoticed.

Falsifier run rather than described. Reverting `version_check_notice`'s `Skipped` arm to `Success(format!("Remote avocado version: {reported}"))` - the pre-PR behaviour - fails two tests:

    a_skipped_check_with_no_warning_says_nothing ... FAILED
    a_skipped_check_never_renders_as_a_green_version_line ... FAILED

One note on where the falsifier now points. Your original was "revert the arm in `runs_on.rs`"; the arm is now a total dispatch over `VersionNotice` and holds no policy, so the meaningful mutation is on `version_check_notice`. Reverting the arm itself would mean deleting a match arm, which does not compile.

1eba3a8.

src/utils/output.rs
jetm: Right, and the mutation proves it. `matches!` over all three variants of a three-variant enum is what the type system already guarantees, so the assertion had no failing input.

Now asserts the specific sink per combination:

    assert_eq!(warning_sink(true,  true),  WarningSink::Renderer);
    assert_eq!(warning_sink(true,  false), WarningSink::Renderer);
    assert_eq!(warning_sink(false, true),  WarningSink::Stderr);
    assert_eq!(warning_sink(false, false), WarningSink::Stdout);

Ran your exact falsifier: forcing `warning_sink` to return `Stdout` unconditionally. Before, this test passed while the other two failed. Now all three fail, this one included.

3f8970d.

src/utils/output.rs
jetm: Fixed in 3f8970d, and folded into the fix for the duplication below it, as you suggested - 4 did largely disappear into 5.

`print_notice_above` binds the renderer `Option` once and matches it alongside the sink:

    (WarningSink::Renderer, Some(active)) => active.print_above(formatted),
    (WarningSink::Renderer, None) | (WarningSink::Stderr, _) => eprintln!("{formatted}"),
    (WarningSink::Stdout, _) => plain(formatted),

So the `Renderer`-sink-with-no-renderer case falls through to stderr instead of aborting. No `expect` left, and the branch is unrepresentable rather than merely unlikely - which is what you asked for, and it means the fix does not depend on settling whether the interleaving is reachable. For the record I think it is: `clear_active_renderer` is `pub` and is the first statement of `TaskRenderer::shutdown`.

src/utils/output.rs
jetm: Fixed in 3f8970d. `print_notice_above` in `output.rs` is now the single implementation, and `container::print_failure_notice` delegates to it - it dropped from nine lines to one.

One thing I did differently from "parameterised on the prefix". The prefix is not the only difference: the plain, neither-renderer-nor-json case sends a warning to **stdout** via `println!` and an error to **stderr** via `print_error`. Parameterising on the prefix alone would have silently moved 11 container call sites from stderr to stdout. So the callback is the plain-case emit rather than the prefix, and everything above that case - which is the part that was actually duplicated - is shared. `warning_sink` stays separate as the pure policy statement, which is also what keeps finding 3's test possible.

Left alone deliberately: the third copy at `commands/runtime/build.rs:817`. It is outside this PR's diff and not in any thread here, so folding it in would widen the change past what was reviewed. Worth a follow-up.

src/utils/output.rs
jetm: Fixed in 3f8970d. `print_warning_above` now emits an event when JSON output is active, alongside the human line on stderr:

    {"event": "warning", "message": "<the notice>"}

On "check what the desktop app already handles rather than invent one" - I checked, and the honest answer is that there was nothing to match exactly. The CLI's existing event vocabulary is `info`, `error` and `output`; there is no warning event today. On the desktop side the only warn handling I found is `src-tauri/src/updates.rs:434`, `Some("warn") | Some("warning") => LogLevel::Warn`, and that is a `level` field in a different parser, not the NDJSON `event` field.

So I matched the `{"event", "message"}` shape the CLI already uses and picked the string that side already accepts. What I could not verify is that avocado-desktop renders it - if its stream consumer matches a closed set of event names, this is dropped silently. That is still strictly better than the current zero events, but I am not claiming it shows up in the UI. Worth confirming from the desktop side before we rely on it.

src/utils/remote.rs
jetm: Agreed - follow-up, not this PR.

The reasoning that decided it: `parse_reported_version` and `is_version_shaped` are pre-existing and this diff adds and removes zero lines mentioning either, so pulling them in makes the PR harder to justify as a unit and mixes a behaviour change into a branch that is otherwise about how an existing outcome is reported. The session working the meta-avocado tree independently reached the same conclusion when I raised it.

That said, the finding stands and it is not cosmetic - `"Kernel 6.6.1\n0.20.0"` yielding `Verified("6.6.1")` against a remote that is actually `0.20.0` means the arm this PR just gave a warning to may rarely run. Gating the anchored field on `is_version_shaped` and continuing to scan is the one-line fix, and it wants its own tests over the three blobs in the original comment.

Resolving because the ask was a decision and it has been made, not because code changed here. Filing it as the follow-up.

mobileoverlord (approved):
Approving. Green CI, and each behavior change has a test that fails if you revert it — that's the bar.

The three-valued gate is the right call and the `Some(false)` vs `None` split is the load-bearing part. I pushed one cleanup commit rather than leaving it as comments:

- `Skipped.reported` was never read by anyone — `version_check_notice` destructured it away, nothing else touched it, and it escaped `dead_code` only because the enum is `pub` in the lib target. The unparseable string is already in the warning text, so it was redundant rather than merely unused. Now `Skipped(Option<String>)`, which also retires the `warning.is_empty()` sentinel for silence.
- `runtime/build.rs:818` still open-coded the routing that `print_notice_above`'s own doc comment claimed it had deduped. Routed it through, comment corrected to three sites.

Net −16 lines, no change to the gate logic, so your regression tests still pin what they pinned. fmt/clippy/full suite green locally.

On size: `VersionGate` earns its keep — those arms sit behind an ssh boundary nothing fakes. `VersionNotice` is a second enum mapping the first onto a three-line match arm, and `warning_sink`'s tests assert a four-line match against a transcription of itself. Not worth churning a green PR over, just where the next one can be smaller.

PR #188, opened 2026-08-03
ADD cve report support

hiagofranco:
Add the new command to generate the cve report. This PR depends on https://github.com/avocado-linux/meta-avocado/pull/252.

### DIFF
--- a/src/commands/cve/report.rs
+++ b/src/commands/cve/report.rs
@@ -1,0 +1,1789 @@
+//! `avocado cve report` — correlate installed packages with known CVEs.
+//!
+//! Reads the JSON published by the Yocto side (`bitbake avocado-cve-report`),
+//! which maps every package a `bitbake world` produced to its recipe and that
+//! recipe's unpatched CVEs. Then it asks RPM what is actually installed in each
+//! sysroot of the project and joins the two on the package name.
+
+use anyhow::{Context, Result};
+use serde::Deserialize;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::sync::Arc;
+
+use crate::utils::config::{ComposedConfig, Config};
+use crate::utils::container::{RunConfig, SdkContainer};
+use crate::utils::lockfile::{SysrootType, RPM_QUERY_ALL_FORMAT};
+use crate::utils::output::{print_info, print_success, print_warning, OutputLevel};
+use crate::utils::output_format::{emit_json_object, JsonOutputGuard, OutputFormat};
+use crate::utils::target::resolve_target_required;
+
+/// Schema version of the document this command emits. Independent of the
+/// version of the source document produced by bitbake.
+const REPORT_VERSION: &str = "1";
+
+/// Marker separating one sysroot's package list from the next in the container
+/// output. Tab-separated so it cannot collide with an RPM package name.
+const SCOPE_MARKER: &str = "##SCOPE\t";
+
+/// Marker emitted when a scope's `rpm -qa` exited non-zero. The query cannot
+/// simply be allowed to fail: the container entrypoint runs under `set -e`, so
+/// an unguarded failure would abort the whole script and lose every later
+/// scope. Recording the failure keeps the script going and keeps an unreadable
+/// sysroot from looking like an empty one.
+const FAILED_MARKER: &str = "##FAILED\t";
+
+/// Walks every sysroot that can hold installed target packages and dumps its
+/// RPM database. `__SDK_QUERY__` is replaced with the host-SDK query, which is
+/// the one case that has no `--root` (its database is found through custom
+/// RPM_CONFIGDIR macros instead).
+///
+/// Roots are de-duplicated by realpath: `$AVOCADO_PREFIX/extensions` is a compat
+/// symlink to the runtime-scoped tree, so it would otherwise be counted twice.
+const DISCOVER_SCRIPT: &str = r#"
+set -u
+SEEN=""
+
+query_root() {
+    scope="$1"
+    root="$2"
+    [ -d "$root" ] || return 0
+    real=$(readlink -f "$root")
+    case " $SEEN " in
+        *" $real "*) return 0 ;;
+    esac
+    SEEN="$SEEN $real"
+    printf '##SCOPE\t%s\t%s\n' "$scope" "$real"
+    # rpm's exit status is reported rather than discarded: unlike `rpm -q`,
+    # which returns non-zero merely because a named package is absent, a
+    # failing `rpm -qa` is always a real error. Guarded so the entrypoint's
+    # `set -e` does not abort the remaining scopes.
+    if ! (unset RPM_ETCCONFIGDIR RPM_CONFIGDIR; rpm -qa --root="$root" __QUERY_FORMAT__); then
+        printf '##FAILED\t%s\n' "$scope"
+    fi
+}
+
+sdk_query() {
+__SDK_QUERY__
+}
+
+# Guarded like every other scope. Without the check an uninstalled project
+# reports the SDK as a failed scan rather than as nothing to scan, which buries
+# the actionable "run avocado install" message: the SDK database is reached
+# through RPM_CONFIGDIR macros, so its absence shows up as an rpm error rather
+# than a missing --root.
+if [ -d "$AVOCADO_SDK_PREFIX/usr/lib/rpm" ]; then
+    printf '##SCOPE\t%s\t%s\n' "sdk" "$AVOCADO_SDK_PREFIX"
+    if ! sdk_query; then
+        printf '##FAILED\t%s\n' "sdk"
+    fi
+fi
+
+query_root "rootfs" "$AVOCADO_PREFIX/rootfs"
+query_root "initramfs" "$AVOCADO_PREFIX/initramfs"
+query_root "target-sysroot" "$AVOCADO_PREFIX/sdk/target-sysroot"
+query_root "includes" "$AVOCADO_PREFIX/includes"
+# A remote extension advertising `avocado-ext-layout(nested)` installs into the
+# shared includes root queried above; one without it gets an installroot, and so
+# an RPM database, of its own (utils/ext_fetch.rs). Both layouts land their
+# content in includes/<name>/, so the per-extension roots have to be queried too
+# or a project whose extensions are all legacy-layout scans nothing here. Nested
+# ones hold no database of their own and drop out as an empty scope.
+for inc_dir in "$AVOCADO_PREFIX"/includes/*/; do
+    [ -d "$inc_dir" ] || continue
+    query_root "includes:$(basename "$inc_dir")" "$inc_dir"
+done
+
+for runtime_dir in "$AVOCADO_PREFIX"/runtimes/*/; do
+    [ -d "$runtime_dir" ] || continue
+    runtime_name=$(basename "$runtime_dir")
+    query_root "runtime:$runtime_name" "$runtime_dir"
+    for ext_dir in "$runtime_dir"extensions/*/; do
+        [ -d "$ext_dir" ] || continue
+        query_root "ext:$runtime_name/$(basename "$ext_dir")" "$ext_dir"
+    done
+done
+
+for ext_dir in "$AVOCADO_PREFIX"/extensions/*/; do
+    [ -d "$ext_dir" ] || continue
+    query_root "ext:$(basename "$ext_dir")" "$ext_dir"
+done
+"#;
+
+// ---------------------------------------------------------------------------
+// The document produced by `bitbake avocado-cve-report`
+// ---------------------------------------------------------------------------
+
+#[derive(Debug, Deserialize)]
+struct SourceReport {
+    #[serde(default)]
+    generated: String,
+    #[serde(default)]
+    machine: Option<String>,
+    // Deliberately not `#[serde(default)]`, for the same reason as the two maps
+    // below: an absent key would default to "" and slip past the Unpatched
+    // check, so a document produced with AVOCADO_CVE_REPORT_STATUS=Patched but
+    // with the key trimmed would be reported as a list of live findings.
+    status: String,
+    // Deliberately not #[serde(default)]: a document missing either map would
+    // otherwise deserialize into empty ones and correlate to zero CVEs, which
+    // reads exactly like a clean scan. An explicitly empty map still parses.
+    recipes: HashMap<String, SourceRecipe>,
+    packages: HashMap<String, SourcePackage>,
+}
+
+#[derive(Debug, Deserialize)]
+struct SourceRecipe {
+    #[serde(default)]
+    cves: Vec<SourceCve>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+struct SourceCve {
+    id: String,
+    #[serde(default)]
+    summary: Option<String>,
+    #[serde(default)]
+    scorev2: Option<String>,
+    #[serde(default)]
+    scorev3: Option<String>,
+    #[serde(default)]
+    vector: Option<String>,
+    #[serde(default)]
+    link: Option<String>,
+}
+
+impl SourceCve {
+    /// Best available score, preferring CVSS v3 over v2. cve-check writes "0.0"
+    /// when a score is absent, so that reads as "unscored" rather than "harmless".
+    fn score(&self) -> Option<f64> {
+        self.resolved_score().map(|(value, _)| value)
+    }
+
+    /// Which CVSS version the resolved score came from, so the JSON document
+    /// says what it ranked by instead of leaving a consumer to guess.
+    fn score_source(&self) -> Option<&'static str> {
+        self.resolved_score().map(|(_, source)| source)
+    }
+
+    fn resolved_score(&self) -> Option<(f64, &'static str)> {
+        for (raw, source) in [(&self.scorev3, "v3"), (&self.scorev2, "v2")] {
+            if let Some(value) = raw.as_ref().and_then(|s| s.parse::<f64>().ok()) {
+                // `is_finite` as well as `> 0.0`: "inf" and "1e400" parse, pass
+                // a bare positive test, and then serialize to JSON `null`
+                // (serde_json cannot represent them) while `score_source` still
+                // says "v3" — and trip every --fail-on-score threshold at once.
+                // "NaN" already falls through, since no comparison holds.
+                if value > 0.0 && value.is_finite() {
+                    return Some((value, source));
+                }
+            }
+        }
+        None
+    }
+}
+
+#[derive(Debug, Deserialize)]
+struct SourcePackage {
+    recipe: String,
+    #[serde(default)]
+    version: String,
+}
+
+// ---------------------------------------------------------------------------
+// What RPM reports as installed
+// ---------------------------------------------------------------------------
+
+#[derive(Debug)]
+struct InstalledPackage {
+    name: String,
+    version: String,
+    arch: String,
+}
+
+/// Every sysroot the scan found, plus what the container wrote to stderr.
+///
+/// The stderr is carried rather than dropped because the `##FAILED` guard makes
+/// the script exit 0 by design: rpm's own diagnosis of an unreadable database
+/// is on stderr and nowhere else, so a bail naming the scope would otherwise be
+/// all the user ever gets.
+#[derive(Debug)]
+struct Sysroots {
+    scopes: Vec<Sysroot>,
+    stderr: String,
+}
+
+#[derive(Debug)]
+struct Sysroot {
+    scope: String,
+    root: String,
+    packages: Vec<InstalledPackage>,
+    /// Its `rpm -qa` exited non-zero. Kept rather than dropped: a sysroot that
+    /// could not be read must never be indistinguishable from one that holds
+    /// no packages.
+    failed: bool,
+}
+
+/// One installed package that a recipe with CVEs produced.
+#[derive(Debug)]
+struct Affected {
+    package: String,
+    /// VERSION-RELEASE as RPM reports it.
+    installed_version: String,
+    arch: String,
+    /// PKGV-PKGR verbatim from the report, i.e. before the `-` -> `+` rewrite
+    /// `package_rpm.bbclass` applies on the way into an RPM. Diffing this
+    /// against `installed_version` raw therefore reports a mismatch for every
+    /// hyphenated PKGV; run `rpm_pkgv` over the PKGV half first, which is what
+    /// the scope-level version_mismatch list already reports the verdict of.
+    report_version: String,
+    recipe: String,
+    cves: Vec<String>,
+}
+
+/// A CVE with every place it was found, so the same CVE shared by several
+/// scopes is counted once.
+#[derive(Debug)]
+struct CveHit {
+    cve: SourceCve,
+    recipes: BTreeSet<String>,
+    packages: BTreeSet<String>,
+    scopes: BTreeSet<String>,
+}
+
+#[derive(Debug, Default)]
+struct ScopeResult {
+    root: String,
+    scanned: usize,
+    /// Packages an extension sysroot only sees because its installroot was
+    /// seeded with the rootfs RPM database. Counted, never attributed here.
+    inherited: usize,
+    affected: Vec<Affected>,
+    unknown: Vec<String>,
+    /// Held by an extension or runtime at a version the rootfs holds
+    /// differently, so whether the scope really ships it cannot be told from
+    /// here. Counted, and attributed to the scope anyway.
+    baseline_divergent: Vec<String>,
+    /// Installed at a different upstream version than the report records.
+    version_mismatch: Vec<String>,
+    /// Same upstream version, different packaging revision (PKGR).
+    revision_mismatch: Vec<String>,
+    /// Matched by name, but one side carried no version, so the cross-check
+    /// could not run. Not the same as having passed it.
+    version_unchecked: Vec<String>,
+    /// The package names a recipe the report has no entry for. The producer
+    /// only writes recipes that carry CVEs, so this is normally the clean
+    /// case — but a truncated document looks identical, hence the counter.
+    recipe_missing: Vec<String>,
+}
+
+impl ScopeResult {
+    fn own_packages(&self) -> usize {
+        self.scanned.saturating_sub(self.inherited)
+    }
+}
+
+/// Split a Yocto package version into (upstream version, packaging revision).
+///
+/// `2.39.3-r0.5` -> ("2.39.3", "r0.5"). The revision is what a rebuild bumps,
+/// so telling the two apart separates "different sources" from "same sources,
+/// repackaged".
+fn split_version(version: &str) -> (&str, &str) {
+    match version.rsplit_once('-') {
+        Some((pkgv, pkgr)) => (pkgv, pkgr),
+        None => (version, ""),
+    }
+}
+
+/// Rewrite a PKGV the way RPM packaging does before comparing it with what
+/// `rpm -qa` reports.
+///
+/// RPM forbids "-" in VERSION, so oe-core's `package_rpm.bbclass` writes
+/// `PKGV.replace('-', '+')` into the package. The report records PKGV verbatim,
+/// so `4-7.1` there is `4+7.1` in the RPM database. Comparing them raw reports
+/// a version mismatch for every such package — 146 of them in a scarthgap
+/// `bitbake world` (blktool, cargo-c, cni, ...).
+fn rpm_pkgv(pkgv: &str) -> String {
+    pkgv.replace('-', "+")
+}
+
+pub struct CveReportCommand {
+    config_path: String,
+    /// The global `--runs-on`, carried only so the command can refuse it.
+    /// `run_in_container_capture` has no remote branch — unlike
+    /// `run_in_container`, which routes on `config.runs_on` — so honouring it
+    /// would take work in `utils::container`, not a field here.
+    runs_on: Option<String>,
+    file: String,
+    target: Option<String>,
+    verbose: bool,
+    container_args: Option<Vec<String>>,
+    output: OutputFormat,
+    /// Exit non-zero when any CVE scores at or above this. Off by default: a
+    /// report is worth reading whatever it finds, and a threshold belongs to
+    /// the policy consuming it, not to the scan.
+    fail_on_score: Option<f64>,
+    sdk_arch: Option<String>,
+    composed_config: Option<Arc<ComposedConfig>>,
+}
+
+impl CveReportCommand {
+    pub fn new(
+        config_path: String,
+        file: String,
+        target: Option<String>,
+        verbose: bool,
+        container_args: Option<Vec<String>>,
+        output: OutputFormat,
+        fail_on_score: Option<f64>,
+    ) -> Self {
+        Self {
+            config_path,
+            file,
+            target,
+            verbose,
+            container_args,
+            output,
+            fail_on_score,
+            runs_on: None,
+            sdk_arch: None,
+            composed_config: None,
+        }
+    }
+
+    /// Record the global `--runs-on` so `execute` can refuse it rather than
+    /// silently scanning the local volume.
+    pub fn with_runs_on(mut self, runs_on: Option<String>) -> Self {
+        self.runs_on = runs_on;
+        self
+    }
+
+    /// Set SDK container architecture for cross-arch emulation
+    pub fn with_sdk_arch(mut self, sdk_arch: Option<String>) -> Self {
+        self.sdk_arch = sdk_arch;
+        self
+    }
+
+    /// Set pre-composed configuration to avoid reloading
+    #[allow(dead_code)]
+    pub fn with_composed_config(mut self, config: Arc<ComposedConfig>) -> Self {
+        self.composed_config = Some(config);
+        self
+    }
+
+    pub async fn execute(&self) -> Result<()> {
+        // Refused rather than ignored. `--runs-on` is a global flag, and the
+        // container helper this command uses reads the local volume with no
+        // remote branch, so honouring it would take work in utils::container.
+        // Until then, accepting it would emit a confident report describing
+        // the wrong host — and a security artifact attributed to a machine it
+        // does not describe is worse than an error.
+        if let Some(host) = &self.runs_on {
+            anyhow::bail!(
+                "--runs-on {host} is not supported by `cve report`: it would read this machine's \
+                 sysroots and attribute the result to {host}. Run the command on that host."
+            );
+        }
+
+        // Silences the print_* helpers so --output json puts nothing but the
+        // document on stdout; without it a warning or --verbose note lands
+        // ahead of the JSON and breaks any consumer parsing it.
+        let _json_guard = self.output.is_json().then(JsonOutputGuard::enable);
+
+        let source = self.load_source()?;
+
+        let composed = match &self.composed_config {
+            Some(cc) => Arc::clone(cc),
+            None => Arc::new(
+                Config::load_composed(&self.config_path, self.target.as_deref())
+                    .context("Failed to load composed config")?,
+            ),
+        };
+        let config = &composed.config;
+        let target = resolve_target_required(self.target.as_deref(), config)?;
+        self.warn_on_machine_mismatch(&source, &target);
+
+        let Sysroots {
+            scopes: sysroots,
+            stderr,
+        } = self.query_sysroots(config, &target).await?;
+
+        // A scope whose RPM database could not be read contributes no packages
+        // and therefore no CVEs. Continuing would report the remaining scopes
+        // as if they were the whole picture. Checked before the empty case
+        // below so a failed query is named as one rather than reported as
+        // nothing to scan.
+        let failed: Vec<&str> = sysroots
+            .iter()
+            .filter(|s| s.failed)
+            .map(|s| s.scope.as_str())
+            .collect();
+        if !failed.is_empty() {
+            anyhow::bail!(
+                "Could not read the RPM database of {} scope(s): {}. Those sysroots were not \
+                 scanned, so any CVE they carry would be missing from this report.{}",
+                failed.len(),
+                failed.join(", "),
+                stderr_tail(&stderr)
+            );
+        }
+
+        // Reporting zero CVEs here would be indistinguishable from a clean
+        // scan. Tested on packages rather than on the scope list: `rpm -qa`
+        // over a wiped database exits 0 with no output, so a project whose
+        // sysroots all exist but hold nothing produces scopes that are present
+        // and empty — the one shape that reached the report as a clean exit 0.
+        if sysroots.iter().all(|s| s.packages.is_empty()) {
+            anyhow::bail!(
+                "No installed package was found in any sysroot for target '{target}', so nothing \
+                 was scanned. Run `avocado install` first — and if the project is installed, its \
+                 RPM databases are unreadable or empty. Note that this command reads the state \
+                 volume of the current directory, not of --config's directory."
+            );
+        }
+
+        let (results, cves) = correlate(&source, &sysroots);
+
+        if self.output.is_json() {
+            emit_json_object(&self.build_json(&source, &target, &results, &cves));
+        } else {
+            self.print_human(&source, &results, &cves);
+        }
+
+        // After the report is emitted either way: a gate that suppresses the
+        // findings it gates on would be useless.
+        if let Some(threshold) = self.fail_on_score {
+            let over = cves_at_or_above(&cves, threshold);
+            if !over.is_empty() {
+                anyhow::bail!(
+                    "{} CVE(s) score {threshold:.1} or higher: {}.",
+                    over.len(),
+                    over.join(", ")
+                );
+            }
+        }
+
+        Ok(())
+    }
+
+    fn load_source(&self) -> Result<SourceReport> {
+        let raw = std::fs::read_to_string(&self.file)
+            .with_context(|| format!("Failed to read CVE report '{}'", self.file))?;
+        let source: SourceReport = serde_json::from_str(&raw)
+            .with_context(|| format!("Failed to parse CVE report '{}'", self.file))?;
+
+        if source.packages.is_empty() {
+            anyhow::bail!(
+                "CVE report '{}' has no 'packages' map; is it an avocado-cve-report document?",
+                self.file
+            );
+        }
+        // An empty map correlates to zero CVEs and reads exactly like a clean
+        // image, and it is what the producer emits when cve-check was not
+        // inherited or its NVD fetch failed. Refuse it as an input rather than
+        // pass its emptiness through as a result.
+        if source.recipes.is_empty() {
+            anyhow::bail!(
+                "CVE report '{}' has an empty 'recipes' map, so nothing could be correlated. \
+                 The producing build most likely ran without cve-check inherited.",
+                self.file
+            );
+        }
+        // The producer records which cve-check status it filtered on. Anything
+        // but Unpatched describes issues that are not live findings, and
+        // correlating it would present them as if they were.
+        if source.status != "Unpatched" {
+            anyhow::bail!(
+                "CVE report '{}' was generated with status '{}', not 'Unpatched'. Those are not \
+                 live findings; regenerate the report with AVOCADO_CVE_REPORT_STATUS unset.",
+                self.file,
+                source.status
+            );
+        }
+        Ok(source)
+    }
+
+    /// The report describes one Yocto MACHINE. Correlating it against a
+    /// different target matches on package name alone and produces a confident
+    /// wrong answer, so say so — but only warn: the MACHINE names a build
+    /// configuration, and more than one can legitimately serve a target.
+    fn warn_on_machine_mismatch(&self, source: &SourceReport, target: &str) {
+        if machine_matches_target(source.machine.as_deref(), target) {
+            return;
+        }
+        let Some(machine) = source.machine.as_deref() else {
+            return;
+        };
+        print_warning(
+            &format!(
+                "Report was generated for MACHINE '{machine}' but this project targets \
+                 '{target}'. Packages are correlated by name only, so the result may not \
+                 describe this image."
+            ),
+            OutputLevel::Normal,
+        );
+    }
+
+    /// Run one container that dumps every sysroot's RPM database.
+    ///
+    /// A single invocation rather than one per sysroot: container startup
+    /// dominates the cost of an `rpm -qa`.
+    async fn query_sysroots(&self, config: &Config, target: &str) -> Result<Sysroots> {
+        let container_image = config.get_sdk_image().cloned().ok_or_else(|| {
+            anyhow::anyhow!("No container image specified in config under 'sdk.image'.")
+        })?;
+
+        let sdk_query = SysrootType::Sdk(target.to_string())
+            .get_rpm_query_config()
+            .build_query_all_command();
+        let command = DISCOVER_SCRIPT
+            .replace("__SDK_QUERY__", &sdk_query)
+            .replace("__QUERY_FORMAT__", RPM_QUERY_ALL_FORMAT);
+
+        if self.verbose {
+            print_info(
+                "Querying installed packages in every sysroot.",
+                OutputLevel::Normal,
+            );
+        }
+
+        let container = SdkContainer::from_config(&self.config_path, config)?.verbose(self.verbose);
+        let run_config = RunConfig {
+            container_image,
+            target: target.to_string(),
+            command,
+            verbose: self.verbose,
+            // The RPM queries need only the entrypoint's base env vars
+            // ($AVOCADO_PREFIX, $AVOCADO_SDK_PREFIX), not the full SDK env.
+            source_environment: false,
+            use_entrypoint: true,
+            interactive: false,
+            repo_url: config.get_sdk_repo_url(),
+            repo_release: config.get_sdk_repo_release(),
+            container_args: config.merge_sdk_container_args(self.container_args.as_ref()),
+            sdk_arch: self.sdk_arch.clone(),
+            ..Default::default()
+        };
+
+        // `run_in_container_capture` rather than `run_in_container_with_output`,
+        // which the sibling commands use. Two things follow from the difference,
+        // and both matter here:
+        //
+        // - `_with_output` returns `Ok(None)` on a non-zero exit and discards
+        //   the captured stdout, so docker being down or the entrypoint
+        //   aborting would arrive as an empty scan — indistinguishable from an
+        //   uninstalled project, and answered with "run `avocado install`".
+        // - it also drops stderr on success, and the `##FAILED` guard makes the
+        //   script exit 0 by design. So rpm's own "cannot open Packages
+        //   database" had nowhere to go, and the bail below telling the user to
+        //   re-run with --verbose promised something --verbose could not do.
+        let out = container.run_in_container_capture(run_config).await?;
+        if !out.success {
+            anyhow::bail!(
+                "Could not query the sysroots of target '{target}': the SDK container exited \
+                 non-zero. This is a container or SDK failure, not an empty project.{}",
+                stderr_tail(&out.stderr)
+            );
+        }
+
+        Ok(Sysroots {
+            scopes: parse_sysroots(&out.stdout),
+            stderr: out.stderr,
+        })
+    }
+
+    fn build_json(
+        &self,
+        source: &SourceReport,
+        target: &str,
+        results: &BTreeMap<String, ScopeResult>,
+        cves: &BTreeMap<String, CveHit>,
+    ) -> serde_json::Value {
+        let scopes: serde_json::Map<String, serde_json::Value> = results
+            .iter()
+            .map(|(scope, result)| {
+                let affected: Vec<serde_json::Value> = result
+                    .affected
+                    .iter()
+                    .map(|a| {
+                        let (pkgv, pkgr) = split_version(&a.report_version);
+                        serde_json::json!({
+                            "package": a.package,
+                            "installed_version": a.installed_version,
+                            "arch": a.arch,
+                            "report_version": a.report_version,
+                            // report_version put through the same PKGV rewrite
+                            // this command compares with, so a consumer can
+                            // diff it against installed_version directly. Left
+                            // alongside the raw field rather than replacing it:
+                            // report_version is the key that looks the package
+                            // back up in the source document, and normalizing
+                            // it would break that lookup.
+                            "report_version_rpm": if pkgr.is_empty() {
+                                rpm_pkgv(pkgv)
+                            } else {
+                                format!("{}-{pkgr}", rpm_pkgv(pkgv))
+                            },
+                            "recipe": a.recipe,
+                            "cves": a.cves,
+                        })
+                    })
+                    .collect();
+                (
+                    scope.clone(),
+                    serde_json::json!({
+                        "root": result.root,
+                        "packages_scanned": result.scanned,
+                        "packages_inherited": result.inherited,
+                        "packages_own": result.own_packages(),
+                        "packages_affected": result.affected.len(),
+                        "packages_unknown": result.unknown.len(),
+                        "packages_baseline_divergent": result.baseline_divergent.len(),
+                        "packages_version_mismatch": result.version_mismatch.len(),
+                        "packages_revision_mismatch": result.revision_mismatch.len(),
+                        "packages_version_unchecked": result.version_unchecked.len(),
+                        "recipes_missing": result.recipe_missing.len(),
+                        "affected": affected,
+                        "unknown": result.unknown,
+                        "baseline_divergent": result.baseline_divergent,
+                        "version_mismatch": result.version_mismatch,
+                        "revision_mismatch": result.revision_mismatch,
+                        "version_unchecked": result.version_unchecked,
+                    }),
+                )
+            })
+            .collect();
+
+        let cve_index: serde_json::Map<String, serde_json::Value> = cves
+            .iter()
+            .map(|(id, hit)| {
+                (
+                    id.clone(),
+                    serde_json::json!({
+                        // The resolved score this command itself ranks by, so a
+                        // consumer gating on severity does not have to
+                        // reimplement the prefer-v3 rule and the "0.0 means
+                        // unscored, not harmless" convention — and disagree
+                        // with the human output when it gets them wrong.
+                        "score": hit.cve.score(),
+                        "score_source": hit.cve.score_source(),
+                        "scorev2": hit.cve.scorev2,
+                        "scorev3": hit.cve.scorev3,
+                        "vector": hit.cve.vector,
+                        "link": hit.cve.link,
+                        "summary": hit.cve.summary,
+                        "recipes": hit.recipes,
+                        "packages": hit.packages,
+                        "scopes": hit.scopes,
+                    }),
+                )
+            })
+            .collect();
+
+        serde_json::json!({
+            "version": REPORT_VERSION,
+            "target": target,
+            "source": {
+                "file": self.file,
+                "machine": source.machine,
+                "status": source.status,
+                "generated": source.generated,
+                // The verdict, not just the two strings it is derived from.
+                // JsonOutputGuard silences the human warning under --output
+                // json, so without this a consumer either re-implements the
+                // suffix rule — the one already gotten wrong once — or gates
+                // CI on a correlation nothing vouched for.
+                "machine_mismatch": !machine_matches_target(source.machine.as_deref(), target),
+            },
+            "counts": {
+                "scopes": results.len(),
+                "packages_scanned": results.values().map(|r| r.scanned).sum::<usize>(),
+                "packages_inherited": results.values().map(|r| r.inherited).sum::<usize>(),
+                "packages_own": results.values().map(|r| r.own_packages()).sum::<usize>(),
+                "packages_affected": results.values().map(|r| r.affected.len()).sum::<usize>(),
+                "packages_unknown": results.values().map(|r| r.unknown.len()).sum::<usize>(),
+                "packages_baseline_divergent":
+                    results.values().map(|r| r.baseline_divergent.len()).sum::<usize>(),
+                "packages_version_mismatch":
+                    results.values().map(|r| r.version_mismatch.len()).sum::<usize>(),
+                "packages_revision_mismatch":
+                    results.values().map(|r| r.revision_mismatch.len()).sum::<usize>(),
+                "packages_version_unchecked":
+                    results.values().map(|r| r.version_unchecked.len()).sum::<usize>(),
+                "recipes_missing":
+                    results.values().map(|r| r.recipe_missing.len()).sum::<usize>(),
+                "cves": cves.len(),
+                // Carried explicitly: these are the CVEs no --fail-on-score
+                // threshold can match, so a consumer gating on severity has to
+                // be able to see how much the gate could not speak for.
+                "cves_unscored": cves.values().filter(|h| h.cve.score().is_none()).count(),
+            },
+            "scopes": scopes,
+            "cves": cve_index,
+        })
+    }
+
+    fn print_human(
+        &self,
+        source: &SourceReport,
+        results: &BTreeMap<String, ScopeResult>,
+        cves: &BTreeMap<String, CveHit>,
+    ) {
+        let any_inherited = results.values().any(|r| r.inherited > 0);
+
+        println!(
+            "{:<34} {:>9} {:>9} {:>6}",
+            "SCOPE", "PACKAGES", "AFFECTED", "CVES"
+        );
+        for (scope, result) in results {
+            let scope_cves: BTreeSet<&String> =
+                result.affected.iter().flat_map(|a| a.cves.iter()).collect();
+            println!(
+                "{:<34} {:>9} {:>9} {:>6}",
+                scope,
+                result.own_packages(),
+                result.affected.len(),
+                scope_cves.len()
+            );
+        }
+        if any_inherited {
+            println!(
+                "\nExtension and runtime counts exclude packages inherited from the rootfs \
+                 database; those are attributed to the rootfs scope alone."
+            );
+        }
+        // The subtraction above needs a rootfs scope with packages in it to
+        // subtract. Without one — a project with no rootfs installed, or one
+        // whose RPM database was wiped, which `rpm -qa` reports as an empty
+        // scan rather than as an error — every seeded scope keeps the whole
+        // base system as its own. Silence there would read as "these scopes
+        // really do ship all this". A present-but-empty rootfs counts as
+        // absent here: it is the wiped-database case, and it is the one that
+        // used to be dropped before this warning could see it.
+        let seeded_without_baseline = results.get("rootfs").is_none_or(|r| r.scanned == 0)
+            && results
+                .keys()
+                .any(|s| s.starts_with("ext:") || s.starts_with("runtime:"));
+        if seeded_without_baseline {
+            print_warning(
+                "No rootfs scope was scanned, so packages the extensions and runtimes inherited \
+                 from the rootfs database cannot be told apart from what they ship: their counts \
+                 include the whole base system.",
+                OutputLevel::Normal,
+            );
+        }
+
+        // Highest-scoring CVEs first — that ordering is the reason to read this
+        // at all, and the full list belongs in --output json.
+        let mut ranked: Vec<&CveHit> = cves.values().collect();
+        ranked.sort_by(|a, b| {
+            b.cve
+                .score()
+                .partial_cmp(&a.cve.score())
+                .unwrap_or(std::cmp::Ordering::Equal)
+                .then_with(|| a.cve.id.cmp(&b.cve.id))
+        });
+
+        if !ranked.is_empty() {
+            println!();
+            for hit in ranked.iter().take(10) {
+                let score = hit
+                    .cve
+                    .score()
+                    .map(|s| format!("{s:.1}"))
+                    .unwrap_or_else(|| "-".to_string());
+                println!(
+                    "{:<18} {:>5}  {:<10} {}",
+                    hit.cve.id,
+                    score,
+                    hit.cve.vector.as_deref().unwrap_or("-"),
+                    hit.recipes.iter().cloned().collect::<Vec<_>>().join(", ")
+                );
+            }
+            if ranked.len() > 10 {
+                println!("... and {} more (use --output json)", ranked.len() - 10);
+            }
+        }
+
+        // An unscored CVE sorts last and so never reaches the ten rows above,
+        // and --fail-on-score compares against a score it does not have. Both
+        // are the right behaviours — a threshold cannot rank what carries no
+        // rank — but together they make these invisible, which is the one
+        // reading the "0.0 is unscored, not harmless" rule exists to prevent.
+        let unscored = cves.values().filter(|h| h.cve.score().is_none()).count();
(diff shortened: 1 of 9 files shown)


hiagofranco:
Eveything was wrote and tested with python first, then used Claude Opus 5 to rewrite to Rust and add the necessary tests to it. Tested against the feed produced by https://github.com/avocado-linux/meta-avocado/pull/252 with the x86 QEMU machine.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+            continue;
+        }
+        current.packages.push(InstalledPackage {
+            name: name.to_string(),
+            version: version.trim().to_string(),
+            arch: fields.next().unwrap_or_default().trim().to_string(),
+        });
+    }
+
+    sysroots.retain(|s| !s.packages.is_empty());
```
jetm: **Empty scopes are dropped, so a failed rpm query vanishes**

`parse_sysroots` ends with `sysroots.retain(|s| !s.packages.is_empty())`, which erases any sysroot whose `rpm -qa` produced no lines. A scope that failed to be read becomes indistinguishable from one that was never installed, and invisible in both the human table and the JSON `scopes` map.

A project has `rootfs`, `runtime:dev`, and the SDK installed. The rootfs RPM database is in a format the SDK's `rpm` cannot open (bdb-vs-sqlite backend mismatch), so `rpm -qa --root=$AVOCADO_PREFIX/rootfs` writes an error to stderr and nothing to stdout. The `##SCOPE\trootfs\t...` marker was already printed, so a `Sysroot` with zero packages is built and then silently discarded here. `sysroots` is still non-empty, so the `bail!` at line 279 does not fire. Output: `[SUCCESS] 0 unpatched CVE(s) across 2 scope(s).`, exit 0, and JSON `counts.scopes` says 2.

The entire shipped root filesystem was never scanned and nothing in the output says so. This is the precise false-clean the module's own doc comments claim to defend against. Keeping the empty scope and reporting it as unscanned costs one field.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+    scope="$1"
+    root="$2"
+    [ -d "$root" ] || return 0
+    real=$(readlink -f "$root")
+    case " $SEEN " in
+        *" $real "*) return 0 ;;
+    esac
+    SEEN="$SEEN $real"
+    printf '##SCOPE\t%s\t%s\n' "$scope" "$real"
+    (unset RPM_ETCCONFIGDIR RPM_CONFIGDIR; rpm -qa --root="$root" __QUERY_FORMAT__) || true
```
jetm: **`|| true` hides every rpm failure from the caller**

`query_root` swallows rpm's exit status, and `DISCOVER_SCRIPT` sets only `set -u` (not `set -e`), so the script exits 0 even when every rpm invocation failed. `run_in_container_with_output` (`src/utils/container.rs:1446-1467`) returns `Ok(None)` and prints a failure notice *only* when `out.success` is false, so it sees a successful run and reports nothing.

The container image's `rpm` is missing, or the state volume is mounted read-only. Every `rpm -qa` exits non-zero, `|| true` masks each one, the final `for` loop exits 0, the container exits 0. `out.success` is true, so the helper returns `Ok(Some(""))` with no diagnostic. The command then hits the `bail!` at line 279 and tells the user "Run `avocado install` first" - a confidently wrong diagnosis for a broken container. In the partial case above it does not bail at all.

Worth contrasting with the sibling `build_query_command` at `src/utils/lockfile.rs:190`, where `|| true` is justified by a documented, narrow reason: `rpm -q` returns non-zero when any *named* package is absent. `rpm -qa` has no such excuse - its non-zero exit is always a real error.

jetm (commented):
Reviewed `avocado cve report` cold. Twelve findings inline, most severe first. The command is well shaped and the doc comments are unusually good about naming what they defend against - which is why the top three matter: each is a path where a scan that did not happen reports as a scan that found nothing.

The first two are the same failure seen from both ends. `query_root` swallows rpm's exit status with `|| true` and `DISCOVER_SCRIPT` sets only `set -u`, so a container where every `rpm -qa` failed still exits 0; then `parse_sysroots` drops the resulting empty scopes with `retain`, so the failure leaves no trace at all. A rootfs that could not be read comes out as `[SUCCESS] 0 unpatched CVE(s)` with the scope count silently one lower. Third is the same shape at the document level: an empty `recipes` map is accepted, correlates to zero CVEs, and is indistinguishable from a clean image - and a producer that ran without `cve-check` inherited emits exactly that.

Three smaller notes not worth their own thread:

`SourceReport::status` is `#[serde(default)]`, so a document without it prints `[SUCCESS] 12  CVE(s) across 3 scope(s).` with a doubled space and no qualifier. It is also never validated, so a document generated with `status = "Patched"` correlates and reports identically - a user handed the wrong file sees already-fixed CVEs presented as live findings. Separately, `with_composed_config` (`report.rs:251`) has no caller in the diff and is silenced with `#[allow(dead_code)]` rather than omitted; `main.rs:3196` builds the command without it. And the CHANGELOG entry lists six scopes but `DISCOVER_SCRIPT` also emits `includes` (line 58) and the un-prefixed `ext:<name>` compat form (line 71), so a consumer keying on the documented set drops those.

Some things I checked and am deliberately not filing, so they do not get re-raised: `emit_json_object_pretty` ignoring write errors matches the established `emit_json_event` convention verbatim; the kernel sysroot is correctly omitted from the discovery script, since `rootfs/install.rs:114-190` shows it holds only copied `Image` files with no queryable rpmdb and the kernel package is scanned in the rootfs; `JsonOutputGuard` does correctly suppress `print_*`, because `tui_is_active()` ORs in `is_json_output_active()` and those helpers write to stdout; and `__SDK_QUERY__` is substituted before `__QUERY_FORMAT__` with no token in its expansion, so the two `replace` calls cannot corrupt each other.

Two things about how this was reviewed. **The PR head is not fetched into any local clone** - `refs/pull/188/head` does not resolve - so the three new files were reviewed as-diffed. For `report.rs` and both test files the diff is the complete file (`@@ -0,0 +1,912 @@`), so coverage of the new code is total; the helpers it calls I read from the base repo on disk, which is on `container-dev-mode` (`2c8e854`) rather than necessarily this PR's merge base. **Nothing was built or run.** Findings 1, 2, 3, and 5 are mechanically certain from the code as written; 9 and 10 are marked plausible inline with the reason.

The gap that would most change this review: I did not read the producer (`bitbake avocado-cve-report`, presumably in meta-avocado). Three things hinge on it - whether `recipes` is guaranteed to hold only CVE-bearing recipes, whether `status` can take values other than `"Unpatched"`, and whether a `PKGV` containing a hyphen (say `1.0-rc1`) can reach `split_version` at `report.rs:195`, where `rsplit_once('-')` would mis-split it against RPM's hyphen-free `%{VERSION}` and produce a spurious `version_mismatch` on every such package. Worth confirming that last one before merge; it is cheap to check and would be noisy in the field.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+        Ok(())
+    }
+
+    fn load_source(&self) -> Result<SourceReport> {
+        let raw = std::fs::read_to_string(&self.file)
+            .with_context(|| format!("Failed to read CVE report '{}'", self.file))?;
+        let source: SourceReport = serde_json::from_str(&raw)
+            .with_context(|| format!("Failed to parse CVE report '{}'", self.file))?;
+
+        if source.packages.is_empty() {
```
jetm, line 474: **Empty `recipes` map accepted as a clean scan**

`load_source` rejects an empty `packages` map but applies no equivalent check to `recipes`, and the unit test at line 896 explicitly blesses `{"recipes": {}, "packages": {...}}` as legitimate.

The Yocto side runs `bitbake avocado-cve-report` on a build where `cve-check` was not inherited, or where the NVD database fetch timed out. The producer still enumerates all 1,400 packages into `packages` but writes `"recipes": {}`. `load_source` passes because packages is non-empty. `correlate` looks up every package's recipe, misses every time at line 677, and returns an empty `cves` map. Output: `[SUCCESS] 0 unpatched CVE(s) across 5 scope(s).`, exit 0, `counts.cves: 0`, `counts.packages_unknown: 0` - indistinguishable from a genuinely clean image.

The module rejects a *missing* `recipes` key for exactly this reason (line 92 comment). An *empty* one is the far likelier producer failure mode and gets no guard.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+
+            let Some(source_package) = source.packages.get(&installed.name) else {
+                result.unknown.push(installed.name.clone());
+                continue;
+            };
+
+            // Matched by name. A different upstream version means the report may
+            // not describe this package at all; a different revision means the
+            // same sources were repackaged, which can still change patches.
+            if !source_package.version.is_empty() && !installed.version.is_empty() {
```
jetm: **Missing version fields silence the mismatch check**

The version-drift comparison is gated on `!source_package.version.is_empty() && !installed.version.is_empty()`, and `SourcePackage::version` carries `#[serde(default)]`. A report whose package entries omit `version` correlates purely on package name, with no `version_mismatch` entry, no warning, and no JSON signal.

A producer or a schema revision emits `"packages": {"libssl3": {"recipe": "openssl"}}` with no version. Every installed package matches by name, `version_mismatch` and `revision_mismatch` stay empty, and the human output prints neither warning. The report then attributes openssl's CVE list to the installed `libssl3` regardless of whether the image ships 3.5.7 or 3.0.2, and the user has no indication the version cross-check was skipped rather than passed.

The parallel guard for the whole document - `recipes`/`packages` deliberately lacking `#[serde(default)]` - shows the author already knows silent defaults are dangerous here. This field got one anyway.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+
+    /// The report describes one Yocto MACHINE. Correlating it against a
+    /// different target matches on package name alone and produces a confident
+    /// wrong answer, so say so — but only warn: the MACHINE names a build
+    /// configuration, and more than one can legitimately serve a target.
+    fn warn_on_machine_mismatch(&self, source: &SourceReport, target: &str) {
+        let Some(machine) = source.machine.as_deref() else {
+            return;
+        };
+        if machine.contains(target) {
```
jetm: **MACHINE check uses substring, not equality**

`warn_on_machine_mismatch` returns early when `machine.contains(target)`, so a target name that is a prefix or substring of an unrelated MACHINE suppresses the warning that exists precisely to stop a cross-machine correlation.

The report was generated for `MACHINE = "avocado-qemuarm64"`. The user runs `avocado cve report -f report.json --target qemuarm` against a 32-bit ARM project. `"avocado-qemuarm64".contains("qemuarm")` is true, so the function returns here and prints nothing. The 32-bit and 64-bit images share most package *names* but not their versions or their patch sets, so the command produces a full, confident, wrong CVE list with no caveat.

The same trap fires for `imx8mp` vs `imx8mp-lpddr4`, `raspberrypi4` vs `raspberrypi4-64`, and any `foo`/`foo64` pair - which covers a large share of the machines this project actually builds.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+
+            let Some(recipe) = source.recipes.get(&source_package.recipe) else {
+                continue; // Recipe has no CVE of the reported status.
+            };
+            if recipe.cves.is_empty() {
+                continue;
+            }
+
+            for cve in &recipe.cves {
+                let hit = cves.entry(cve.id.clone()).or_insert_with(|| CveHit {
```
jetm, line 1150: **First recipe seen wins the CVE's score and summary**

`cves.entry(cve.id.clone()).or_insert_with(|| CveHit { cve: cve.clone(), .. })` keeps only the first `SourceCve` payload observed for a given ID and silently discards later ones. When two recipes report the same CVE with different score metadata, the retained copy is whichever the sysroot iteration order reached first.

Both `openssl` and `openssl-native` list `CVE-2026-1234`; the `openssl-native` entry has `"scorev3": "0.0"` (cve-check's unscored placeholder) while the `openssl` entry has `"scorev3": "9.8"`. If the unscored copy is inserted first, `CveHit::cve.score()` returns `None` for the rest of the run. In `print_human` the descending sort at line 507 places `None` last (Rust orders `None < Some`), so a 9.8 critical is pushed below every scored CVE, past the `take(10)` cut at line 517, and appears only inside `... and N more`. The JSON `cves` index likewise reports `"scorev3": "0.0"`.

A user triaging by the human output never sees the critical. Merging on max score rather than first-wins closes it.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+
+        let (results, cves) = correlate(&source, &sysroots);
+
+        if self.output.is_json() {
+            emit_json_object_pretty(&self.build_json(&source, &target, &results, &cves));
+        } else {
+            self.print_human(&source, &results, &cves);
+        }
+
+        Ok(())
```
jetm, line 465: **Exit status is 0 no matter how many CVEs are found**

`execute` returns `Ok(())` unconditionally once correlation succeeds, so the command exits 0 whether it found zero CVEs or five hundred, and there is no threshold flag (`--fail-on`, `--max-score`) to change that.

A CI job adds `avocado cve report -f report.json --output json` as a release gate. The build ships with three CVSS 9.8 CVEs in openssl. The step exits 0, the pipeline goes green, and the image is published.

Gating today requires the job to pipe stdout into `jq` and reimplement severity thresholding - which, per the JSON-score finding below, it cannot do correctly without also reimplementing the score-resolution rule. For a command whose stated purpose is that a missed CVE is worse than a false positive, having no non-zero exit path for *findings* (only for infrastructure failures) is the wrong default.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+            })
+            .collect();
+
+        let cve_index: serde_json::Map<String, serde_json::Value> = cves
+            .iter()
+            .map(|(id, hit)| {
+                (
+                    id.clone(),
+                    serde_json::json!({
+                        "scorev2": hit.cve.scorev2,
```
jetm, line 669: **JSON omits the resolved score used for ranking**

`build_json`'s CVE index emits the raw `scorev2` and `scorev3` strings but never the value `SourceCve::score()` computes from them, so every machine consumer has to independently reimplement the prefer-v3 rule *and* the non-obvious "`0.0` means unscored, not harmless" convention documented at line 118.

A CI gate does `jq '[.cves[] | select((.scorev3 // "0") | tonumber >= 7.0)] | length'`. For a CVE whose entry is `{"scorev3": "0.0", "scorev2": "9.0"}` - cve-check's shape for a CVE with only a v2 score - the filter reads `0.0 >= 7.0`, excludes it, and the gate passes on a critical this tool's own human output would have ranked first.

The consumer's numbers and the CLI's numbers disagree and nothing in the JSON reveals which is authoritative. Emitting `"score": 9.0` alongside (and ideally `"score_source": "v2"`) makes the document self-describing.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+        .unwrap_or_default();
+
+    for sysroot in sysroots {
+        let is_extension = sysroot.scope.starts_with("ext:");
+        let result = results.entry(sysroot.scope.clone()).or_default();
+        result.root = sysroot.root.clone();
+        result.scanned += sysroot.packages.len();
+
+        for installed in &sysroot.packages {
+            if is_extension && baseline.get(installed.name.as_str()) == Some(&&*installed.version) {
```
jetm: **Inherited-package check ignores arch** (PLAUSIBLE - mechanism traced, precondition unconfirmed)

The baseline map built at line 636 is keyed on package *name* only, mapping to a single version, so `baseline.get(name) == Some(&version)` can match an extension's genuinely-own package against an unrelated same-name rootfs entry and mark it inherited. `InstalledPackage::arch` is parsed at line 616 and echoed into JSON but never consulted during matching.

The rootfs contains two entries named `libfoo` at `1.0-r0`, one `noarch` and one `core2_64` (multilib, or a mixed-arch feed); the `HashMap` retains whichever was read last. An extension ships its own `libfoo 1.0-r0` for the other arch. Here the name and version both match, so the extension's package is counted as inherited and `continue`d past - its recipe's CVEs are never attributed to that extension scope, and `scopes["ext:dev/foo"].affected` omits it.

Ranked here rather than higher because the CVE still surfaces under `rootfs`, so this loses scope attribution rather than the finding itself. Marked plausible because I could not confirm from the diff whether same-name-different-arch entries actually occur in an Avocado rootfs.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,912 @@
...
+                let (installed_pkgv, _) = split_version(&installed.version);
+                let (report_pkgv, _) = split_version(&source_package.version);
+                if installed_pkgv != report_pkgv {
+                    result.version_mismatch.push(installed.name.clone());
+                } else if installed.version != source_package.version {
+                    result.revision_mismatch.push(installed.name.clone());
+                }
+            }
+
+            let Some(recipe) = source.recipes.get(&source_package.recipe) else {
```
jetm, line 1138: **Missing recipe entry is counted as clean, and uncounted** (PLAUSIBLE - design intent confirmed, producer contract not)

When a package's `recipe` string has no corresponding key in `source.recipes`, the loop `continue`s with no counter and no list. The package lands in none of `affected`, `unknown`, `version_mismatch`, or `revision_mismatch`, so the report offers no way to tell "this recipe genuinely has no CVEs" from "this recipe's entry is missing".

The producer truncates its output - disk full, or a partial write during `bitbake` shutdown - emitting a complete `packages` map but only the first 200 of 900 `recipes` entries. Every package whose recipe fell off the end takes this `continue`. The `unknown` counter, explicitly designed to surface "unchecked, not known to be clean" (line 561), stays at zero, because those packages *were* found in `packages`. The user sees a warning-free report.

The design intent here is legitimate and confirmed by the fixture at line 744 (`bash` appears in `packages` with no `recipes` entry, and the test at 793 asserts it is neither affected nor unknown), so this is an observability gap rather than an outright bug. A `recipes_missing` counter alongside `unknown` closes it. Marked plausible because I could not read the producer to confirm whether `recipes` is documented as CVE-bearing-only.

tests/commands/avocado/cve/report.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[test]
+fn test_file_is_required() {
+    common::refute_cmd(&["cve", "report"], None, None);
+}
+
+#[test]
+fn test_missing_report_file_fails() {
+    let result = common::run_cli_in_temp_with_config(&["cve", "report", "-f", "no-such.json"]);
+    assert_ne!(result.exit_code, 0, "a missing report must not succeed");
```
jetm, line 72: **Tests assert only a non-zero exit code**

All four negative-path tests (lines 36, 49, 65, 78) assert nothing but `assert_ne!(result.exit_code, 0)`, so each passes for *any* failure reason and cannot detect whether the validation it names actually ran.

Delete `load_source`'s `packages.is_empty()` guard entirely, or reorder `execute` so `Config::load_composed` runs first, and `test_report_without_packages_map_fails` still passes - the command now fails later, at config resolution, at `resolve_target_required`, or at the container step (no Docker in CI), and still exits non-zero.

Workspace `CLAUDE.md`: "Every test must fail when the implementation is wrong (flip the logic and check)" and "Assert exact expected state, not just non-nil / no-error / truthy." The fix is available and has in-repo precedent: `TestResult` exposes `stderr` (`tests/common/mod.rs:11`) and `tests/commands/avocado/fetch/mod.rs:36` already asserts on `format!("{}{}", result.stdout, result.stderr)`. Each test should assert its own error text.

src/utils/lockfile.rs
```diff
@@ -185,7 +185,48 @@ pub struct RpmQueryConfig {
...
+            // that would fight with --root, so clear them first, exactly as
+            // build_query_command does.
+            let mut env_setup = String::from("unset RPM_ETCCONFIGDIR RPM_CONFIGDIR; ");
+            if let Some(ref etcconfigdir) = self.rpm_etcconfigdir {
+                env_setup.push_str(&format!("export RPM_ETCCONFIGDIR=\"{etcconfigdir}\"; "));
+            }
+            if let Some(ref configdir) = self.rpm_configdir {
+                env_setup.push_str(&format!("export RPM_CONFIGDIR=\"{configdir}\"; "));
+            }
+            format!("({env_setup}rpm -qa --root=\"{root_path}\" {qf}) || true")
```
jetm: **Shell values interpolated unescaped, and no unit test pins the format**

`build_query_all_command` splices `root_path`, `rpm_etcconfigdir`, and `rpm_configdir` into a shell string inside plain double quotes with no escaping - and unlike its sibling `build_query_command`, which has three unit tests at lines 1759, 1772, and 2124, this one ships with none.

Injection path: `root_path` for the SDK case derives from `SysrootType::Sdk(target.to_string())` where `target` comes from `--target` or `avocado.yaml`. A target value containing `$(...)` or a `"` reaches the generated command and executes inside the SDK container. Severity is limited since the user already controls what runs in their own container, but `--target` is exactly the sort of field that gets populated from a CI variable.

The untested-contract half matters more. `RPM_QUERY_ALL_FORMAT` at line 194 is deliberately `pub` so this builder and `report.rs`'s parser share one definition, yet nothing asserts the two agree. Add a field to the `--qf`, or swap a tab for a space, and `parse_sysroots`' `split('\t')` at `report.rs:604` breaks silently: every line fails the two-field check, every scope comes back empty, `retain` drops them all, and the command bails with the misleading "run `avocado install`" message.

hiagofranco:
Thanks for the review, I pushed the new version with the corrections. 
Only two notes about the review:

- **Injection via `root_path` (finding 12).** This one is not reachable. The SDK
  case constructs `SysrootType::Sdk(_arch)`, which discards the target string
  and leaves `root_path: None`, so the interpolating branch is never taken for
  it; every value that does reach the format string is a compile-time literal.
  The untested-contract half of the finding was the real one, and is fixed.
- **`with_composed_config` and `#[allow(dead_code)]`.** This is the established
  pattern rather than a local shortcut: 35 commands define this method, 27 of
  them with the identical `#[allow(dead_code)]` and the identical doc comment,
  against 13 call sites. Dropping it only here would make `cve report` the
  odd one out. Happy to remove it if you would rather not extend the pattern.

src/utils/lockfile.rs
jetm: You're right, and I checked rather than taking it on trust. Withdrawing the injection half.

`get_rpm_query_config()` pins `root_path: None` for `Sdk` at `lockfile.rs:101-107` (the `_arch` binding discards the argument), and `report.rs:472` is the only production construction that reaches `build_query_all_command`, at `report.rs:474`. So the `if let Some(ref root_path)` arm at `lockfile.rs:210` is never entered outside tests, and the `else` arm interpolates only `rpm_etcconfigdir` / `rpm_configdir`, both hardcoded at `lockfile.rs:103-104`. Your own `test_sdk_query_config_ignores_target_value` at `lockfile.rs:1809-1818` already pins it - `Sdk("$(id); \"")` and `Sdk("x86_64")` produce byte-identical commands. Agreed the untested-contract half was the real finding, and it's fixed.

One thing my finding got wrong in a way worth naming, since it cuts against the "every value reaching that format string is a compile-time literal" generalization rather than against your scoping. The sibling `build_query_command` at `lockfile.rs:261-266` has a live production `root_path` interpolation via `container.rs:1643`, reached by `Extension`, `Runtime`, `Kernel` and `RuntimeExtension` - all carrying config- and CLI-derived names (`ext/install.rs:107,111`, `runtime/install.rs:375,458`, `rootfs/install.rs:231`). That's outside this PR and outside the CVE report path, so it isn't yours to fix here; I'm noting it because I'd otherwise be leaving you with a conclusion that doesn't hold one function over.

Happy for you to resolve this one - I can't, since I'm reviewer-role on it.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1266 @@
...
+        .map(|s| {
+            s.packages
+                .iter()
+                .map(|p| ((p.name.as_str(), p.arch.as_str()), p.version.as_str()))
+                .collect()
+        })
+        .unwrap_or_default();
+
+    for sysroot in sysroots {
+        let is_extension = sysroot.scope.starts_with("ext:");
```
jetm: **Runtime scopes count the whole rootfs as their own**

The inherited-baseline exclusion is gated on `scope.starts_with("ext:")`, but `runtime:<name>` installroots are seeded from the rootfs RPM database by the identical mechanism - `runtime/install.rs:491` and `runtime/dnf.rs:220` both run `cp -rf $AVOCADO_PREFIX/rootfs/var/lib/rpm $AVOCADO_PREFIX/runtimes/<rt>/var/lib`, the same copy `ext/install.rs:603` does. So every runtime scope attributes the entire base system to itself.

I ran `correlate()` on a rootfs holding `libssl3` + `libz1` and a `runtime:dev` holding those two plus its own `openssh`. It returned `inherited=0, own=3, affected=2`; the correct answer, and what the `ext:` path returns for the identical input, is `inherited=2, own=1, affected=0`. `CVE-2026-1` and `CVE-2026-3` came back with `scopes={"rootfs","runtime:dev"}`, telling the reader those CVEs are present in the runtime when only the inherited base database carries them.

Nothing warns, either: `any_inherited` at line 666 stays false when a project has runtimes but no extensions, so the "counts exclude inherited packages" note never prints.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1266 @@
...
+    case " $SEEN " in
+        *" $real "*) return 0 ;;
+    esac
+    SEEN="$SEEN $real"
+    printf '##SCOPE\t%s\t%s\n' "$scope" "$real"
+    # rpm's exit status is reported rather than discarded: unlike `rpm -q`,
+    # which returns non-zero merely because a named package is absent, a
+    # failing `rpm -qa` is always a real error. Guarded so the entrypoint's
+    # `set -e` does not abort the remaining scopes.
+    if ! (unset RPM_ETCCONFIGDIR RPM_CONFIGDIR; rpm -qa --root="$root" __QUERY_FORMAT__); then
```
jetm, line 60: **A missing rpmdb exits 0, so the scope is dropped rather than failed**

The `##FAILED` safety net fires only on a non-zero `rpm -qa` exit. A sysroot whose RPM database is absent or wiped exits 0 with no output - verified locally: `rpm -qa --root=<dir with no var/lib/rpm> --qf '%{NAME}\n'` prints nothing and exits 0.

So a rootfs whose database was deleted or truncated produces `##SCOPE rootfs ...` with zero package lines and no `##FAILED`. `parse_sysroots` drops it at line 791, `execute()` sees a non-empty `sysroots` because other scopes survived, and the command prints a report and exits 0 with the rootfs entirely absent.

That falsifies the change's own CHANGELOG claim: "A sysroot whose RPM database cannot be read fails the command instead of being reported as empty." It also poisons `correlate()`'s baseline - line 812 then finds no `rootfs` scope, so every extension's inherited packages get attributed to the extension, which is the same wrong answer as the runtime finding above but reached by a different route.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1266 @@
...
+    printf '##SCOPE\t%s\t%s\n' "sdk" "$AVOCADO_SDK_PREFIX"
+    if ! sdk_query; then
+        printf '##FAILED\t%s\n' "sdk"
+    fi
+fi
+
+query_root "rootfs" "$AVOCADO_PREFIX/rootfs"
+query_root "initramfs" "$AVOCADO_PREFIX/initramfs"
+query_root "target-sysroot" "$AVOCADO_PREFIX/sdk/target-sysroot"
+query_root "includes" "$AVOCADO_PREFIX/includes"
```
jetm, line 84: **Legacy per-extension includes roots are never scanned**

`includes` is queried as one flat root, but legacy-layout remote extensions each get their own installroot. `ext_fetch.rs:197-201` picks `INSTALLROOT="$AVOCADO_PREFIX/includes"` only when the package advertises `avocado-ext-layout(nested)`; otherwise it uses `INSTALLROOT="$AVOCADO_PREFIX/includes/<ext_name>"`, giving that extension its own rpmdb.

`query_root "includes" "$AVOCADO_PREFIX/includes"` reads only `includes/var/lib/rpm`. For a project whose remote extensions are all legacy-layout that shared db does not exist, so per the finding above rpm exits 0 empty, the scope is dropped, and every package those extensions installed goes unscanned while the command reports success.

Worth contrasting with lines 86-99, which do correctly enumerate `runtimes/*/`, `runtimes/*/extensions/*/` and `extensions/*/`. `includes` is the one scope family that never iterates.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1266 @@
...
+        } else {
+            self.print_human(&source, &results, &cves);
+        }
+
+        // After the report is emitted either way: a gate that suppresses the
+        // findings it gates on would be useless.
+        if let Some(threshold) = self.fail_on_score {
+            let over: Vec<&str> = cves
+                .values()
+                .filter(|hit| hit.cve.score().is_some_and(|s| s >= threshold))
```
jetm: **Unscored CVEs can never trip `--fail-on-score`**

The release gate filters on `hit.cve.score().is_some_and(|s| s >= threshold)`, and `score()` returns `None` for any CVE with no CVSS fields or with cve-check's `"0.0"` placeholder. So unscored CVEs are treated as harmless at every threshold.

I ran the predicate directly. A CVE parsed from `{"id":"CVE-2026-X","summary":"critical, not yet scored by NVD"}` yields `score=None` and `trips gate: false` at 7.0, at 0.1, and even at 0.0; same for `{"scorev2":"0.0","scorev3":"0.0"}`. A CI job running `avocado cve report --fail-on-score 7.0` therefore exits 0 on an image whose only findings are NVD-unscored CVEs.

This contradicts the intent stated three lines above `score()` at line 145 - "so that reads as 'unscored' rather than 'harmless'". The human path hides them too: `ranked.sort_by` at line 649 puts `None` last via `Option` ordering and line 659 takes only the top 10, so they render as `-` below the cut or not at all. Neither output carries an unscored count.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1266 @@
...
+    /// could not be read must never be indistinguishable from one that holds
+    /// no packages.
+    failed: bool,
+}
+
+/// One installed package that a recipe with CVEs produced.
+#[derive(Debug)]
+struct Affected {
+    package: String,
+    /// VERSION-RELEASE as RPM reports it, directly comparable to report_version.
```
jetm: **JSON publishes `report_version` un-normalized**

`installed_version` is documented as "directly comparable to report_version", but `build_json` emits `report_version` verbatim from the source document while `installed_version` carries RPM's `-`->`+` rewrite. The two published fields are not comparable.

`rpm_pkgv` exists at line 267 precisely because they differ - its own comment counts 146 such packages in a scarthgap `bitbake world` (blktool, cargo-c, cni). So `build_json` lines 530-531 publish `installed_version: "4+7.1-r0.0"` alongside `report_version: "4-7.1-r0.0"` for the same package, with no per-entry flag saying the internal cross-check passed.

A JSON consumer that diffs the two fields to decide "is my report stale for this package" gets a false mismatch on every hyphenated PKGV - the exact class of false positive this command fixed for its own use and did not fix for its readers. The scope-level `version_mismatch` list is correct, but it is keyed by name only and lives outside the `affected` entries.

src/utils/lockfile.rs
```diff
@@ -185,7 +185,53 @@ pub struct RpmQueryConfig {
...
+    ///
+    /// `build_query_command` asks about a known package list; the CVE report
+    /// needs the full dependency closure instead, since a vulnerable package is
+    /// usually one nothing declared.
+    ///
+    /// Unlike `build_query_command` this does not append `|| true`. That guard
+    /// exists there because `rpm -q` returns non-zero merely when a named
+    /// package is absent; `rpm -qa` has no such case, so a non-zero exit is
+    /// always a real error and the caller must be able to see it.
+    pub fn build_query_all_command(&self) -> String {
```
jetm, line 213: **`build_query_all_command`'s `root_path` branch is unreachable**

Same ground as the thread I just conceded above, arriving from the other side: now that the interpolation is confirmed unreachable, the branch itself is dead production code rather than a hazard.

`report.rs:474` is the only non-test call site and it is `SysrootType::Sdk(target).get_rpm_query_config().build_query_all_command()`; `lockfile.rs:101-107` pins `root_path: None` for that variant. Per-root queries are hand-built in shell inside `DISCOVER_SCRIPT` at line 60 instead - which the `RPM_QUERY_ALL_FORMAT` doc comment at `lockfile.rs:189` already concedes ("callers that discover sysroots at run time have to build the rpm invocation in shell rather than here").

The cost is ~12 lines of unreachable production code plus `test_build_query_all_command_with_root`, which hand-constructs an `RpmQueryConfig` to pin the output of a branch nothing calls. That test will keep passing while the shell string that actually runs drifts away from it, which is the part that makes this worth removing rather than leaving.

src/utils/output_format.rs
```diff
@@ -55,6 +55,17 @@ pub fn emit_json_object(value: &serde_json::Value) {
     emit_json_event(value)
 }
 
+/// Emit one indented JSON object as the entire stdout output of a single-shot
+/// command. Same content as `emit_json_object`, formatted to be read by a human
+/// as well as parsed — for documents large enough that a single line is not
+/// reviewable. Never use for event streams: NDJSON consumers parse per line.
+pub fn emit_json_object_pretty(value: &serde_json::Value) {
```
jetm: **Only single-shot command emitting multi-line JSON**

`emit_json_object_pretty` makes `cve report` the sole `--output json` single-shot command in the repo that emits multi-line JSON; every sibling uses the compact one-line `emit_json_object`.

I checked five: `connect/projects.rs:26,87`, `connect/auth.rs:632,663,676,773,780,823`, `config_show.rs:51`, `profiles.rs:136,176`, `vm/config.rs:29,38` - all `emit_json_object`. The module doc at `output_format.rs:22-25` states the contract as "a single JSON object for single-shot commands, or NDJSON ... for long-running ones".

A consumer written against that contract - read one line, parse - gets a bare `{` from `avocado cve report --output json` and fails. Both forms are valid JSON to `jq`, so the harm is confined to line-oriented readers, but this is the only command in the repo that has the divergence.

jetm (commented):
Round 2 at `0655f4b`, seven findings inline. First, both of your pushbacks - I checked them rather than taking them on trust, and you are right on both.

The `root_path` injection is unreachable, exactly as you scoped it: `Sdk` pins `root_path: None` at `lockfile.rs:101-107` and `report.rs:472` is the only production construction reaching that call. I've withdrawn it on the thread, with one note about the sibling `build_query_command`, which does have a live interpolation - outside this PR, so not yours to fix here, but it means the "every value is a compile-time literal" generalization does not extend one function over.

`with_composed_config` I'm withdrawing too, and your read of the pattern is right: 35 definitions, 27 carrying `#[allow(dead_code)]`, 13 call sites - all three match what I measured. The only figure off is the conjunction: 25 carry both the attribute and the doc comment, not 27 (two have the attribute without the comment, at `initramfs/install.rs:80` and `rootfs/install.rs:972`, and eight have the comment without the attribute). The instance here is textually identical to that 25-member majority, so dropping it would make `cve report` the odd one out, which was your point. Leave it.

Of the seven new ones, three are the same bug reached three ways, and they compound. Runtime scopes get no inherited-baseline exclusion even though their installroots are seeded from the rootfs rpmdb by the identical `cp -rf`, so a runtime attributes the whole base system to itself - I ran `correlate()` and got `inherited=0, own=3, affected=2` where the `ext:` path returns `inherited=2, own=1, affected=0` for the same input. A sysroot whose rpmdb is missing exits 0 rather than non-zero, so it is dropped by `retain` instead of failing the command, which falsifies the CHANGELOG line and poisons the baseline the other two depend on. And legacy-layout remote extensions each have their own installroot that `includes` never iterates, unlike every other scope family.

The fourth is the release gate: unscored CVEs return `None` from `score()` and so cannot trip `--fail-on-score` at any threshold, including 0.0. I ran the predicate. A CI job gating at 7.0 exits 0 on an image whose only findings are NVD-unscored, and they are also invisible in the human path below the top-10 cut.

The rest are smaller: JSON publishes `report_version` un-normalized next to a normalized `installed_version` it calls comparable, the now-dead `build_query_all_command` root branch, and this being the only single-shot command emitting multi-line JSON.

Gates clean at `0655f4b` in a throwaway worktree with a private `CARGO_TARGET_DIR`: fmt, clippy `-D warnings` (forced a fresh recheck to rule out a cache hit), and `cargo test` including your 16 new unit and 11 new integration tests. I'm reviewer-role here, so I can't resolve any of these threads - the two withdrawals above are yours to close whenever you like.

hiagofranco:
Checked all reviews and five applied as written, three applied in parts:

- Fixed: runtime scopes count the whole rootfs as their own.
- Fixed: legacy per extension includes roots are never scanned.
- Fixed: build_query_all_command root_path branch is unreachable, removed.
- Fixed: only single-shot command emitting multi-line JSON.

- Half fixed: missing rpmdb exits 0, so the scope is dropped. Failing on an empty scope would be wrong: parse_sysroots already documents that a sysroot with no database is ordinary, and a project with no runtimes would start failing. What was wrong is the CHANGELOG sentence you quoted, so it was rewritten to say what actually happens.

- Half fixed: unscored CVEs can never trip `--fail-on-score`. I don't think the gate should change. `--fail-on-score 9.0` firing on a CVE with no score means the flag no longer means what it says, and every unscored CVE in a scarthgap build would trip every threshold (the flag becomes unusable). Option ordering puts them last, the top-10 cut drops them, and nothing counted them. Now `counts.cves_unscored` in the JSON, an info line in the human output, and the `--fail-on-score` help says outright that no threshold can match them.

- Half fixed: `report_version` has to stay, it's the key a consumer looks back up in the source document, and rewriting it there would break that. What was false is the doc comment on `installed_version` claiming the two are "directly comparable". Rewritten.

src/commands/cve/report.rs
jetm: Verified fixed. `Sysroot.failed` at report.rs:203, the `##FAILED` marker emitted per scope (:60-62, :76-78), `retain(|s| !s.packages.is_empty() || s.failed)` at :843, and `execute` bailing with the failed scopes named at :371-384. The unit test at :1033 pins it. A failed rpm query can no longer vanish into a clean-looking report.

src/commands/cve/report.rs
jetm: Verified fixed. The `|| true` is gone: `if ! (unset ...; rpm -qa ...); then printf '##FAILED...'` at :60-62, and the SDK query is guarded the same way at :76-78, with the caller bailing at :376-384. The regression test asserting the script does not contain `|| true` (:1056-1060) is the right shape - it pins the absence rather than the current spelling. The rationale recorded in lockfile.rs:203-206 is a nice touch.

src/commands/cve/report.rs
jetm: Verified fixed. `load_source` now bails on `source.recipes.is_empty()` at :430-436 and runs first in `execute` (:342), with the integration test pinning the message at tests/.../cve/report.rs:110-119.

One leftover, not worth another round: the unit-test comment at :1321-1323 still calls `{"recipes":{},"packages":{}}` "a legitimate document". True at the serde layer, but `load_source` now rejects it, so the comment reads as if the empty map were still accepted. Fold it in if you touch the file again.

src/commands/cve/report.rs
jetm: Verified fixed. The new `version_unchecked` list at :248, populated when either side has no version (:904-907), surfaced per-scope and in `counts` (:562, :620-621) and as a human warning (:735-743). Test at :1140. A missing version field is now visible rather than silently skipping the mismatch check.

src/commands/cve/report.rs
jetm: Verified fixed. `machine == target || machine.ends_with(&format!("-{target}"))` at :465. I re-ran my three original pairs against it - `avocado-qemuarm64`/`qemuarm`, `avocado-imx8mp-lpddr4`/`imx8mp`, `avocado-raspberrypi4-64`/`raspberrypi4` - and all three now warn where the substring test accepted them.

src/commands/cve/report.rs
jetm: Verified fixed. `if cve.score() > hit.cve.score() { hit.cve = cve.clone(); }` at :936-943, and `None < Some` makes the scored copy win regardless of iteration order - which is the part that matters, since the order was the bug. `the_highest_scoring_copy_of_a_cve_wins` at :1111-1137.

src/commands/cve/report.rs
jetm: Verified fixed. `--fail-on-score <SCORE>` at main.rs:1378-1383 (field :295), enforced with `bail!` at :396-409 after the report is emitted - emitting first and failing second is the right order, since a gating failure that also destroys the evidence is worse than no gate.

src/commands/cve/report.rs
jetm: Verified fixed. The CVE index now emits `"score"` and `"score_source"` at :585-586, backed by `resolved_score()`/`score_source()` at :161-174. A consumer can now see which score drove the ranking rather than re-deriving it and hoping it matches.

src/commands/cve/report.rs
jetm: Verified fixed. The baseline is now keyed on `(name, arch)` - `HashMap<(&str,&str),&str>` at :869-878, looked up with both at :888-890 - and `same_name_different_arch_is_not_treated_as_inherited` at :1176-1196 pins it.

src/commands/cve/report.rs
jetm: Half fixed - the data side landed, the human side did not.

The `recipe_missing` list at :252, populated at :922, and emitted as `recipes_missing` per-scope and in `counts` (:563, :622-623) with the test at :1160 - that all checks out.

What still stands is the point the finding actually made: `print_human` never mentions it. :730-771 warns on `unknown`, `version_mismatch`, `revision_mismatch` and `version_unchecked` only, so a truncated `recipes` map is still warning-free on the human path. The comparison I drew was to `unknown`, which does warn - a package whose recipe entry is missing is in the same class as a package whose recipe is unknown, and it should be as visible. Leaving this open for one line in the warn block.

tests/commands/avocado/cve/report.rs
jetm: Verified fixed, and more thoroughly than I asked. `assert_fails_with` at :34-41 asserts exit non-zero **and** the expected text, and every negative test now pins its own message - `missing field packages`, `has no 'packages' map`, `missing field recipes`, `empty 'recipes' map`, `not 'Unpatched'`, `Failed to parse/read CVE report`. The switch to `common::run_cli` plus the module doc at :8-13 naming the exact false-pass is what makes this stay fixed rather than regress the next time a test is added.

src/utils/lockfile.rs
jetm: Verified fixed, and by removal rather than escaping - which is the better answer. `build_query_all_command` no longer has a `root_path` branch (:213-224), so nothing caller-supplied reaches the string at all, and `test_sdk_query_config_ignores_target_value` (:1798-1806) pins that a `$(id); "` target produces identical output.

The untested-format half is closed too: `test_rpm_query_all_format_is_three_tab_separated_fields` (:1752-1757) pins the exact `--qf` the tab parser in report.rs depends on, so the two can no longer drift apart silently.

src/commands/cve/report.rs
jetm: Verified fixed. The exclusion is now `scope.starts_with("ext:") || scope.starts_with("runtime:")` at :881-882, with the `cp -rf .../rootfs/var/lib/rpm` rationale recorded at :860-863 - worth having written down, since the reason is not visible from the code. Test asserts `results["runtime:dev"].inherited == 1` at :1219.

src/commands/cve/report.rs
jetm: The doc half is genuinely fixed - the CHANGELOG sentence I quoted is rewritten (CHANGELOG.md:22-25) and now says what the code does. Thanks.

The behavior half stands, and I want to push back on the reason given for keeping it. `retain` still drops an empty non-failed scope (:840-843), so an existing-but-wiped `rootfs` still yields a full report at exit 0.

On "a project with no runtimes would start failing": `query_root` returns early on a non-existent directory (:49), so a missing dir emits no `##SCOPE` line at all and cannot reach the retain. Only directories that *exist* and yield zero packages are affected - and a rootfs in that state is not the ordinary case `parse_sysroots` documents, it is a broken one. So a guard scoped to `rootfs` specifically was available without touching the runtime case.

The new mitigation at :669-680 is a human-output warning gated on `!results.contains_key("rootfs") && any ext:/runtime: scope`. Two gaps: `JsonOutputGuard` (:340) suppresses it under `--output json` with no equivalent JSON field, so the machine-readable consumer - the one that would gate CI on this - sees nothing; and it never fires for a project with no ext/runtime scopes at all.

Not blocking on my account, but it is still a silent-zero path.

src/commands/cve/report.rs
jetm: Verified fixed. `for inc_dir in "$AVOCADO_PREFIX"/includes/*/ ... query_root "includes:$(basename "$inc_dir")"` at :91-94, alongside the flat root at :84, with the nested-vs-legacy layout rationale at :85-90 and the scope name documented in CHANGELOG.md:15-16. Test at :1064-1071.

src/commands/cve/report.rs
jetm: Your argument holds and I withdraw the gate change. A CVSS threshold cannot rank an unranked item, and auto-tripping every threshold on unscored CVEs would make `--fail-on-score` unusable on a scarthgap build - which is worse than the gap I was pointing at. The right shape for that is a separate opt-in, and I did not ask for one.

The invisibility was two-thirds of what I raised and it is fixed: `counts.cves_unscored` in the JSON (:628), the info line in the human output (:719-728), and `--fail-on-score`'s help saying outright that no threshold can match them (main.rs:1379-1381). Saying it in the help text is the part that stops the next person filing this same finding.

Resolving. Noting for later, not as a blocker: gating on unscored findings still means parsing the JSON.

src/commands/cve/report.rs
jetm: Agreed on `report_version` staying - it is the key a consumer looks back up in the source document, and normalizing it there would break that lookup. That part I withdraw.

The correction did not land where the finding was, though. `installed_version`'s doc comment no longer claims the two are "directly comparable" (:210-218), and that is right - but it is a Rust doc comment, and the consumer this affects reads JSON, not source. `build_json` still publishes `installed_version` and `report_version` raw side by side (:543, :545) with no per-entry verdict, so a consumer diffing them still gets a false mismatch on every hyphenated PKGV - the 146-package class that `rpm_pkgv` exists for (:273-283).

A per-`affected` flag, or a normalized companion field, is what closes it. Small, and it means a consumer does not have to reimplement `rpm_pkgv` to read the report correctly.

src/utils/lockfile.rs
jetm: Verified fixed. The `--root` branch is deleted (:213-224) and the doc at :208-212 records why root_path is deliberately not honoured. Replacing `test_build_query_all_command_with_root` with `test_build_query_all_command_ignores_root_path` (:1760-1775) is the right move - it pins the absence, so the dead branch cannot quietly come back.

src/utils/output_format.rs
jetm: Verified fixed. report.rs now imports and calls the compact `emit_json_object` (:17, :389), matching every sibling single-shot command, and `src/utils/output_format.rs` is out of the PR's changed files entirely at this head.

jetm:
Went back through all 19 of my threads against this head and replied on each with what I found. Tally: 15 fixed, 3 partial, 1 I withdraw.

The three partials are `recipes_missing` (in the JSON, still absent from the human warn block), the empty-`rootfs` silent-zero path (CHANGELOG corrected, `retain` unchanged, and the new warning is suppressed under `--output json`), and `report_version` (the "directly comparable" claim is gone from the doc comment, but the JSON a consumer actually reads still publishes both versions raw).

The unscored-CVE gate I withdraw outright - your argument that a CVSS threshold cannot rank an unranked item holds, and putting it in `--fail-on-score`'s help is what stops the next reviewer filing it again.

Detail is on each thread. Resolving is yours - the tooling I use deliberately refuses to resolve threads on a PR I did not author.

hiagofranco:
> Went back through all 19 of my threads against this head and replied on each with what I found. Tally: 15 fixed, 3 partial, 1 I withdraw.
> 
> The three partials are `recipes_missing` (in the JSON, still absent from the human warn block), the empty-`rootfs` silent-zero path (CHANGELOG corrected, `retain` unchanged, and the new warning is suppressed under `--output json`), and `report_version` (the "directly comparable" claim is gone from the doc comment, but the JSON a consumer actually reads still publishes both versions raw).
> 
> The unscored-CVE gate I withdraw outright - your argument that a CVSS threshold cannot rank an unranked item holds, and putting it in `--fail-on-score`'s help is what stops the next reviewer filing it again.
> 
> Detail is on each thread. Resolving is yours - the tooling I use deliberately refuses to resolve threads on a PR I did not author.

Thanks for the reviews. I pushed the four fixes proposed, please take a look.

src/main.rs
```diff
@@ -1346,6 +1352,38 @@ enum ConnectClaimTokensCommands {
...
+        container_args: Option<Vec<String>>,
+        /// Output format
+        #[arg(long, value_enum, default_value_t = OutputFormat::Human)]
+        output: OutputFormat,
+        /// Exit non-zero if any CVE scores at or above this CVSS value, for
+        /// use as a release gate. The report is still printed. CVEs carrying
+        /// no CVSS score cannot match any threshold; the report counts them
+        /// separately as cves_unscored.
+        #[arg(long, value_name = "SCORE")]
+        fail_on_score: Option<f64>,
```
jetm, line 1391: **`--fail-on-score` fails open on nan, inf, and out-of-range values**

`fail_on_score: Option<f64>` has no `value_parser`, so clap accepts `nan`, `NaN`, `inf`, `Infinity`, `1e400`, `99` and `=-1`. I checked all of these against the built binary (each reaches `load_source`), and ran the exact filter from `report.rs:404` in a probe: `threshold=NaN` matches nothing, `threshold=99` matches nothing.

So CI running `avocado cve report -f r.json --fail-on-score 99` - the 0-100 scale confusion, which is an easy mistake to make against a CVSS 0-10 field - exits 0 against a report containing a 9.8 CVE, and nothing says the threshold was uninterpretable. Same for `nan` arriving from a bad template variable. Only bare `-1` and non-numeric text are rejected today.

A `value_parser` with a 0.0-10.0 finite range turns every one of these into a startup error instead of a silent pass.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+            anyhow::bail!(
+                "CVE report '{}' has an empty 'recipes' map, so nothing could be correlated. \
+                 The producing build most likely ran without cve-check inherited.",
+                self.file
+            );
+        }
+        // The producer records which cve-check status it filtered on. Anything
+        // but Unpatched describes issues that are not live findings, and
+        // correlating it would present them as if they were.
+        if !source.status.is_empty() && source.status != "Unpatched" {
```
jetm: **The Unpatched gate is bypassed by omitting the `status` key entirely**

`status: String` carries `#[serde(default)]` while the check is `if !source.status.is_empty() && source.status != "Unpatched"`. So an absent key yields an empty string and skips the bail.

What makes this look unintended: `recipes` and `packages` deliberately *omit* `#[serde(default)]` for exactly this hazard, with the comment at :124 saying so. `status` got the opposite treatment. And `report.rs:815` then relabels the absent case as `"unpatched"` under a comment reading "load_source has already rejected any status but Unpatched" - which is false precisely for the absent case.

A document generated with `AVOCADO_CVE_REPORT_STATUS=Patched` but with `status` trimmed or absent therefore parses, skips the bail, and prints "N unpatched CVE(s)" for CVEs that are all already fixed, while the JSON publishes `"status": ""`.

Dropping the `#[serde(default)]` would make it match its two siblings.

src/main.rs
```diff
@@ -3158,6 +3199,30 @@ async fn main() -> Result<()> {
                 Ok(())
             }
         },
+        Commands::Cve { command } => match command {
```
jetm, line 3210: **Global `--runs-on` is silently ignored, so the report describes the wrong host**

`runs_on` and `nfs_port` are `global = true` (main.rs:98-104), but this arm threads neither, `CveReportCommand` has no such field, and `RunConfig { ..Default::default() }` leaves `runs_on: None`. `Commands::Cve` was added to `needs_vm_routing` (main.rs:2049), and `ensure_routed_for_process` skips VM routing precisely when `runs_on.is_some()` - so the request is neither honoured nor routed.

Compounding it: `run_in_container_capture` (container.rs:1472-1480) has no remote branch at all, going straight to `VolumeManager::get_or_create_volume(&self.cwd)`. So adding `.with_runs_on(..)` here alone would not fix it.

`avocado --runs-on user@buildbox cve report -f r.json` therefore reads the **local** volume: with no local install it bails "Run `avocado install` first", and with a stale local install it emits a confident exit-0 CVE report for the wrong tree.

For context, 7 of 17 `needs_vm_routing` groups omit `with_runs_on` (`Clean`, `Deploy`, `Sign`, `Prune`, `Hitl`, `Connect::Upload`, and now `Cve`), so this joins an existing inconsistency rather than creating it - but it matters more here than elsewhere, because the output is a security report attributed to a host.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+
+        if self.output.is_json() {
+            emit_json_object(&self.build_json(&source, &target, &results, &cves));
+        } else {
+            self.print_human(&source, &results, &cves);
+        }
+
+        // After the report is emitted either way: a gate that suppresses the
+        // findings it gates on would be useless.
+        if let Some(threshold) = self.fail_on_score {
```
jetm, line 454: **The release gate has no behavioral test coverage, and no way to scope it**

Two things about this block, both worth addressing together since anyone touching one will be looking at the other.

First, coverage. The only test that touches it (`tests/commands/avocado/cve/report.rs:59`) greps `--help` for the literal flag string. The threshold comparison, the `bail!`, and the emit-then-gate ordering are never executed. Flip `>=` to `>` and a CVE scoring exactly the threshold passes the gate with every test still green; delete the whole `if let Some(threshold)` block and the flag becomes a no-op while CI reports success. For a release gate that is the one test that matters.

Second, breadth. `over` is built from the whole deduped `cves` map, which unions every scope - including `sdk` (host nativesdk tooling) and `target-sysroot` (cross-compile `-dev` deps, sdk/install.rs:256-260 and 398). Neither ships in the image. The per-CVE `scopes` set is in the JSON so an external consumer *can* filter, but the built-in gate cannot: there is no `--scope` or `--exclude-scope`. So `--fail-on-score 7.0` blocks a release because the SDK's host `python3` or a target-sysroot `openssl-dev` scores 9.8, while `rootfs` and `initramfs` are clean.

jetm (commented):
Second pass over this at 79895a7, with the CVE report command built and exercised in a scratch worktree: `cargo fmt --check` and `cargo clippy --all-targets --all-features -D warnings` both exit 0, and the 24 new unit tests plus 14 new integration tests pass (the integration ones need no Docker - they all bail inside `load_source`). Twelve new findings inline; my four earlier threads are still open separately and I have not restated them here.

The two I would fix before merge both make a release gate pass when it should fail. `--fail-on-score` has no `value_parser`, so `nan`, `inf` and `99` are all accepted and all match nothing - I verified `threshold=NaN` and `threshold=99` against the real filter, so a 9.8 CVE exits 0 with no indication the threshold was uninterpretable. And the Unpatched check is bypassed by omitting `status` entirely, because `#[serde(default)]` is on that field where its two siblings deliberately do without it for this exact reason. A "Patched" document with the key trimmed prints "N unpatched CVE(s)" for CVEs that are all fixed.

Worth calling out as a theme rather than a single finding: several of these are cases where the report is confidently wrong about *what it scanned* rather than about a CVE - the ignored global `--runs-on` (reads the local volume while naming a remote host), the never-refreshed ext/runtime rpmdb seed (attributes rootfs packages to an extension), and the installroot plumbing directories published as scopes. For an artifact whose whole job is provenance, those seem worth more than their individual severity suggests.

Two I deliberately scoped down rather than overstating. The ext/runtime misattribution inflates by the packages whose full `PKGV-PKGR` moved since the seed, typically one plus its closure - not the whole base system, which I initially thought; the direction is over-attribution so nothing is hidden. And the `Ok(None)` laundering is a pre-existing repo-wide pattern in untouched code - what is new here is only the misleading bail text at this call site.

Refuted, so nobody re-derives them: the kernel sysroot omission is correct (those are `cp -a`'d boot artifacts with no rpmdb, and the packages are already in the rootfs scope); `split_version` cannot misparse a hyphenated PKGV because `PR = "r0"` is unconditional in bitbake.conf:245; `set -u` on the prefix variables is safe since the entrypoint exports both unconditionally; shell injection via `--target` does not reach the script (the `Sdk(_arch)` arm discards it); and the `##SCOPE`/`##FAILED` marker protocol is the third instance of an existing in-repo pattern rather than a reinvention.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+
+    for sysroot in sysroots {
+        let seeded_from_rootfs =
+            sysroot.scope.starts_with("ext:") || sysroot.scope.starts_with("runtime:");
+        let result = results.entry(sysroot.scope.clone()).or_default();
+        result.root = sysroot.root.clone();
+        result.scanned += sysroot.packages.len();
+
+        for installed in &sysroot.packages {
+            let inherited = baseline
```
jetm: **A stale-seeded ext/runtime rpmdb misattributes rootfs packages as extension-owned**

Ext and runtime installroots are seeded once with a copy of `$AVOCADO_PREFIX/rootfs/var/lib/rpm`, guarded on the directory not existing (ext/install.rs:600-603, ext/dnf.rs:218-249, runtime/install.rs:488-513, runtime/dnf.rs:195-221), and never refreshed - nothing in `src/commands/rootfs/` invalidates them. `inherited` then requires an exact `PKGV-PKGR` match against the *current* rootfs.

So: rootfs at libssl3 `3.5.7-r0.0`, install ext `sshd`, upgrade rootfs to `3.5.8-r0.0`. On the next run `ext:dev/sshd` gets `packages_own=1`, `affected=["libssl3"]`, and that scope is added to `cves[...].scopes` - a vulnerable openssl reported as shipped by an extension that never installed one. A PKGR-only rebuild bump triggers it too. When the report describes the *old* generation, no warning of any kind fires on that scope; the newer-generation case only downgrades to `print_info`.

One correction to my own first read, so the magnitude is not overstated: the inflation is the number of rootfs packages whose full `PKGV-PKGR` moved since the seed (typically one plus its closure), not the whole base system - that only holds after a whole-image refresh. The direction is over-attribution, so nothing is hidden, but the scope attribution in a published security artifact is wrong.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+        // below so a failed query is named as one rather than reported as
+        // nothing to scan.
+        let failed: Vec<&str> = sysroots
+            .iter()
+            .filter(|s| s.failed)
+            .map(|s| s.scope.as_str())
+            .collect();
+        if !failed.is_empty() {
+            anyhow::bail!(
+                "Could not read the RPM database of {} scope(s): {}. Those sysroots were not \
```
jetm, line 422: **This bail's own remedy is unreachable: rpm's stderr is discarded**

The message says "Re-run with `--verbose` to see rpm's own error." But the `##FAILED` guard makes the script exit 0, so `run_in_container_with_output` takes `if out.success { return Ok(Some(out.stdout...)) }` (container.rs:1449-1451) and drops `out.stderr`, which is referenced only in the `!success` arm. stderr is piped rather than inherited (container.rs:1582), and `--verbose` prints only the mount line and the docker argv (container.rs:1564-1576).

So on a corrupt `rootfs/var/lib/rpm` the user gets a bail naming the scope, `--verbose` adds nothing, and there is no route to rpm's actual `error: cannot open Packages database` short of running the container by hand.

The sibling `run_in_container_capture` returns `{success, stdout, stderr}` and would make the advice true.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+query_root "initramfs" "$AVOCADO_PREFIX/initramfs"
+query_root "target-sysroot" "$AVOCADO_PREFIX/sdk/target-sysroot"
+query_root "includes" "$AVOCADO_PREFIX/includes"
+# A remote extension advertising `avocado-ext-layout(nested)` installs into the
+# shared includes root queried above; one without it gets an installroot, and so
+# an RPM database, of its own (utils/ext_fetch.rs). Both layouts land their
+# content in includes/<name>/, so the per-extension roots have to be queried too
+# or a project whose extensions are all legacy-layout scans nothing here. Nested
+# ones hold no database of their own and drop out as an empty scope.
+for inc_dir in "$AVOCADO_PREFIX"/includes/*/; do
```
jetm, line 91: **Installroot plumbing directories become phantom scopes in the published JSON**

For a nested-layout project, `$AVOCADO_PREFIX/includes` **is** itself a dnf `--installroot` (ext_fetch.rs:194-199, documented at :142-146), so its top-level entries are `var/`, `etc/`, `usr/` - not extension names.

I ran the substituted `DISCOVER_SCRIPT` against such a tree and it emitted `##SCOPE includes:etc`, `includes:myext`, `includes:usr`, `includes:var`. The realpath dedup does not catch them (the realpaths genuinely differ), and every marker becomes a `results` entry unconditionally at :930-932.

So any project with one nested-layout extension prints table rows `includes:var 0 0 0`, `includes:etc 0 0 0`, `includes:usr 0 0 0` and publishes `scopes["includes:var"]` and friends in the JSON artifact - non-existent sysroots presented as scanned scopes, plus a wasted `rpm -qa` per phantom directory.

(The related theory that this hard-aborts is refuted: `rpm -qa --root=<empty dir>` exits 0 with no output on rpm 6.0.1.)

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+        self.resolved_score().map(|(value, _)| value)
+    }
+
+    /// Which CVSS version the resolved score came from, so the JSON document
+    /// says what it ranked by instead of leaving a consumer to guess.
+    fn score_source(&self) -> Option<&'static str> {
+        self.resolved_score().map(|(_, source)| source)
+    }
+
+    fn resolved_score(&self) -> Option<(f64, &'static str)> {
```
jetm, line 168: **A non-finite parsed score contradicts itself across every output**

Same root as the `--fail-on-score` parsing finding, but on the ingest side: `"inf"` and `"1e400"` pass the `> 0.0` gate here, so `score()` returns `Some(inf)`. The JSON then publishes `"score": null` beside `"score_source": "v3"`, `counts.cves_unscored` excludes it, the gate trips at *every* threshold, and human output ranks it first printing `inf`.

`"NaN"` behaves correctly by comparison - it falls through to unscored.

The input is contrived (a producer would have to emit it), so I am flagging it as hardening rather than a live bug. `value > 0.0 && value.is_finite()` on the line already being written covers it.

src/main.rs
```diff
@@ -1346,6 +1352,38 @@ enum ConnectClaimTokensCommands {
...
+        #[arg(short = 'c', long, default_value = "./avocado.yaml")]
+        config: String,
+        /// Target architecture
+        #[arg(short = 't', long)]
+        target: Option<String>,
+        /// Verbose output
+        #[arg(short, long)]
+        verbose: bool,
+        /// Additional arguments to pass to the container runtime
+        #[arg(long, value_delimiter = ' ', num_args = 1..)]
```
jetm: **`--container-args` cannot express hyphen-prefixed values, and the sibling spelling is rejected**

36 siblings use `long = "container-arg", num_args = 1, allow_hyphen_values = true, action = Append`. This one uses `long, value_delimiter = ' ', num_args = 1..`. Against the built binary:

- `--container-arg --privileged` gives `error: unexpected argument '--container-arg' found`
- `--container-args --privileged` gives `error: unexpected argument '--privileged' found`
- `--container-args="-v /my dir:/b"` parses to `["-v", "/my", "dir:/b"]` (probed with the identical attrs on clap 4.6)

Only `--container-args=--privileged` works, which is undiscoverable from `--help` and still cannot express any value containing a space. Matching the sibling attrs fixes both halves.

Related and cheaper to fix now than later: this subcommand takes `-c` for `--config` where 78 sibling uses take `-C` (2 take `-c`), so `avocado cve report -C avocado.yaml` gives `error: unexpected argument '-C' found`. The repo's own `insert_config_flag` helper (tests/common/mod.rs:86) hardcodes `-C` and so cannot be used against this subcommand. One character now; a breaking change after `1.0.0-rc.1`.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+                installed_version: installed.version.clone(),
+                arch: installed.arch.clone(),
+                report_version: source_package.version.clone(),
+                recipe: source_package.recipe.clone(),
+                cves: recipe.cves.iter().map(|c| c.id.clone()).collect(),
+            });
+        }
+
+        result.affected.sort_by(|a, b| a.package.cmp(&b.package));
+        result.unknown.sort();
```
jetm, line 1181: **Multilib double-counts: 4 of the 5 name lists are sorted but not deduped**

`recipe_missing` gets `.sort()` then `.dedup()`; `unknown`, `version_mismatch`, `revision_mismatch` and `version_unchecked` get only `.sort()` - on adjacent lines, which is what makes it read as an oversight rather than a decision.

It matters because this file's own baseline comment at :912-915 and the `same_name_different_arch_is_not_treated_as_inherited` test both assert that one sysroot can hold two same-named packages of different arch.

So a rootfs holding `libfoo` at `core2_64` and `i686`, neither in the report, makes `counts.packages_unknown` read 2 for a single package name, and `scopes.rootfs.unknown` ships a duplicate. A CI gate ratioing `packages_unknown / packages_scanned` is then wrong by the multilib count.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+        // A plain `contains` was wrong: "avocado-qemuarm64" contains "qemuarm",
+        // as does every foo/foo64, imx8mp/imx8mp-lpddr4 and
+        // raspberrypi4/raspberrypi4-64 pair this project builds. Those images
+        // share package names but neither versions nor patch sets, which is
+        // exactly what this warning exists to catch. MACHINE names are
+        // "avocado-<target>", so require that whole suffix.
+        if machine == target || machine.ends_with(&format!("-{target}")) {
+            return;
+        }
+        print_warning(
```
jetm, line 516: **The MACHINE-mismatch warning has no JSON counterpart**

`JsonOutputGuard` (enabled at :340) routes through `tui_is_active()` (output.rs:25-28) and silences `print_warning` (output.rs:58-62), and `print_human` is not called at all in JSON mode (:393-397). Meanwhile `build_json` emits `source.machine` and `target` as raw strings with no mismatch flag - so a consumer has to reimplement `machine == target || machine.ends_with("-{target}")`, which is exactly the rule whose naive `contains` form was already gotten wrong once and fixed (your comment at :466-473).

That also cuts against this PR's own stated principle behind adding `score`, `score_source`, `report_version_rpm` and `cves_unscored` to the JSON: put the derived judgement in the document rather than making every consumer re-derive it.

Concretely, `--output json --fail-on-score 7` against a report for `avocado-imx8mp-lpddr4` while targeting `imx8mp` exits 0 with a valid document and zero signal that the correlation is untrustworthy.

src/commands/cve/report.rs
```diff
@@ -0,0 +1,1461 @@
...
+            repo_release: config.get_sdk_repo_release(),
+            container_args: config.merge_sdk_container_args(self.container_args.as_ref()),
+            sdk_arch: self.sdk_arch.clone(),
+            ..Default::default()
+        };
+
+        let output = container
+            .run_in_container_with_output(run_config)
+            .await?
+            .unwrap_or_default();
```
jetm: **A container failure is laundered into "Run `avocado install` first"** (lower confidence on scope, not on mechanism)

`run_in_container_with_output` returns `Ok(None)` rather than `Err` on a non-zero exit and discards captured stdout (container.rs:1446-1467 - whose own comment warns "otherwise callers convert `None` into misleading errors"). Here `.unwrap_or_default()` gives `""`, an empty `Vec`, `.all()` is vacuously true, and the :384 bail fires.

So docker being down, an image pull failure, or an entrypoint abort on a fully-installed project produces the real container error (that part *is* printed) **and then** false advice to install, with any partial scope output thrown away.

Scoping this honestly: the `Ok(None)` pattern is pre-existing and repo-wide (`prerequisites.rs:56`, `ext/clean.rs:235`, `ext/build.rs:303`, `kernel_resolver.rs:286` all do it) and `container.rs` is not touched by this PR. What is new is the misleading bail text at this call site. Switching to the sibling `run_in_container_capture` would fix both this and the `--verbose` remedy finding above.

hiagofranco:
Output generated with Opus 5:

Fixes: 

- --fail-on-score fails open — fixed
- No behavioral coverage for the gate — fixed
- Absent status bypasses the Unpatched gate — fixed by dropping #[serde(default)] (report.rs:126)
- Non-finite parsed score — fixed, value > 0.0 && value.is_finite() (report.rs:176), on the line you pointed at.
- --verbose remedy unreachable and container failure laundered into "run avocado install" — fixed together by switching to run_in_container_capture
- MACHINE mismatch has no JSON counterpart — fixed
- Stale seeded ext/runtime rpmdb — fixed as far as it can be
- --container-args and -c — fixed both

Other comments:

--runs-on silently ignored — refused rather than ignored (report.rs:382, main.rs:3229). Threading with_runs_on alone would not have helped, exactly as you said: run_in_container_capture has no remote branch, only run_in_container routes on config.runs_on. Implementing remote capture is bigger than this PR; until then a security report attributed to a host it did not read is worse than an error.

Multilib dedup — I disagree, and added a test to pin the reasoning. recipe_missing is deduped because it names recipes, and one recipe backs many packages, so a repeat restates one fact. The other four name packages, and multilib libfoo at core2_64 and i686 is two packages — counts.packages_unknown counts packages, so 2 is right. Deduping by name would leave scopes.rootfs.unknown.len() disagreeing with the count it is the detail for. Comment at the sort site and multilib_names_stay_in_the_lists_that_count_packages record it, so the next reader does not file this again.

Installroot plumbing as phantom scopes — I don't think this reaches the output. retain at report.rs:1052 drops any empty non-rootfs scope before correlate ever sees it, so an includes:var marker with no packages never becomes a results entry.

I have direct evidence rather than a reading: on a nested-layout qemuarm64 project whose tree really does contain includes/etc, includes/opt and includes/var, cve report --output json published exactly 8 scopes and no includes:*. I know the directories were scanned because a second command I am building on the same discovery walk — which did not have the filter — listed all three at 0 packages. That is how I found the bug in that command.

What does survive from the finding is the wasted rpm -qa per plumbing directory. That is cost inside one already-running container rather than a wrong document, so I left it; say the word if you would rather the walk skipped directories with no var/lib/rpm.

Not doing: --scope / --exclude-scope. You are right that sdk and target-sysroot inflate the gate — I hit the same thing elsewhere and ended up excluding build-host scopes by default there. But it is a new flag and a new default mid-review, and the JSON already carries the per-CVE scopes set an external gate can filter on. Happy to do it as a follow-up if you want it before 1.0.0.

cargo test --lib: 1146 passed. Integration: 18. Clippy clean.

hiagofranco:
Last push: I rebased to fix the CHANGELOG conflict.

hiagofranco:
As discussed today, I will be closing this PR for now. I can reopen if needed.

PR #191, opened 2026-08-06
ci: test and publish against the 2024 'next' channel

mobileoverlord:
## Problem

The 2024 legs of both extension workflows still name the `edge-next` channel, which is the pre-rename name for what is now `next` in the `next -> edge -> stable` promotion chain. CI has been exercising a channel under its old name.

## Change

Two lines:

- `.github/workflows/ext-test.yml` — the `build` matrix's only entry, `2024/edge-next` -> `2024/next`
- `.github/workflows/ext-release.yml` — the 2024 leg of the `publish` matrix, `2024/edge-next` -> `2024/next`

The 2026 leg of `ext-release` already tracks `edge` and is untouched.

The channel is a feed path segment — the reusable workflow documents the feed as `<repo-url>/<distro-release>/<distro-channel>/...` — so this repoints both jobs at the renamed path. CI on this PR is the verification: the `build (qemux86-64, 2024, next)` job either resolves the feed or it does not.

## Note for a follow-up, not changed here

`ext-test` covers only the 2024 leg while `ext-release` publishes to both `2026/edge` and `2024/next`, so the 2026 target is published without ever being tested. Worth adding a `2026/edge` entry to the test matrix, but that is a scope change rather than a rename and belongs in its own PR.

### DIFF
--- a/.github/workflows/ext-release.yml
+++ b/.github/workflows/ext-release.yml
@@ -49,7 +49,7 @@ jobs:
       matrix:
         include:
           - { target: qemux86-64, distro-release: "2026", distro-channel: edge }
-          - { target: qemux86-64, distro-release: "2024", distro-channel: edge-next }
+          - { target: qemux86-64, distro-release: "2024", distro-channel: next }
     uses: avocado-linux/actions/.github/workflows/extension-release.yml@v1
     with:
       target:         ${{ matrix.target }}
--- a/.github/workflows/ext-test.yml
+++ b/.github/workflows/ext-test.yml
@@ -42,7 +42,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { target: qemux86-64, distro-release: "2024", distro-channel: edge-next }
+          - { target: qemux86-64, distro-release: "2024", distro-channel: next }
     uses: avocado-linux/actions/.github/workflows/extension-test.yml@v1
     with:
       target:         ${{ matrix.target }}


jetm (approved):
Approving. I checked the thing most likely to be wrong in a rename like this - whether every occurrence moved - and it did: `git grep -E 'edge-next|edge_next'` over the whole tree at `d025616f` returns zero hits, and the only channel literals left in `.github/` are the three matrix lines, so no untouched job still references the old name.

Two things I confirmed rather than assumed, since both could have bitten:

`next` is not a YAML 1.1 boolean (`n`/`no`/`on` are; `next` is not), so it loads as the string it looks like, and `distro-release` stays `"2024"` as a string - the existing quotes there are still load-bearing and you kept them. And the SDK image tag interpolates `AVOCADO_DISTRO_RELEASE` only, so the 2024 leg still pulls `sdk:2024` and needed no edit.

One caveat I want on the record rather than buried: the only checks on this head are the three CodeQL legs. `ext-test` - the workflow this PR edits - produced no run here, so nothing in CI actually demonstrates that the `2024/next` feed exists. That premise lives in the reusable workflow and the feed, neither of which is visible from this repo. I am approving on the reasoning that if the channel is wrong this fails loudly on that leg rather than silently publishing to a bad path, so the cost of being wrong is a red build and not a corrupted release. Worth a glance at the first `ext-test` run that does fire.

Unrelated to this change, so not a request: `ext-release` publishes two legs (`2026/edge` and `2024/next`) while `ext-test` gates only one, and with `fail-fast: false` a 2026 break still lets 2024 publish. I checked `origin/main` and that asymmetry is identical before this PR, so the rename neither widens nor re-exposes it.

PR #196, opened 2026-08-11
fix(vm): reset var on a version bump so extensions match the rootfs

lee-reinhardt:
## Problem

`avocado vm update` replaces kernel/initramfs/rootfs and leaves `var` alone (`seed_only`). But `var` holds more than user state: the VM's own system extensions live in `/var/lib/avocado/images` and are merged onto `/usr` and `/etc` at boot. So an update boots a new kernel and rootfs against the old userspace.

Nothing catches this. Each extension's `extension-release` is synthesized at merge time with `ID=_any` (systemd-sysext's "matches any OS" wildcard), so stale extensions merge silently and the VM looks healthy while running old code.

## Change

- On a version bump whose var image sha differs from the installed one, fetch the new seed and delete the live `var.btrfs` after committing artifacts. `lifecycle::start` already re-seeds a missing disk and re-applies the configured size, so the boot path stays the single owner of both, and a mid-update crash leaves no half-written disk.
- The reset fires only when the var sha actually changed (a boot-artifact-only release costs neither the ~450 MB download nor the user's state), an installed manifest exists (a first install can't delete a disk seeded from a dev `--vm-source` tree), and the live disk exists (a never-started VM has nothing to reset, and host applications don't see a `var_reset` for a reset that didn't happen). A seed file that merely went missing from the install dir is re-fetched at the same sha without touching the live disk.
- The disk's size is recorded as `runtime.var_size` in `config.yaml` before the delete, and `vm start` resolves size as flag > config > default. Growing the disk used to be sticky only via the file itself, which the reset would otherwise quietly shrink back to the 50G default.
- Authorization is explicit. Interactively, the confirmation spells out what is reset and requires typing `update` (matching `vm reset`). With `--yes`, a distinct `--reset-var` flag is required — scripts that pass `-y` to skip a y/N never wipe state they didn't opt into. `--output json` never prompts (it would corrupt the NDJSON stream) and emits a `{"event":"var_reset"}` line so host applications can drop cached install state and tell the user to re-run `avocado install` / `avocado build`.
- Surface the `min_cli_version` refusal instead of swallowing it: previously an incompatible CLI reported "no update available" forever. Now `vm update` fails with the actionable "run `avocado upgrade` first" message, and under `--output json` a structured `{"update_available": true, "cli_too_old": true, ...}` line goes out before the non-zero exit so stdout parsers can tell "blocked on CLI version" from "check broke". Newness is checked before compatibility so an already-installed release stays quiet.
- Fetch the remote manifest once instead of twice, so the installed manifest always describes the artifacts that were actually committed.

## Testing

- 17 new unit tests over `plan_downloads`, `should_reseed_var`, `record_var_size_floor`, config round-trip, and the update-check decision.
- End-to-end against a fake channel: too-old CLI fails with the actionable message; changed var sha re-seeds and emits `var_reset`; unchanged var sha leaves the live disk byte-for-byte; aborting the prompt leaves everything intact.
- Validated against a real running VM: update stopped the VM, reset var, restarted; the new extensions were merged after boot, SSH reconnected across the host-key regeneration, and the var disk was re-grown to its configured size.

### DIFF
--- a/src/commands/vm/update.rs
+++ b/src/commands/vm/update.rs
@@ -6,11 +6,31 @@
 //!
 //! - `replace` — always re-downloaded when the sha differs (kernel,
 //!   initramfs, rootfs).
-//! - `seed_only` — downloaded only on first install. On subsequent
-//!   updates we skip it entirely so the user's `var` (Docker volumes,
-//!   project caches in `/data`, the persistent `var.btrfs`) survives
-//!   across image bumps. Refresh via `avocado vm reset-var` if you
-//!   actually want a clean slate.
+//! - `seed_only` — the `var` image. Fetched on first install and, on a
+//!   version bump, re-fetched and re-seeded. See below.
+//!
+//! ## Why a version bump resets `var`
+//!
+//! `seed_only` was written to preserve user state across image bumps,
+//! on the assumption that the in-VM agent would migrate the existing
+//! btrfs at boot. That
+//! migration does not exist yet, and `var` holds more than user state:
+//! the VM's own system extensions live in `/var/lib/avocado/images` and
+//! are merged onto `/usr` and `/etc` at boot. Carrying `var` across a
+//! bump therefore boots a new kernel and rootfs against the *old*
+//! userspace.
+//!
+//! Nothing catches that. Each extension's `extension-release` is
+//! synthesized at merge time carrying `ID=_any`, which is
+//! systemd-sysext's explicit "matches any OS" wildcard — so the
+//! mismatched extensions merge silently and the VM looks healthy.
+//!
+//! Until the guest can migrate in place, a version bump takes the new
+//! seed and discards the live disk. That costs the user their Docker
+//! volumes, SDK images and `/data` work, so they must re-run `avocado
+//! install` and `avocado build` — a tradeoff deliberately accepted over
+//! shipping a VM whose halves don't match. An unchanged `var` sha means
+//! no reset.
 //!
 //! Behaviour with a running VM: query lifecycle, stop it cleanly,
 //! perform the swap, restart with the same `start` options. The
@@ -33,13 +53,17 @@ use crate::utils::vm::channel::ChannelPointer;
 use crate::utils::vm::manifest::{Manifest, UpdatePolicy};
 use crate::utils::vm::staging::StagingDir;
 use crate::utils::vm::state::VmPaths;
-use crate::utils::vm_update_check::{check_for_vm_update, DEFAULT_BASE};
+use crate::utils::vm_update_check::{check_for_vm_update, VmUpdateStatus, DEFAULT_BASE};
 
 /// CLI surface — keep this in sync with the clap variant in main.rs.
 pub struct UpdateCommand {
     pub channel: Option<String>,
     pub check_only: bool,
     pub assume_yes: bool,
+    /// Authorizes the var reset when combined with `--yes`. Interactive
+    /// runs consent at the typed prompt instead, so this is only
+    /// consulted when `assume_yes` is set.
+    pub reset_var: bool,
     pub output: OutputFormat,
 }
 
@@ -73,14 +97,35 @@ impl UpdateCommand {
             .and_then(|m| m.version.as_deref())
             .map(|s| s.to_string());
 
-        // Channel poll (24h cached). Returns Some only when newer
-        // *and* CLI is compatible with min_cli_version.
-        let avail = check_for_vm_update(&channel_name, installed_version.as_deref()).await;
-
-        // Print the "what's available" summary so --check is useful
-        // even when nothing's new.
-        let Some(avail) = avail else {
-            return print_up_to_date(installed_version.as_deref(), &channel_name, self.output);
+        // Channel poll (24h cached).
+        let avail = match check_for_vm_update(&channel_name, installed_version.as_deref()).await {
+            VmUpdateStatus::Available(avail) => avail,
+            // A newer release exists that this CLI must not install.
+            // Fail rather than print-and-succeed: host applications map
+            // a non-zero `--check` exit to an error status carrying our
+            // stderr, so the actionable "run `avocado upgrade` first"
+            // message reaches their UI without them needing to
+            // understand a new field. In JSON mode a structured line
+            // goes out first so stdout parsers can tell "blocked on CLI
+            // version" from "check broke".
+            VmUpdateStatus::CliTooOld { message, remote } => {
+                if self.output.is_json() {
+                    crate::utils::output_format::emit_json_object(&json!({
+                        "channel": channel_name,
+                        "installed": installed_version,
+                        "remote": remote,
+                        "update_available": true,
+                        "cli_too_old": true,
+                        "message": message,
+                    }));
+                }
+                bail!(message)
+            }
+            // Print the "what's available" summary so --check is useful
+            // even when nothing's new.
+            VmUpdateStatus::NoUpdate => {
+                return print_up_to_date(installed_version.as_deref(), &channel_name, self.output)
+            }
         };
 
         if self.check_only {
@@ -113,11 +158,6 @@ impl UpdateCommand {
             )
         })?;
 
-        // Confirm with the user (unless --yes).
-        if !self.assume_yes {
-            confirm(&avail.pointer, installed_version.as_deref())?;
-        }
-
         // HTTP client — `connect_timeout` is bounded so a stalled DNS /
         // TCP handshake fails fast, but the overall request timeout is
         // unset because artifact downloads can run several minutes on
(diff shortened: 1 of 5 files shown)


src/commands/vm/update.rs
```diff
@@ -230,19 +270,40 @@ impl UpdateCommand {
...
-            serde_json::to_vec_pretty(&serde_json::from_str::<serde_json::Value>(
-                &http
-                    .get(&platform_entry.manifest_url)
-                    .send()
-                    .await?
-                    .error_for_status()?
-                    .text()
-                    .await?,
-            )?)?;
+            serde_json::to_vec_pretty(&serde_json::from_str::<serde_json::Value>(&manifest_raw)?)?;
```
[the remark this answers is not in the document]
lee-reinhardt: Added context to the re-parse (`re-parsing remote manifest for the install dir`). Note it is near-unreachable — the same bytes already parsed as a typed `Manifest` above — but the context is cheap insurance against drift.

src/commands/vm/update.rs
```diff
@@ -286,42 +349,46 @@ struct PlannedDownload {
...
+        let sha_differs = installed_sha != Some(art.sha256.as_str());
+        let wanted = match art.update_policy {
+            // On a first install, the file's presence is the only signal
+            // we have — there's no installed manifest to compare against,
+            // and a var image already in place is one we just fetched.
+            // On an update the sha decides, so an unchanged var image
+            // costs neither a ~450 MB download nor the user's state.
+            UpdatePolicy::SeedOnly => match installed {
+                Some(_) => sha_differs,
+                None => !install_dir.join(&art.file).exists(),
```
nicksinas, line 432: Dropping the presence check loses self-healing. If `install_dir/var.btrfs` is missing while the sha matches, it is never re-fetched: `vm update` reports "nothing to download", `verify_all` skips `seed_only` and `seed_var_disk` no-ops on a missing source, so `vm start` boots with no /var attached at all (qemu.rs only attaches when the file exists), and `vm reset` bails telling the user to run `vm update`.

Suggest `Some(_) => sha_differs || !install_dir.join(&art.file).exists()` — the sha gate still avoids the ~450 MB and the state loss in the normal case. Worth a test next to `update_leaves_an_unchanged_var_image_alone`.

src/commands/vm/update.rs
```diff
@@ -230,19 +270,42 @@ impl UpdateCommand {
...
+        // the VM is down. Deleting rather than copying is deliberate:
+        // `lifecycle::start` already re-seeds a missing var.btrfs from
+        // the artifact dir and then re-applies the configured size, so
+        // the boot path stays the single owner of both, and an update
+        // that dies here leaves no half-copied disk to mistake for
+        // state — the next start just seeds it.
+        if reseed_var {
+            let var = paths.var_disk();
+            if var.exists() {
+                std::fs::remove_file(&var)
```
nicksinas, line 335: `var_size` is never persisted — it's only a `vm start` flag, and `VmConfig` has no key for it — so the `var_size: None` restart below resolves to `DEFAULT_VAR_SIZE` (50G). Deleting the file is what loses the size: it used to be sticky because `grow_var_file` only ever grows. A user on `--var-size 200G` silently ends up at 50G.

Either capture `metadata(&var).len()` before the delete and pass it as `var_size`, or persist the requested size in `~/.avocado/vm/config.yaml` and have `lifecycle::start` read it. The PR description's "re-applies the configured size" only holds for the default.

src/commands/vm/update.rs
```diff
@@ -153,6 +178,21 @@ impl UpdateCommand {
             return Ok(());
         }
 
+        // Whether this run replaces the live var disk. Only on an update
+        // of an existing install: with no installed manifest there is no
+        // user state to destroy, and `seed_var_disk` on the next start
+        // handles a missing disk anyway. Gating on the planned download
+        // means an unchanged var sha leaves the disk alone.
+        let reseed_var = installed.is_some() && downloads.iter().any(|d| d.seed_only);
```
nicksinas: `installed.is_some() && any(seed_only)` doesn't imply there is state to lose. An install whose VM was never started has no `var.btrfs`, but still gets the full destructive prompt and still emits `var_reset` — making host apps drop cached install state and nag for `install`/`build` when nothing was reset. Add `paths.var_disk().exists()` to the condition.

Also: this is the only decision in the change that destroys user data, and the only one not unit-tested. `fn should_reseed_var(installed: Option<&Manifest>, downloads: &[PlannedDownload], var_exists: bool) -> bool` would test alongside `plan_downloads`.

src/commands/vm/update.rs
```diff
@@ -153,6 +178,21 @@ impl UpdateCommand {
...
+
+        // Confirm with the user (unless --yes). Deliberately after
+        // planning: whether this destroys the VM's state is the material
+        // fact about the operation, and we can't know it before reading
+        // the remote manifest.
+        if !self.assume_yes {
```
nicksinas, line 219: `-y` skips this entirely, so the flag's meaning changes from "skip a y/N on a non-destructive update" to "authorize wiping VM state". Existing scripts, CI and host apps on `avocado vm update -y` lose SDKs, Docker volumes and /data on the first release with a new var sha, having never opted in.

Suggest requiring a distinct `--reset-var` alongside `-y` when `reseed_var`, failing with a message that names the flag. At minimum, update the `--yes` help text in main.rs (currently "Skip the interactive confirmation prompt") and the release notes.

src/commands/vm/update.rs
```diff
@@ -73,14 +93,21 @@ impl UpdateCommand {
...
+        // Channel poll (24h cached).
+        let avail = match check_for_vm_update(&channel_name, installed_version.as_deref()).await {
+            VmUpdateStatus::Available(avail) => avail,
+            // A newer release exists that this CLI must not install.
+            // Fail rather than print-and-succeed: host applications map
+            // a non-zero `--check` exit to an error status carrying our
+            // stderr, so the actionable "run `avocado upgrade` first"
+            // message reaches their UI without them needing to
+            // understand a new field.
+            VmUpdateStatus::CliTooOld { message } => bail!(message),
```
nicksinas: Right that this shouldn't be swallowed, but under `--output json` it bails before any JSON is written. A host app polling `--check --output json` goes from `{"update_available": true, ...}` to empty stdout plus prose on stderr, and can't distinguish "blocked on CLI version" from "check failed".

Suggest emitting `{"update_available": true, "cli_too_old": true, "remote": ..., "message": ...}` first (mirroring `print_update_available`), then returning the error, so both the exit code and the structured output carry it.

src/commands/vm/update.rs
```diff
@@ -345,9 +412,37 @@ async fn is_vm_running() -> bool {
...
+        println!("update resets the VM's internal state:");
+        println!();
+        println!("  - installed SDKs and their container images");
+        println!("  - Docker volumes and build caches");
+        println!("  - anything resident in /data inside the VM");
+        println!();
+        println!("Your projects on the host are untouched, but each will need");
+        println!("`avocado install` and `avocado build` run again afterwards.");
+        println!();
+        print!("Type 'update' to confirm: ");
```
nicksinas, line 533: Under `--output json` without `--yes`, this prompt writes prose into the NDJSON stream and then blocks on stdin; with a non-TTY stdin the empty read surfaces as a bare "aborted by user". Since this is the path host applications drive, consider bailing when `output.is_json() && !assume_yes` with something like "refusing to prompt in --output json mode; re-run with --yes (this resets the VM's /var)".

nicksinas (commented):
Reviewed against a local checkout; `cargo test` is green. The direction is right and the docs carry their reasoning well. Six things I'd want resolved, inline. #1 is the one I'd call blocking; #2 contradicts a claim in the PR description.

src/commands/vm/update.rs
lee-reinhardt: Restored in 4184fa7 as `sha_differs || !exists`, with a test (`update_refetches_a_missing_seed_even_with_a_matching_sha`). This also forced the reseed decision off the planned downloads and onto the manifests: a seed re-fetched for presence at the same sha must not reset the live disk (`repairing_a_missing_seed_does_not_reseed_the_live_disk`).

src/commands/vm/update.rs
lee-reinhardt: Took the config route in 4184fa7: `runtime.var_size` in config.yaml, `vm start` resolves flag > config > default, and the update records the live disk size (rounded up to GiB; an explicit larger config wins) before deleting. Covers the stopped-VM start that capture-and-pass-through would miss. Old CLIs round-trip the key via the `extra` flatten. 4 tests on `record_var_size_floor` plus the config round-trip.

src/commands/vm/update.rs
lee-reinhardt: Extracted `should_reseed_var(installed, new_manifest, var_disk_exists)` in 4184fa7 with the exists gate and 3 tests. Keyed on the manifests rather than the planned downloads so the missing-seed repair (thread above) cannot read as a var change.

src/commands/vm/update.rs
lee-reinhardt: Added `--reset-var` in 4184fa7: a reseeding update under `--yes` fails naming the flag; interactive runs still consent at the typed prompt. Updated the `--yes` help text and the `vm update` subcommand doc (its "preserves the existing var partition" claim was stale too). The desktop will pass `--yes --reset-var` after its own rebuild dialog.

src/commands/vm/update.rs
lee-reinhardt: Done in 4184fa7: `--output json` emits `{channel, installed, remote, update_available: true, cli_too_old: true, message}` before the non-zero exit; `CliTooOld` now carries the remote version.

src/commands/vm/update.rs
lee-reinhardt: Done in 4184fa7: `--output json` without `--yes` fails with "refusing to prompt in --output json mode; re-run with --yes" (plus a `--reset-var` pointer when the update would reseed), before anything is staged.

mobileoverlord (approved):
Correct diagnosis and guards (sha-gated, first-install and never-started cases excluded), and merging as the base of #206, which replaces the disk deletion with guest-side runtime adoption — the pair was E2E-verified together with avocado-vm#3. Net shipped behavior is the sync, not the reset.

PR #198, opened 2026-08-12
commands/sbom: emit an SPDX 3.0 SBOM of the installed packages

hiagofranco:
commands/sbom: emit an SPDX 3.0 SBOM of the installed packages

Generate SPDX 3.0 SBOM json output from the installed RPM packages.

The package list is read from each installed sysroot's RPM database, so
it is the transitive closure the device actually holds rather than the
packages avocado.yaml declares. Generate one document with one root per
scope rather than one document per scope.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/src/commands/sbom/generate.rs
+++ b/src/commands/sbom/generate.rs
@@ -1,0 +1,2505 @@
+//! `avocado sbom` — emit an SPDX 3.0.1 SBOM of what this project installed.
+//!
+//! Asks RPM what is in every sysroot (the walk lives in `utils::sysroot_scan`)
+//! and writes one `software_Sbom` of type `deployed`.
+//!
+//! One document with one root per scope, rather than one document per scope,
+//! for a reason that is about correctness rather than tidiness: a package
+//! present in both `rootfs` and `initramfs` is one package in two places. Split
+//! across files it becomes two SPDX elements, and a scanner counting packages
+//! or matching CVEs counts the same exposure twice. Here it is one element with
+//! two `contains` relationships.
+
+use anyhow::{Context, Result};
+use sha2::{Digest, Sha256};
+use std::borrow::Cow;
+use std::collections::{BTreeMap, BTreeSet};
+use std::sync::Arc;
+
+use crate::utils::config::{ComposedConfig, Config};
+use crate::utils::lockfile::{LockFile, RepoSnapshot, RPM_SBOM_FIELDS, RPM_SBOM_FORMAT};
+use crate::utils::output::{print_info, print_success, OutputLevel};
+use crate::utils::output_format::{emit_json_object, JsonOutputGuard, OutputFormat};
+use crate::utils::sysroot_scan::{self, ScanRequest};
+use crate::utils::target::resolve_target_required;
+
+/// Version of the SPDX specification the emitted document declares.
+const SPEC_VERSION: &str = "3.0.1";
+
+/// JSON-LD context for that version.
+const CONTEXT: &str = "https://spdx.org/rdf/3.0.1/spdx-context.jsonld";
+
+/// Scopes that exist on the build host and ship nothing to a device. Auditing
+/// them is a legitimate thing to want — hence `--include-sdk` — but a document
+/// answering "what is on this device" must not list the cross toolchain by
+/// default: it is 290 `nativesdk-*` packages against the ~400 that ship.
+const BUILD_HOST_SCOPES: &[&str] = &["sdk", "target-sysroot"];
+
+/// One installed package, with the provenance SPDX asks for.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+struct Package {
+    name: String,
+    version: String,
+    release: String,
+    arch: String,
+    epoch: String,
+    license: String,
+    url: String,
+    packager: String,
+    sourcerpm: String,
+    sha256: String,
+    buildtime: String,
+    installtid: String,
+    summary: String,
+}
+
+impl Package {
+    /// Identity for de-duplication across scopes. Deliberately excludes the
+    /// scope: the same package in two sysroots is one package. Includes the
+    /// epoch, without which `1:2.39-r0.2` and `2.39-r0.2` — different packages
+    /// by rpm's own ordering — would collapse into one element.
+    fn key(&self) -> (&str, &str, &str, &str, &str) {
+        (
+            &self.name,
+            self.epoch(),
+            &self.version,
+            &self.release,
+            &self.arch,
+        )
+    }
+
+    /// The epoch as a comparable string, `""` when rpm reports none. rpm writes
+    /// `(none)` for an unset epoch, which is not the same as epoch 0 in the
+    /// tag but compares equal for every purpose here.
+    fn epoch(&self) -> &str {
+        present(&self.epoch).unwrap_or("")
+    }
+
+    /// `2.39-r0.2`, or `1:2.39-r0.2` when the package carries an epoch. This is
+    /// the version a human reads and the one rpm compares on.
+    fn evr(&self) -> String {
+        match self.epoch() {
+            "" => format!("{}-{}", self.version, self.release),
+            e => format!("{}:{}-{}", e, self.version, self.release),
+        }
+    }
+
+    /// The package URL a consumer matches on.
+    ///
+    /// Nothing in the SPDX shapes validates this field — a malformed one passes
+    /// validation and fails silently downstream — so the components are encoded
+    /// here rather than interpolated raw.
+    ///
+    /// The epoch travels as the `epoch` qualifier rather than inside the
+    /// version, which is what the purl spec reserves for it; qualifiers are
+    /// emitted in the sorted order the spec requires, so `arch` precedes it.
+    /// The `arch` qualifier is dropped when rpm has none to report rather than
+    /// carrying `(none)`. A qualifier is a claim a scanner matches on, and
+    /// `?arch=(none)` matches no purl anything else derives from the same rpm —
+    /// it is worse than the absent qualifier, which at least degrades to a
+    /// name/version match. `gpg-pubkey`, which dnf installs into an installroot
+    /// whenever it imports a repo signing key, is the row that has neither an
+    /// arch nor a source rpm.
+    fn purl(&self) -> String {
+        let mut purl = format!(
+            "pkg:rpm/avocado/{}@{}-{}",
+            purl_encode(&self.name),
+            purl_encode(&self.version),
+            purl_encode(&self.release),
+        );
+        let mut sep = '?';
+        if let Some(arch) = present(&self.arch) {
+            purl.push_str(&format!("{sep}arch={}", purl_encode(arch)));
+            sep = '&';
+        }
+        if !self.epoch().is_empty() {
+            purl.push_str(&format!("{sep}epoch={}", purl_encode(self.epoch())));
+        }
+        purl
+    }
+
+    /// `glibc-2.39+git0+ce65d944e3-r0.2.src.rpm` -> `glibc`.
+    ///
+    /// The last two dash-separated fields are VERSION and RELEASE, so the
+    /// recipe is everything before them. Split from the right rather than
+    /// matched, because the version carries dashes of its own in the
+    /// `2.39+git0+...` form Yocto writes for a git recipe.
+    ///
+    /// Worth having: it means a device SBOM can name the recipe behind a
+    /// package without being joined against the build's pkgdata.
+    /// `None` when rpm reports no source rpm: the fallback is the stem, and for
+    /// an unset tag the stem is the literal `(none)`.
+    fn recipe(&self) -> Option<&str> {
+        let stem = present(&self.sourcerpm)?
+            .strip_suffix(".src.rpm")
+            .unwrap_or_else(|| self.sourcerpm.trim());
+        match stem.rsplitn(3, '-').nth(2) {
+            Some(recipe) if !recipe.is_empty() => Some(recipe),
+            _ => present(stem),
+        }
+    }
+}
+
+/// A sysroot and the packages it holds that are its own.
+#[derive(Debug)]
+struct Scope {
+    name: String,
+    root: String,
+    packages: Vec<Package>,
+    failed: bool,
+    /// Rows rpm printed that could not be mapped onto fields. Carried so the
+    /// command can say a package is missing: this command refuses a scope it
+    /// could not read at all, and a row it could not read is the same claim in
+    /// miniature.
+    unreadable: usize,
+}
+
+/// Whether a scope's installroot was seeded with a copy of the rootfs RPM
+/// database, and so needs the seed subtracted before its packages can be read
+/// as its own content. See `parse_scopes` for what the seed is and why.
+fn is_seeded_scope(name: &str) -> bool {
+    name.starts_with("ext:") || name.starts_with("runtime:")
+}
+
+/// Tripwires on the seeded-scope subtraction, which is a heuristic and has been
+/// wrong before. An installroot seeded from the rootfs holds the whole base, so
+/// the last time the subtraction failed it did so silently, on every seeded
+/// scope at once, and still produced a plausible-looking SBOM.
+///
+/// Returns the lines to print rather than printing them, so the conditions can
+/// be asserted against a scope list — the caller sits past a container round
+/// trip, and a guard that cannot be reached by a test is a guard that can be
+/// inverted without anything noticing.
+fn seeding_warnings(scopes: &[Scope]) -> Vec<String> {
+    let base_count = scopes
+        .iter()
+        .find(|s| s.name == "rootfs")
+        .map_or(0, |s| s.packages.len());
+    let seeded: Vec<&str> = scopes
+        .iter()
+        .filter(|s| is_seeded_scope(&s.name) && !s.packages.is_empty())
+        .map(|s| s.name.as_str())
+        .collect();
+
+    let mut warnings = Vec::new();
+
+    // The subtraction runs only against rows the rootfs contributed, so a
+    // rootfs that contributed none — absent from the dump, or every row of it
+    // unreadable — means nothing was subtracted and every seeded scope still
+    // carries its whole copy of the base. That is the harmful direction, and
+    // the count tripwire below cannot catch it: it needs the rootfs count it no
+    // longer has. Said separately for that reason.
+    if base_count == 0 && !seeded.is_empty() {
+        warnings.push(format!(
+            "[WARN] The rootfs contributed no packages, so nothing was subtracted from the {} \
+             seeded scope(s): {}. Their installroots are seeded with a copy of the rootfs RPM \
+             database, so what they report is the base system plus their own content, not what \
+             the extension ships.",
+            seeded.len(),
+            seeded.join(", ")
+        ));
+    }
+
+    // A seeded scope reporting as many packages as the rootfs itself is the
+    // shape the subtraction failing takes.
+    if base_count > 1 {
+        let suspect: Vec<&str> = scopes
+            .iter()
+            .filter(|s| is_seeded_scope(&s.name) && s.packages.len() >= base_count)
+            .map(|s| s.name.as_str())
+            .collect();
+        if !suspect.is_empty() {
+            warnings.push(format!(
+                "[WARN] {} scope(s) report at least as many packages as the rootfs ({}): {}. \
+                 Either they really do install that much, or the rootfs database they were seeded \
+                 from could not be told apart from what they installed themselves — check before \
+                 treating their contents as shipped by the extension.",
+                suspect.len(),
+                base_count,
+                suspect.join(", ")
+            ));
+        }
+    }
+
+    warnings
+}
+
+/// Rewrite a Yocto license string as an SPDX license expression.
+///
+/// Yocto writes `&` and `|` where SPDX wants `AND` and `OR`, and uses a handful
+/// of identifiers the SPDX list either never carried or has since deprecated.
+/// Those become `LicenseRef-`, which is the escape hatch the spec provides for
+/// exactly this — emitting them verbatim would produce an expression no
+/// consumer can resolve.
+///
+/// An unset tag is `NOASSERTION`. rpm prints `(none)` for one, and without the
+/// [`present`] guard the parentheses below are read as structure: `(none)`
+/// tokenises to the expression `( none )`, which looks resolvable and is not.
+fn spdx_license(raw: &str) -> String {
+    let raw = match present(raw) {
+        Some(raw) => raw,
+        None => return "NOASSERTION".to_string(),
+    };
+
+    let mut out: Vec<String> = Vec::new();
+    let mut token = String::new();
+
+    let flush = |token: &mut String, out: &mut Vec<String>| {
+        let t = token.trim();
+        if !t.is_empty() {
+            out.push(non_spdx_id(t).to_string());
+        }
+        token.clear();
+    };
+
+    for c in raw.chars() {
+        match c {
+            '&' => {
+                flush(&mut token, &mut out);
+                out.push("AND".to_string());
+            }
+            '|' => {
+                flush(&mut token, &mut out);
+                out.push("OR".to_string());
+            }
+            '(' | ')' => {
+                flush(&mut token, &mut out);
+                out.push(c.to_string());
+            }
+            _ => token.push(c),
+        }
+    }
+    flush(&mut token, &mut out);
+
+    if out.is_empty() {
+        return "NOASSERTION".to_string();
+    }
+    out.join(" ")
+}
+
+/// Identifiers Yocto emits that the SPDX license list does not carry.
+///
+/// Kept as a table rather than a heuristic: guessing at a license identifier is
+/// worse than declaring it unresolvable, since a wrong one is indistinguishable
+/// from a right one downstream. Anything not listed is passed through, because
+/// deciding "not a real SPDX identifier" needs the license list itself and this
+/// command does not carry one — an unrecognised identifier is far more often a
+/// listed one this table has no opinion about than a Yocto-ism.
+///
+/// The entries are the Yocto-isms that reach a device image. `Proprietary` and
+/// the `Firmware-*` family matter most: they are what vendor blobs and BSP
+/// firmware declare, so they are exactly the packages whose licensing a reader
+/// came to the document to check, and emitting them bare produces an expression
+/// no consumer can resolve.
+fn non_spdx_id(id: &str) -> Cow<'_, str> {
+    // Yocto writes one of these per firmware blob — `Firmware-qcom`,
+    // `Firmware-amd-ucode`, and so on down the BSP. Listing them individually
+    // would be a table that goes stale with every new SoC, and the prefix is
+    // unambiguous: no SPDX-listed identifier begins with it.
+    if let Some(rest) = id.strip_prefix("Firmware-") {
+        return Cow::Owned(format!("LicenseRef-Firmware-{rest}"));
+    }
+
+    let mapped = match id {
+        "bzip2-1.0.4" => "LicenseRef-bzip2-1.0.4",
+        "PD" => "LicenseRef-PD",
+        // Yocto's marker for a recipe with no redistributable license. Common
+        // enough on a device build to matter, and emitting it bare would read
+        // as a listed SPDX identifier that no consumer can resolve.
+        "CLOSED" => "LicenseRef-CLOSED",
+        // What a vendor blob declares. Not on the SPDX list in any casing.
+        "Proprietary" => "LicenseRef-Proprietary",
+        "GPL-2.0-with-OpenSSL-exception" => "LicenseRef-GPL-2.0-with-OpenSSL-exception",
+        "GPL-3.0-with-GCC-exception" => "LicenseRef-GPL-3.0-with-GCC-exception",
+        other => other,
+    };
+    Cow::Borrowed(mapped)
+}
+
+/// Percent-encode the characters that carry structural meaning in a purl.
+///
+/// Only those: `:` `/` `@` `?` `#` and `%` itself separate a purl's type,
+/// namespace, name, version and qualifiers, so leaving one literal moves a
+/// component boundary and a parser reads a different package. Everything else
+/// stays as written — in particular `+`, which appears in 29 of the 273
+/// packages on a real qemuarm64 device (`libstdc++6`, and the `2.39+git0+...`
+/// form Yocto gives a git recipe). It is not a purl separator, and it sits
+/// ahead of the `?` where a query-string parser could not read it as a space.
+///
+/// Anything outside printable ASCII is encoded too, since a purl is a URI.
+fn purl_encode(component: &str) -> String {
+    let mut out = String::with_capacity(component.len());
+    for byte in component.bytes() {
+        match byte {
+            b':' | b'/' | b'@' | b'?' | b'#' | b'%' | b'&' | b'=' => {
+                out.push_str(&format!("%{byte:02X}"))
+            }
+            0x21..=0x7E => out.push(byte as char),
+            _ => out.push_str(&format!("%{byte:02X}")),
+        }
+    }
+    out
+}
+
+/// An id component that is safe to embed in an IRI *and* distinct for distinct
+/// inputs.
+///
+/// [`slug`] alone is not injective: it collapses every run of non-alphanumerics
+/// to one `-` and trims the ends, so `GPL-2.0` and `GPL-2.0+` both come out as
+/// `GPL-2.0`. Two graph nodes would then share an `spdxId` while carrying
+/// different content, and a consumer resolving that id gets whichever it
+/// happened to read last — a package reported under the wrong license.
+///
+/// A string the slug preserves exactly is used as-is, which keeps the common
+/// case readable (`busybox-1.36.1-r0.2.cortexa57`). Anything the slug altered
+/// gets a digest of the *original* appended, so the mapping is injective again
+/// without making every id unreadable.
+fn slug_id(s: &str) -> String {
+    let base = slug(s);
+    if base == s && !base.is_empty() {
+        return base;
+    }
+    let digest = Sha256::digest(s.as_bytes());
+    let suffix: String = digest[..4].iter().map(|b| format!("{b:02x}")).collect();
+    if base.is_empty() {
+        suffix
+    } else {
+        format!("{base}-{suffix}")
+    }
+}
+
+/// The document's `created` timestamp.
+///
+/// `SOURCE_DATE_EPOCH` wins when it is set and parses, so the same installed
+/// set produces the same bytes twice running. `namespace_digest` already holds
+/// every element id still for that case; a wall-clock `created` would undo it
+/// one field later, and a consumer diffing two SBOMs would see churn that is
+/// not there. It also lets anything that pins or signs the document reference
+/// it by content hash.
+///
+/// Unset or unparseable falls back to now, which is the honest answer when the
+/// caller has not named a reproducible one. Seconds since the epoch, UTC, as
+/// the reproducible-builds convention defines it.
+fn created_timestamp(source_date_epoch: Option<&str>) -> String {
+    const FORMAT: &str = "%Y-%m-%dT%H:%M:%SZ";
+    source_date_epoch
+        .and_then(|s| s.trim().parse::<i64>().ok())
+        .and_then(|secs| chrono::DateTime::from_timestamp(secs, 0))
+        .map(|t| t.format(FORMAT).to_string())
+        .unwrap_or_else(|| chrono::Utc::now().format(FORMAT).to_string())
+}
+
+/// Make a string safe to embed in an SPDX id, which must be an IRI.
+///
+/// Not injective — see [`slug_id`], which is what id construction must use.
+fn slug(s: &str) -> String {
+    let mut out = String::with_capacity(s.len());
+    let mut last_dash = true;
+    for c in s.chars() {
+        if c.is_ascii_alphanumeric() || c == '.' {
+            out.push(c);
+            last_dash = false;
+        } else if !last_dash {
+            out.push('-');
+            last_dash = true;
+        }
+    }
+    out.trim_matches('-').to_string()
+}
+
+/// rpm writes `(none)` rather than an empty string for an unset tag, and an
+/// SPDX property carrying the literal `(none)` is worse than an absent one.
+fn present(value: &str) -> Option<&str> {
+    match value.trim() {
+        "" | "(none)" => None,
+        v => Some(v),
+    }
+}
+
+/// Epoch seconds -> the timestamp format SPDX pins: no fractional part, no
+/// offset, `Z` only. Anything else is rejected by the spec's own pattern.
+fn spdx_time(epoch: &str) -> Option<String> {
+    let secs: i64 = epoch.trim().parse().ok()?;
+    let dt = chrono::DateTime::from_timestamp(secs, 0)?;
+    Some(dt.format("%Y-%m-%dT%H:%M:%SZ").to_string())
+}
+
+pub struct SbomCommand {
+    config_path: String,
+    /// The global `--runs-on`, carried only so the command can refuse it.
+    /// `run_in_container_capture` has no remote branch — unlike
+    /// `run_in_container`, which routes on `config.runs_on` — so honouring it
+    /// would take work in `utils::container`, not a field here.
+    runs_on: Option<String>,
+    target: Option<String>,
+    output_path: Option<String>,
+    include_sdk: bool,
+    verbose: bool,
+    container_args: Option<Vec<String>>,
+    output: OutputFormat,
+    sdk_arch: Option<String>,
+    composed_config: Option<Arc<ComposedConfig>>,
+}
+
+impl SbomCommand {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        config_path: String,
+        target: Option<String>,
+        output_path: Option<String>,
+        include_sdk: bool,
+        verbose: bool,
+        container_args: Option<Vec<String>>,
+        output: OutputFormat,
+    ) -> Self {
+        Self {
+            config_path,
+            target,
+            output_path,
+            include_sdk,
+            verbose,
+            container_args,
+            output,
+            runs_on: None,
+            sdk_arch: None,
+            composed_config: None,
+        }
+    }
+
+    /// Record the global `--runs-on` so `execute` can refuse it rather than
+    /// silently describing the local machine.
+    pub fn with_runs_on(mut self, runs_on: Option<String>) -> Self {
+        self.runs_on = runs_on;
+        self
+    }
+
+    pub fn with_sdk_arch(mut self, sdk_arch: Option<String>) -> Self {
+        self.sdk_arch = sdk_arch;
+        self
+    }
+
+    #[allow(dead_code)]
+    pub fn with_composed_config(mut self, config: Arc<ComposedConfig>) -> Self {
+        self.composed_config = Some(config);
+        self
+    }
+
+    pub async fn execute(&self) -> Result<()> {
+        // Refused rather than ignored. The container helper this command uses
+        // reads the local volume with no remote branch, so
+        // accepting it would describe this machine's sysroots in a document
+        // named for another host — and an SBOM is a claim about what is on a
+        // particular device.
+        if let Some(host) = &self.runs_on {
+            anyhow::bail!(
+                "--runs-on {host} is not supported by `sbom`: it would describe this machine's \
+                 sysroots as {host}'s. Run the command on that host."
+            );
+        }
+
+        let _json_guard = self.output.is_json().then(JsonOutputGuard::enable);
+
+        let composed = match &self.composed_config {
+            Some(cc) => Arc::clone(cc),
+            None => Arc::new(
+                Config::load_composed(&self.config_path, self.target.as_deref())
+                    .context("Failed to load composed config")?,
+            ),
+        };
+        let config = &composed.config;
+        let target = resolve_target_required(self.target.as_deref(), config)?;
+
+        // Progress goes to stdout (`utils::output::print_info` is a `println!`),
+        // and so does the document when no `-o` was given. Left on, `avocado
+        // sbom --verbose > sbom.json` writes `[INFO] Querying installed
+        // packages...` ahead of the JSON and the file will not parse. The
+        // document is the output that matters, so the commentary yields —
+        // announced on stderr, which is where a note about stdout belongs.
+        let scan_verbose = self.verbose && self.output_path.is_some();
+        if self.verbose && !scan_verbose {
+            eprintln!(
+                "[INFO] --verbose is suppressed while the document goes to stdout; it would be \
+                 written into the JSON. Pass -o <path> to see it."
+            );
+        }
+
+        let output = sysroot_scan::run_discovery(
+            config,
+            ScanRequest {
+                config_path: &self.config_path,
+                target: &target,
+                verbose: scan_verbose,
+                container_args: self.container_args.as_ref(),
+                sdk_arch: self.sdk_arch.clone(),
+                query_format: RPM_SBOM_FORMAT,
+            },
+        )
+        .await?;
+
+        let scopes = self.parse_scopes(&output.stdout);
+
+        // A scope whose database could not be read contributes no packages. An
+        // SBOM silently missing a sysroot is worse than no SBOM: it is a
+        // complete-looking inventory of an incomplete scan.
+        let failed: Vec<&str> = scopes
+            .iter()
+            .filter(|s| s.failed)
+            .map(|s| s.name.as_str())
+            .collect();
+        if !failed.is_empty() {
+            anyhow::bail!(
+                "Could not read the RPM database of {} scope(s): {}. Their packages would be \
+                 missing from the SBOM.{}",
+                failed.len(),
+                failed.join(", "),
+                sysroot_scan::stderr_tail(&output.stderr)
+            );
+        }
+
+        // Not fatal — one unreadable row out of four hundred is a worse reason
+        // to produce no SBOM than to produce one short a package. But it is
+        // said out loud, on stderr so it survives the document owning stdout,
+        // because the alternative is a document that looks complete and is
+        // not.
+        let unreadable: usize = scopes.iter().map(|s| s.unreadable).sum();
+        if unreadable > 0 {
+            let where_ = scopes
+                .iter()
+                .filter(|s| s.unreadable > 0)
+                .map(|s| format!("{} ({})", s.name, s.unreadable))
+                .collect::<Vec<_>>()
+                .join(", ");
+            eprintln!(
+                "[WARN] {unreadable} package row(s) could not be read and are missing from the \
+                 SBOM: {where_}. A tab inside an rpm tag shifts every field after it, so the row \
+                 is dropped rather than mapped onto the wrong columns."
+            );
+        }
+
+        for warning in seeding_warnings(&scopes) {
+            eprintln!("{warning}");
+        }
+
+        if scopes.iter().all(|s| s.packages.is_empty()) {
+            anyhow::bail!(
+                "No installed package was found in any sysroot for target '{target}'. Run \
+                 `avocado install` first, and note that this command reads the state volume of \
+                 the current directory, not of --config's directory."
+            );
+        }
+
+        // Provenance, and never a reason to fail: the document is built from
+        // the RPM databases, and the lockfile only says which feed filled them.
+        // A missing or unreadable lockfile costs the reader that pointer, which
+        // is worth strictly less than the inventory itself.
+        let src_dir = config
+            .get_resolved_src_dir(&self.config_path)
+            .unwrap_or_else(|| {
+                std::path::Path::new(&self.config_path)
+                    .parent()
+                    .unwrap_or(std::path::Path::new("."))
+                    .to_path_buf()
+            });
+        let snapshot = LockFile::load(&src_dir)
+            .ok()
+            .and_then(|lock| lock.get_repo_snapshot(&target).cloned());
+
+        let doc = self.build_document(&scopes, &target, snapshot.as_ref());
+
+        match &self.output_path {
+            Some(path) => {
+                std::fs::write(path, serde_json::to_string_pretty(&doc)?)
+                    .with_context(|| format!("Failed to write SBOM to '{path}'"))?;
+                self.print_summary(&scopes, Some(path));
+            }
+            None if self.output.is_json() => emit_json_object(&doc),
+            None => {
+                println!("{}", serde_json::to_string_pretty(&doc)?);
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Map the raw dump onto packages, dropping what a scope only sees because
+    /// its installroot was seeded from the rootfs.
+    fn parse_scopes(&self, output: &str) -> Vec<Scope> {
+        let dumps = sysroot_scan::parse_scopes(output);
+
+        let to_packages = |rows: &[Vec<String>]| -> Vec<Package> {
+            rows.iter()
+                // A row with any other field count holds a tab inside one of
+                // the fields, or is a truncated read. Either way the offsets
+                // below have shifted, and mapping it anyway would put a
+                // description where a checksum belongs — so it is dropped and
+                // counted. Rejoining the tail into SUMMARY would be right only
+                // if SUMMARY were the field that held the tab; when it was
+                // LICENSE or PACKAGER instead, that recovers a package with a
+                // plausible and wrong checksum, which is worse than losing it.
+                .filter(|r| r.len() == RPM_SBOM_FIELDS)
+                .map(|r| Package {
+                    name: r[0].clone(),
+                    version: r[1].clone(),
+                    release: r[2].clone(),
+                    arch: r[3].clone(),
+                    epoch: r[4].clone(),
+                    license: r[5].clone(),
+                    url: r[6].clone(),
+                    packager: r[7].clone(),
+                    sourcerpm: r[8].clone(),
+                    sha256: r[9].clone(),
+                    buildtime: r[10].clone(),
+                    installtid: r[11].clone(),
+                    summary: r[12].clone(),
+                })
+                .collect()
+        };
+
+        // An extension's installroot is seeded with a copy of the rootfs RPM
+        // database so dnf can resolve against the base without reinstalling it.
+        // Those rows are not what the extension ships, and a `contains` for
+        // them would claim the extension carries the whole base system.
+        //
+        // The seed is taken once, when the installroot is first created
+        // (`ext/install.rs` runs the `cp -rf .../rootfs/var/lib/rpm` setup only
+        // `if !sysroot_exists`, and `runtime/install.rs` likewise), so it goes
+        // stale as soon as the rootfs moves on.
+        //
+        // What separates the two is that rpm records an install *transaction*,
+        // not a per-package timestamp: every package of one `dnf install`
+        // shares one `INSTALLTID`. The seed is a byte copy of the rootfs
+        // database, so it arrives as whole transactions, and everything the
+        // extension installs afterwards forms transactions of its own. A
+        // transaction is therefore the seed exactly when every package in it
+        // is one the rootfs also holds.
+        //
+        // Two rules this replaces, both wrong on real data:
+        //
+        //   * Subtracting by NVRA drops a package the extension genuinely
+        //     installed whenever the rootfs holds the same version — on a real
+        //     project, 57 of the 113 packages a runtime installed, because
+        //     curl's and vim's dependency closure overlaps the base.
+        //
+        //   * Matching a row against the rootfs by transaction id assumes the
+        //     rootfs still carries the ids the seed was copied from. `install
+        //     --force` reinstalls the rootfs in one new transaction, which
+        //     replaces all of them at once — so nothing matched, and every
+        //     extension reported the whole 139-package base system as its own
+        //     content. `--output json` turns `--force` on by itself, so that
+        //     was the default in CI.
+        //
+        // Matching whole transactions survives both: the seed keeps its own
+        // ids whatever the rootfs does, and a transaction that installed
+        // anything the rootfs lacks is kept entire, overlap included.
+        //
+        // Membership is by name and architecture, not by version. A rootfs
+        // that upgrades one package leaves the extension holding a stale copy
+        // of it; on version, that one row would make its whole transaction —
+        // the entire seeded base — look like the extension's own content, and
+        // a single upgraded package would put 139 packages back into every
+        // extension. The name survives an upgrade, so the seed stays
+        // recognisable.
+        //
+        // What this cannot see: a transaction in which the extension installed
+        // *only* packages the rootfs already carries by name. It reads as
+        // seed, and the extension loses its `contains` for them. That is the
+        // benign direction — the packages are still in the document, under the
+        // rootfs — and it needs an extension that adds nothing the base lacks,
+        // which is an extension with no reason to exist. The opposite error
+        // hands every extension the whole base system.
+        let base_names: BTreeSet<(String, String)> = dumps
+            .iter()
+            .find(|d| d.scope == "rootfs")
+            .map(|d| {
+                to_packages(&d.rows)
+                    .into_iter()
+                    .map(|p| (p.name, p.arch))
+                    .collect()
+            })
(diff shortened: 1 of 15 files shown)


yoctopidg3 (approved):
This looks good. One thing maybe for the future is a --validate using https://tools.spdx.org/app/validate/

hiagofranco:
> This looks good. One thing maybe for the future is a --validate using https://tools.spdx.org/app/validate/

I played with their API, it works, however it keeps the SBOM json-ld file public for 10 days in their servers. So instead of doing that without the user consent, I just documented this and added as a 'note' output to suggest the user to upload himself, knowing it will be there for 10 days.

hiagofranco:
Rebased, checks are now passing. Fixed an issues where 'install -f' would fail after the second run trying to generate the sbom output. This is now ready for reviews.

mobileoverlord:
Two things — one standalone, one an interaction with another open PR.

## The document isn't byte-stable

`namespace_digest` is careful about this and says so:

> Derived from the content rather than randomly so the document stays byte-stable: the same installed set must produce the same bytes twice running, or a consumer diffing two SBOMs sees churn that is not there.

But `build_document` then sets

```rust
let created = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string();
```

so two runs over an unchanged sysroot still produce different bytes — exactly the churn the comment is guarding against. The namespace work does its job and `created` undoes it a few lines later.

Honoring `SOURCE_DATE_EPOCH` when set, falling back to `now()`, would close it. That also makes the document referenceable by content hash from anything that pins or signs it, which is worth having before consumers start depending on the current behavior.

## Interaction with #193

#193 changes the rpmdb seed source from the rootfs to a dependency's sysroot. That breaks the assumption behind `base_names` here: `find(|d| d.scope == "rootfs")` stops being the extension's actual base, so a dependency's transactions contain packages the rootfs lacks, fail the `all_in_base` test, and read as extension content. It compounds up the chain — #193 documents the composition as `rootfs ∪ base ∪ mid`.

Concretely, #193 measures `kiosk-a` at 3.5 MB because weston isn't in the image, while this would emit a document saying `kiosk-a` contains weston — over-reporting by exactly the content #193 deduplicates. The comment here already names that as the direction to avoid: "the opposite error hands every extension the whole base system."

Short version of the fix: have the seed step record its transaction ids at `cp -rf` time — right after the copy, before anything is installed, every row in the destination db is seed by definition — and drop exactly those here. That replaces `base_names` and the transaction grouping entirely, and it also closes the two failure modes documented above it (NVRA subtraction dropping 57 of 113 packages; `--force` replacing the rootfs's ids) plus the acknowledged blind spot where a transaction installs only packages the rootfs already carries by name. All of those are the same root cause: `cp -rf` of an rpmdb leaves no record of what arrived in the copy, so everything downstream has to infer it.

Whichever of these two lands second should carry the change.

hiagofranco:
> Two things — one standalone, one an interaction with another open PR.
> ## The document isn't byte-stable
> 
> `namespace_digest` is careful about this and says so:
> 
> > Derived from the content rather than randomly so the document stays byte-stable: the same installed set must produce the same bytes twice running, or a consumer diffing two SBOMs sees churn that is not there.
> 
> But `build_document` then sets
> 
> ```rust
> let created = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string();
> ```
> 
> so two runs over an unchanged sysroot still produce different bytes — exactly the churn the comment is guarding against. The namespace work does its job and `created` undoes it a few lines later.
> 
> Honoring `SOURCE_DATE_EPOCH` when set, falling back to `now()`, would close it. That also makes the document referenceable by content hash from anything that pins or signs it, which is worth having before consumers start depending on the current behavior.
> ## Interaction with #193
> 
> #193 changes the rpmdb seed source from the rootfs to a dependency's sysroot. That breaks the assumption behind `base_names` here: `find(|d| d.scope == "rootfs")` stops being the extension's actual base, so a dependency's transactions contain packages the rootfs lacks, fail the `all_in_base` test, and read as extension content. It compounds up the chain — #193 documents the composition as `rootfs ∪ base ∪ mid`.
> 
> Concretely, #193 measures `kiosk-a` at 3.5 MB because weston isn't in the image, while this would emit a document saying `kiosk-a` contains weston — over-reporting by exactly the content #193 deduplicates. The comment here already names that as the direction to avoid: "the opposite error hands every extension the whole base system."
> 
> Short version of the fix: have the seed step record its transaction ids at `cp -rf` time — right after the copy, before anything is installed, every row in the destination db is seed by definition — and drop exactly those here. That replaces `base_names` and the transaction grouping entirely, and it also closes the two failure modes documented above it (NVRA subtraction dropping 57 of 113 packages; `--force` replacing the rootfs's ids) plus the acknowledged blind spot where a transaction installs only packages the rootfs already carries by name. All of those are the same root cause: `cp -rf` of an rpmdb leaves no record of what arrived in the copy, so everything downstream has to infer it.
> 
> Whichever of these two lands second should carry the change.

Thanks, fixed the bug and left the second one open, as you mentioned, whatever PR is merged first should fix it. Please take a look.

jetm (commented):
Reviewed at 1faa682. The correctness gate I most wanted to check passes: enumeration
is `rpm -qa --root="$root"` per sysroot, so this is the full installed closure and not
avocado.yaml's declared `deps_map`. Reading the declared list would have produced ~15
packages instead of ~400 and reported clean, which is the failure mode that makes an
SBOM worse than none. avocado.yaml is used only to find which extension dirs are
scannable. `cargo fmt`, `clippy -D warnings` and the 136 integration tests are clean.

The SPDX 3.0 shapes look right to me: `software_Package` with `software_packageVersion`
/ `software_packageUrl` / `software_attributionText`, `Hash` under `verifiedUsing`,
`simplelicensing_LicenseExpression` referenced by id, and `Relationship` with
from/relationshipType/to. Sharing license and supplier elements by id rather than
repeating them per package is the right call, and `slug_id`'s injectivity argument
(generate.rs:262) is the kind of thing that is easy to get wrong and expensive to debug
later. The `hasDeclaredLicense` relationship id derives from the package id's last
segment, which I checked for collisions: `emitted` is keyed on
(name, epoch, version, release, arch) so each package is emitted once globally, and
slug_id is injective, so distinct packages cannot share a relationship id.

## Requested change: test_output_path_is_optional is flaky, and the mechanism is not quite what it looks like

It fails in roughly two of three full `cargo test` runs and passes 10/10 when the sbom
tests run alone. `tests/integration.rs:5` declares `mod interpolation;`, so
tests/interpolation.rs and tests/commands/avocado/sbom.rs compile into one binary and
share one process. tests/interpolation.rs sets `AVOCADO_TARGET=riscv64-unknown-linux-gnu`
process-globally, `resolve_target_required` (src/utils/target.rs:82) takes
CLI arg then AVOCADO_TARGET then config, so target resolution succeeds and the assertion
on "No target architecture specified" fails. The observed failure names
riscv64-unknown-linux-gnu exactly.

Two details change what the fix has to be, so they are worth stating.

First, interpolation.rs does clean up after itself: it calls
`env::remove_var("AVOCADO_TARGET")` nine times and every test that sets the variable is
`#[serial]`. But `#[serial]` only serialises against other `#[serial]` tests, and none of
the sbom tests carry it. So the sbom tests run on the parallel pool alongside a serial
test that has the variable live, and they observe the window between set and remove.
That is why this is intermittent rather than deterministic.

Second, `tests/common/mod.rs` does not filter the variable out, it forwards it. The loop
maps `AVOCADO_TARGET` to `TEST_AVOCADO_TARGET` in the child env, and the bash wrapper
above it then does `unset AVOCADO_TARGET` followed by
`export AVOCADO_TARGET="$TEST_AVOCADO_TARGET"` when that is non-empty. The comment reads
"so the script can decide whether to use it", and the script's decision is to use it. So
spawning through that mechanism is what carries the leaked value into the CLI, and
"spawn with AVOCADO_TARGET removed" via that path will not do it.

Marking the sbom tests `#[serial]` is the smallest fix and matches what every other test
touching this variable already does. A `run_cli` variant that declines to forward it
would also work if you would rather not serialise.

## Requested change: the seeded-scope tripwire cannot fire in the case it exists for

The subtraction at generate.rs:598-699 is well argued and I am not asking you to change
it. Grouping by INSTALLTID and treating a transaction as seed exactly when every package
in it is in the rootfs by name and arch survives both failure modes the comment
documents, and the writeup of why NVRA subtraction and per-row tid matching were wrong is
genuinely useful.

The problem is the interaction between two guards. The subtraction runs only
`if seeded && !base_names.is_empty()` (generate.rs:671), and `base_names` is built from
the rootfs dump's readable rows. The tripwire that catches a silent subtraction failure
runs only when `rootfs.packages.len() > 1` (generate.rs:507).

So when the rootfs scan fails, or when every rootfs row is unreadable because a tag
carried a tab, `base_names` is empty, no subtraction happens, and every seeded scope
keeps its full copy of the base. That is the direction the comment calls out as the
harmful one, "hands every extension the whole base system". And because the rootfs
contributed zero packages, `rootfs.packages.len() > 1` is false and the tripwire that was
written for exactly this is disabled by the same condition. The command does not bail,
because the extensions have packages, so it emits a plausible-looking document.

A failed-scope warning does fire, so this is not literally silent, but it says the rootfs
is missing rather than that every extension's contents are now unsubtracted. Making the
empty-`base_names` case warn in its own right, or refusing to emit seeded scopes at all
when the rootfs contributed nothing, would close it.

## PURL namespace mismatch with the Yocto-side SBOM

This emits `pkg:rpm/avocado/{name}@{version}-{release}?arch={arch}` (generate.rs:98). The
Yocto build's SPDX emits `pkg:yocto/{layer}/{recipe}@{version}`. Different purl type and
different granularity, binary package against recipe, so a correlation service fed both
cannot join them on the purl.

You already capture SOURCERPM and derive the recipe from it (generate.rs:119), which is
the bridge. My only suggestion is that it currently lands in
`software_attributionText` as the free-text string "recipe: glibc", so joining on it
means parsing prose. SPDX 3.0's `ExternalIdentifier` with
`externalIdentifierType: packageUrl` is repeatable, so the document could carry the
yocto-form purl alongside the rpm one and make the join structural. Worth at least
naming the mapping explicitly somewhere a consumer will find it.

## Not a finding: the inventory/assessment boundary is drawn in the right place

The document carries no CPE identifiers and no VEX assessments, and that is correct and
deliberate rather than a gap. This command emits inventory; assessment is a separate
surface, tracked in ENG-2325 and ENG-2369. Emitting CPEs here would actively invite the
confusion it might look like it prevents, because a document carrying vulnerability
identifiers reads as though it carried an assessment.

`test_the_command_offers_no_way_to_upload_the_document` pins the neighbouring concern,
and pins it harder than I would have thought to ask for: `--validate`,
`--validator-url` and `--upload` are all asserted absent from help, on the stated
grounds that the service running the reference SPDX tools keeps uploads for about ten
days and serves them back unauthenticated, while this document names every package and
version on the target. That is the right instinct and worth saying so.

One point sits outside this PR rather than in it: a clean `avocado sbom` output must
never be read as a clean vulnerability assessment. That belongs in the M5 docs
limitation note, not in this diff, so it is not a request here.

Nothing above blocks the shape of the work, which is solid. The two requested changes are
the flaky test and the tripwire gap.

hiagofranco:
1 and 2 are fixed, "Flaky test_output_path_is_optional" and "Tripwire gap".

3 was not done, since adding `ExternalIdentifier` with yocto-form purl means we will guess the layer name, which `SOURCERPM` does not carry (`pkg:yocto/{layer}/{recipe}` needs the layer). Emitting a purl with a wrong or invented namespace is worse than the current text `recipe: glibc`.

tests/commands/avocado/sbom.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::common;
+use serial_test::serial;
+
+/// The message target resolution produces, which is as far as a run gets in a
+/// workspace whose config names no default target. Reaching it means the flags
+/// parsed and the config loaded — everything this file can observe.
+const REACHED_TARGET_RESOLUTION: &str = "No target architecture specified";
+
+///
+/// Every caller is `#[serial]`. tests/interpolation.rs compiles into the same
```
jetm: **The `#[serial]` fix skips the one assertion the leak actually disarms**

The four `assert_parses` callers got `#[serial]`, but `test_runs_on_is_refused_rather_than_ignored` (fn at 95, assertion at 108) did not — and it is the only consumer of `REACHED_TARGET_RESOLUTION` whose assertion is *negative*. A leaked `AVOCADO_TARGET` makes the other four flake, which is visible; it makes this one pass by construction, which is not.

Verified by running the built binary: with `AVOCADO_TARGET` set, `sbom` never emits "No target architecture specified" (`grep -c` → 0), so `!stderr.contains(...)` at 108 is true regardless of ordering.

Concrete regression it would wave through: move the `--runs-on` refusal to *after* target resolution and all three of its asserts still pass — so the sole ordering guarantee rubber-stamps the bug it exists to catch. `git show 1faa682:` has no `#[serial]` anywhere in this file, so which tests got it is this increment's choice rather than inherited.

One line at 95, and the doc claim here at 22 wants widening past `assert_parses` callers.

src/commands/sbom/generate.rs
```diff
@@ -0,0 +1,2279 @@
...
+            .map(|s| s.name.as_str())
+            .collect();
+
+        // The subtraction runs only against rows the rootfs contributed, so a
+        // rootfs that contributed none — absent from the dump, or every row of
+        // it unreadable — means nothing was subtracted and every seeded scope
+        // still carries its whole copy of the base. That is the harmful
+        // direction, and the count tripwire below cannot catch it: it needs the
+        // rootfs count it no longer has. Said separately for that reason.
+        if base_count == 0 && !seeded.is_empty() {
```
jetm: **Both new tripwire predicates invert with the whole suite still green**

Flipping `base_count == 0` → `!= 0` and `base_count > 1` → `> 0` leaves `cargo test --lib` at 1431 passed / 0 failed. The repo CLAUDE.md is explicit here: "Every test must fail when the implementation is wrong (flip the logic and check)."

Both branches sit inside `pub async fn execute` past `run_discovery(...).await`, so nothing unit-testable can reach them — the guards are unverifiable where they stand, not merely unverified.

Extracting `fn seeding_warnings(scopes: &[Scope]) -> Vec<String>` makes both assertable against the fixtures the test module already builds at 28 call sites, without touching the async path.

jetm (commented):
Re-reviewed the increment since `1faa682` — 2 blocking findings inline, both mutation-proven rather than argued.

The refactor itself is right: `is_seeded_scope` is behaviour-identical at both call sites, the `base_count > 1` re-gate is a pure hoist of a loop-invariant term with no lost coverage across the old-vs-new truth table, and the SPDX document is provably unaffected since both branches only `eprintln!` and warnings go to stderr. The `#[serial]` fix addresses a real leak — `interpolation.rs` is `mod`-included into the integration binary and all four of its `AVOCADO_TARGET` setters are already serial. `cargo fmt`, `clippy -D warnings`, `cargo test --lib sbom` (34) and the sbom integration tests (10) all pass at `48c3cd7`.

Five advisory notes withheld to keep the blocking ones readable — happy to post them if useful.

tests/commands/avocado/sbom.rs
hiagofranco: Fixed, please take a look

src/commands/sbom/generate.rs
hiagofranco: Fixed, please take a look

src/utils/vm/route.rs
```diff
@@ -93,17 +93,16 @@ pub async fn ensure_routed_for_process(
...
                  Set AVOCADO_VM_DIR or run `avocado vm start --vm-source <dir>` to enable VM routing.",
-                OutputLevel::Normal,
             );
             return Ok(RoutingMode::OptedOut);
         };
-        print_info(
-            &format!("Starting avocado-vm from {}…", vm_source.display()),
-            OutputLevel::Normal,
-        );
+        print_info_stderr(&format!(
```
jetm, line 102: **VM auto-start still writes warnings to stdout**

This increment moves `route.rs`'s own notices to stderr, but the auto-start it kicks off four lines later (`lifecycle::start`, route.rs:106) still prints to stdout via `print_warning`/`print_info` at lifecycle.rs:293, :307, :323, :578 and :1000. So the exact failure this change's own rationale names is still reachable on the same run.

Failure path: `AVOCADO_VM_DIR` set, no VM running, user runs `avocado sbom > sbom.json`. `Commands::Sbom` is in `needs_vm_routing` (main.rs:2048), so auto-start proceeds; any of the three documented non-fatal hiccups fires (btrfs resize, network config apply, docker socket forward) and `[WARNING] btrfs resize on /var failed: ...` is `println!`'d to stdout ahead of the SPDX document. `sbom.json` is then not parseable JSON - precisely the outcome output.rs:146-152 gives as the reason for this change. `tui_is_active()` is false pre-dispatch, so nothing suppresses it.

src/utils/output.rs
```diff
@@ -141,6 +141,30 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// there, and a `[WARNING]` line ahead of it leaves a file no consumer can
+/// parse. These notices describe the environment rather than the result, so
+/// stderr is where they belong whichever command comes next.
+///
+/// Printed unconditionally, unlike the stdout printers. They suppress under
+/// [`tui_is_active`] — which is also true for `--json` — but the reason there is
+/// that prose would interleave with a stream stdout carries. On stderr it
+/// cannot, and a routing notice that vanishes under `--json` is a run whose
+/// container went somewhere the user was never told about.
+pub fn print_warning_stderr(message: &str) {
```
jetm, line 221: **New stderr printers bypass `print_notice_above`**

`print_warning_stderr`/`print_info_stderr` are a fourth open-coding of notice routing, added 60 lines below the comment at output.rs:92-99 that declares `print_notice_above` "the one implementation" after the policy "had been open-coded three times".

`print_notice_above(&format!("\x1b[93m[WARNING]\x1b[0m {msg}"), |l| eprintln!("{l}"))` is a behaviorally identical drop-in that also keeps renderer-awareness. All three arms of `print_notice_above` land on stderr with an `eprintln!` closure - `warning_sink(false, false)` -> `Stdout` -> closure, `print_above` uses `eprintln!` at renderer.rs:261, and both queue drains use `writeln!(stderr)` at renderer.rs:336 and :467. container.rs:406 and runtime/build.rs:821 already use that closure form.

Why it matters beyond tidiness: the current form is correct only while `ensure_routed_for_process` stays pre-dispatch. Both functions are `pub` and carry no pre-dispatch-only contract, so the first caller under a live `TaskRenderer` (the `TuiGuard` pattern at ext/build.rs:147, ext/image.rs:146, ext/install.rs:148) writes a raw `eprintln!` into the display region the renderer owns, and the task list paints over it on the next frame - silently.

src/utils/output.rs
```diff
@@ -141,6 +141,30 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// notice emitted before the command even dispatches, because the command that
+/// follows may own stdout: `avocado sbom > sbom.json` writes an SPDX document
+/// there, and a `[WARNING]` line ahead of it leaves a file no consumer can
+/// parse. These notices describe the environment rather than the result, so
+/// stderr is where they belong whichever command comes next.
+///
+/// Printed unconditionally, unlike the stdout printers. They suppress under
+/// [`tui_is_active`] — which is also true for `--json` — but the reason there is
+/// that prose would interleave with a stream stdout carries. On stderr it
+/// cannot, and a routing notice that vanishes under `--json` is a run whose
```
jetm: **"On stderr it cannot interleave" is false in this repo**

The doc comment's load-bearing justification for dropping `--json` suppression is that stderr cannot interleave with a stdout stream. main.rs:4297-4304 contradicts that in this same binary: it records that "the avocado-desktop CLI runner merges stdout + stderr in causal order" and that a stderr banner "ended up feeding the banner into a JSON parser, which then refused the entire payload". It suppresses the upgrade banner under `--json` for exactly that reason, calling silent-when-machine-readable "the safer contract".

Failure path: avocado-desktop runs `avocado sbom --output json` through that merging runner on a mac with no `AVOCADO_VM_DIR`. route.rs:96 emits its `[WARNING]` on stderr, the runner merges it into the payload, and the desktop app rejects the whole response.

Separately, unlike `print_warning_above` (output.rs:82-86) the new function emits no NDJSON event, so a consumer reading only the NDJSON stream still learns nothing about the fallback - the blindness the comment says it is fixing.

jetm (commented):
Re-reviewed the amend since 48c3cd7 (same base, so this is the 581-line increment only, not the whole PR). Three blocking findings inline, all on the stdout/stderr routing change rather than the SBOM generation itself.

Five advisory notes were withheld rather than appended here, so the blocking ones stay readable.

src/utils/vm/route.rs
hiagofranco: Pushed new version, please take a look

src/utils/output.rs
hiagofranco: Pushed new version, please take a look

src/utils/output.rs
hiagofranco: Pushed new version, please take a look

src/utils/output.rs
```diff
@@ -141,6 +141,46 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// [`tui_is_active`] — which is also true for `--json` — and a routing notice
+/// that vanishes is a run whose container went somewhere the user was never
+/// told about. Stderr is not a free pass on its own: `main`'s upgrade banner
+/// had to be suppressed under `--json` because the avocado-desktop CLI runner
+/// merges stdout and stderr in causal order and fed the banner to a JSON
+/// parser. So the warning also emits an NDJSON event, the same pairing
+/// [`print_warning_above`] uses: the stream consumer gets a parseable record
+/// instead of prose it has to tolerate.
+pub fn print_warning_stderr(message: &str) {
+    if crate::utils::output_format::is_json_output_active() {
```
jetm: **The json guard is never true at any caller of this function**

The doc added directly above claims "the warning also emits an NDJSON event" so a merged-stream consumer still gets a parseable record. `is_json_output_active()` is false at every one of the ten call sites, so that branch is dead and the prose the comment says must not reach the parser is emitted unconditionally.

The reason is an ordering one: VM routing runs pre-dispatch, and every `JsonOutputGuard::enable` site runs post-dispatch. All six enable sites (main.rs:1972 via `run_with_json_lifecycle`, sbom/generate.rs:500, connect init/deploy/upload, init.rs:922) are inside command `execute()`, strictly after `ensure_routed_for_process` at main.rs:2119. `lifecycle::start`'s other callers (vm start.rs:94, reset.rs:111, update.rs:277) enable no guard at all.

Failure path: macOS host, VM not running, avocado-desktop runs `avocado sbom --output json` (`Commands::Sbom` is in `needs_vm_routing`, main.rs:2047). `ensure_routed_for_process` fires route.rs:96 or :102 before sbom/generate.rs:500 enables the guard, so the `[WARNING] avocado-vm not running...` goes to stderr with no event. The desktop runner merges both streams in causal order into a JSON parser, which refuses the payload - the identical failure main.rs:4297-4304 documents for the upgrade banner.

Worth noting main.rs already solves exactly this: it scans `std::env::args()` for `--output json`, which is the predicate that works before dispatch. This uses the one that does not.

Second edge on the same line: if the branch ever did fire under a single-shot `--output json` command, `sbom` emits one object via `emit_json_object` (generate.rs:615), so prepending an NDJSON warning line would break the one-object contract.

tests/no_stdout_on_the_vm_path.rs
```diff
@@ -0,0 +1,84 @@
...
+//! If this fails: use `print_warning_stderr` / `print_info_stderr`. They route
+//! through `print_notice_above`, so an active renderer still gets the notice
+//! above its task list instead of having it painted over.
+
+use std::fs;
+use std::path::{Path, PathBuf};
+
+/// Spellings that put a line on stdout. `print_plain` and `print_debug` are
+/// absent on purpose — both are already `eprintln!`.
+const NEEDLES: [&str; 4] = [
```
jetm: **NEEDLES omits `print_warning_above`, the sibling stdout emitter**

The module doc one screen up claims "nothing under `src/utils/vm` writes to stdout"; the check enforces something narrower. The needle set misses `print_warning_above(` (output.rs:88, whose plain arm is `println!` and so stdout on the ordinary no-renderer no-json path), bare `print!(`, and a direct stdout handle write.

Verified rather than inferred: I appended `print_warning_above("x")`, `print!("y")` and a `write_all(&mut std::io::stdout(), b"z")` to `src/utils/vm/state.rs` and the guard still passed.

The live path to that mattering is short. `print_warning_stderr` is now `print_warning_above` with one token changed - identical bodies, identical duplicated event block, differing only in the plain closure. A "dedupe these two" refactor collapsing them puts the VM path back on stdout, `avocado sbom > sbom.json` is corrupt again, and every test stays green, because `warning_sink`'s unit tests assert the enum rather than the stream.

One note when fixing: adding `print!(` needs `eprint!` blanked alongside `eprintln!`, or every `eprint!` call site becomes a false positive.

Credit where due - the guard is not vacuous. Reverting lifecycle.rs to 4569921 turns it red and it names all five reverted sites.

jetm (commented):
Re-reviewed the amend since 4569921, checking my three earlier findings rather than re-reading the whole change.

Two are closed. The stdout-on-auto-start one is genuinely fixed - I re-enumerated the emitters rather than trusting my old line numbers, and there is no `println!`, `print!` or stdout handle left anywhere under `src/utils/vm`. The `print_notice_above` bypass is fixed too, so the renderer-collision risk is gone.

The third is partial: the false "cannot interleave" claim is gone and the new prose is accurate, but the mitigation it promises never fires. That plus one gap in the new guard are inline, two findings.

The new test is not vacuous - reverting lifecycle.rs turns it red and it names all five sites. Six advisory notes were withheld rather than appended here.

hiagofranco:
Both applied, please take a look.

src/utils/output_format.rs
```diff
@@ -111,6 +111,38 @@ pub fn is_json_output_active() -> bool {
...
+    json_requested_in(std::env::args())
+}
+
+/// The scan itself, over an injected `argv` so it can be tested.
+///
+/// Both spellings clap accepts have to be recognised. The separated form alone
+/// was what `main`'s upgrade banner checked, which leaves `--output=json` — a
+/// form clap takes and callers do write — reading as a human-output run.
+fn json_requested_in(args: impl Iterator<Item = String>) -> bool {
+    let mut args = args.skip_while(|a| a != "--output" && a != "--output=json");
```
jetm: **The argv scan is subcommand-blind and first-occurrence-wins**

`json_requested_in` `skip_while`s to the FIRST `--output`-looking token anywhere on the line. clap is last-wins (`ArgAction::Set`), and three subcommands declare `trailing_var_arg` + `allow_hyphen_values` pass-through args - `sdk run` (main.rs:1519), `sdk dnf` (:1578), `runtime dnf` (:1787) - none of which has an `--output` flag of its own.

Both directions were reproduced with probe tests in a scratch copy, and both failed:

- `json_requested_in(["avocado","sdk","run","--","mytool","--output","json"])` returns `true`
- `json_requested_in(["avocado","sbom","--output","plain","--output","json"])` returns `false`

The first: on macOS/Windows (route.rs `resolve_mode` gates on `is_docker_desktop`), `avocado sdk run -- mytool --output json` is a pure human-output run, but `Commands::Sdk` is in `needs_vm_routing` (main.rs:2040) and the scan says json - so `print_stderr_notice` returns early and the user never sees "avocado-vm not running and AVOCADO_VM_DIR is unset; falling back to local docker" (route.rs:96). The build silently runs against local Docker Desktop.

The second is worse, because it is this change's own failure mode: a wrapper appending `--output json` to a command line that already carries `--output plain` makes clap emit JSON while the scan says human, so a raw ANSI `\x1b[93m[WARNING]\x1b[0m docker socket forward ... is missing` lands in the merged stream the desktop runner parses.

A single last-wins loop over argv that also skips everything after a bare `--` fixes both.

src/utils/output.rs
```diff
@@ -141,6 +141,51 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// `emit_json_object`, so a prepended event line breaks the one-object contract
+/// exactly as prose would. Silence is the only output that keeps the payload
+/// parseable, which is the same trade `main` already made.
+///
+/// The predicate is
+/// [`crate::utils::output_format::json_requested_on_command_line`], not
+/// `is_json_output_active`: the flag is set inside command `execute()`s and is
+/// still false everywhere these are called from.
+fn print_stderr_notice(formatted: &str) {
+    if crate::utils::output_format::json_requested_on_command_line() {
```
jetm: **The early return deletes VM failure notices from every stream**

`print_stderr_notice` returns before emitting anything, on all streams, whenever `--output json` appears in argv. The suppression is only needed for the one consumer that merges stdout and stderr; for the post-dispatch `lifecycle` callers an NDJSON event on the stream `vm update` is *already writing* would have been consumable, exactly as `print_warning_above` does at output.rs:82-86.

These are not chatter - they are non-fatal failure notices. lifecycle.rs:994 self-documents "Don't fail the whole boot", and route.rs:134's own comment says "Be loud".

Failure path: `avocado vm update --output json` where the hibernation supervisor fails to bind `127.0.0.1:<user_port>` within 5s. lifecycle.rs:994 warns and returns `Ok`, so update.rs:277 continues, update.rs:281 prints "avocado vm update: updated to <v>." and `execute` returns `Ok(())`. Before this increment the "SSH may not work until you restart" line reached stderr; now it exists nowhere on either stream, and the desktop sees a clean successful update over a VM whose SSH proxy is down. Same shape for the btrfs `/var` resize (lifecycle.rs:293/306) and the docker socket forward (lifecycle.rs:321).

src/utils/output.rs
```diff
@@ -141,6 +141,51 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+///
+/// Silent under `--output json`, and otherwise never suppressed — the inverse
+/// of the stdout printers, which drop the message under [`tui_is_active`] and
+/// so print nothing on every interactive run.
+///
+/// Stderr does not make a line safe here. The avocado-desktop CLI runner merges
+/// stdout and stderr in causal order and parses the result, which is why the
+/// upgrade banner in `main` is suppressed under `--output json`. Emitting an
+/// NDJSON event instead is not the escape it is for [`print_warning_above`]:
+/// these callers all run before the dispatch match, and `sbom` — the command
```
jetm: **"These callers all run before the dispatch match" is false**

That premise is load-bearing - it is what licenses "silence is the only output that keeps the payload parseable". Five of the nine callers are lifecycle.rs:293, :306, :321, :575 and :994, inside `lifecycle::start` / `apply_network_config`, reached from `UpdateCommand::execute` (commands/vm/update.rs:277), `StartCommand` (start.rs:94) and `ResetCommand` (reset.rs:111). `needs_vm_routing` (main.rs:2033) explicitly excludes `Commands::Vm`, so those five have **no** pre-dispatch path at all.

`vm update` also carries its own `--output json` (`VmCommands::Update`) and answers with a stream of `emit_json_object` calls (update.rs:366/387/446/468/485), not the single object the sentence attributes to `sbom`.

This is the same defect class as the finding I raised on this doc block last round: a stated fact about the call graph that the code contradicts. The cost is that the next author trusts it and designs around a call graph that does not exist.

jetm (commented):
Re-reviewed the increment since be54d62.

The needle-set fix is **confirmed fixed**, and confirmed the way it was found: injecting `print_warning_above("x")`, `print!("y")` and a raw `stdout().write_all(b"z")` separately into a scratch copy of `state.rs` turns the guard red on each, naming the needle it matched. A `writeln!(stdout().lock(), ...)` injection goes red too, and `eprint!`/`eprintln!` both stay green - so the blanking order does not false-positive, which was the trap I flagged when raising it. The earlier round's fixes have not regressed: no `println!`, `print!` or stdout handle remains reachable under `src/utils/vm/`.

The json-guard finding is **partial**. The dead `is_json_output_active()` branch and the false NDJSON-event claim are both gone, and the new predicate genuinely is true pre-dispatch. But the replacement introduces a new wrong call-graph claim in the same doc block, and the argv scan itself gives two demonstrably wrong answers. Three findings inline; all three are one-file fixes and only the argv one changes shipped behaviour.

Seven advisory notes were withheld rather than appended here. One is worth naming since it sits directly on this design: `NEEDLES` does not cover `emit_json_event`/`emit_json_object`, which write via `stdout().lock()` and match none of the seven spellings - and those are precisely the alternative output.rs:158-164 points a future author toward.

hiagofranco:
All three applied, please take a look

src/utils/output_format.rs
```diff
@@ -111,6 +111,58 @@ pub fn is_json_output_active() -> bool {
...
+/// - **Both spellings.** `--output=json` is valid clap input.
+///
+/// Known bound: a value-taking flag with `allow_hyphen_values` can swallow the
+/// literal `--output` (`--dnf-arg --output json`), which this reads as a format
+/// request. Resolving that needs the full arg spec — clap's job, not a scan's.
+fn json_requested_in(args: impl Iterator<Item = String>) -> bool {
+    let mut requested = false;
+    let mut args = args;
+    while let Some(arg) = args.next() {
+        if arg == "--" {
```
jetm: **The `--` cutoff misses separator-less `trailing_var_arg` pass-throughs**

Both inputs I reported last round are now correct - measured against the real in-tree function, not read:

- `["avocado","sdk","run","--","mytool","--output","json"]` → `false` (was `true`)
- `["avocado","sbom","--output","plain","--output","json"]` → `true` (was `false`)

Attached `--output=json`, mixed last-wins orderings, a second `--` inside the pass-through, `--output` as a valueless final token, and `--` as the *value* of `--output` all answer correctly too.

But the class is not closed, and it is the same subcommand-blindness the original finding was about. `trailing_var_arg = true, allow_hyphen_values = true` does **not** require a `--` separator: clap absorbs everything from the first unmatched token. That applies to `sdk run` (main.rs:1519), `sdk dnf` (:1578), `runtime dnf` (:1787) and `ext dnf` (:4449) - four commands, where the doc bullet at :141-144 names three and claims the class is handled.

Verified against the built binary, not inferred: `avocado sdk run --output json mytool` and `avocado sdk run mytool --output json` both parse with no clap error, and the control `avocado sdk run --totally-bogus-flag -C /nonexistent.yaml` falls back to the default `avocado.yaml` - proving `-C` and everything after the first unmatched token is absorbed into `command`.

Failure path: on macOS with `AVOCADO_VM_DIR` unset, `avocado sdk run mytool --output json` is a human-output run. The scan returns `true`, `ensure_routed_for_process` takes the silent branch at output.rs:190, and "avocado-vm not running and AVOCADO_VM_DIR is unset; falling back to local docker" (route.rs:96) never prints. The user watches docker fail with no explanation, on a run that has no JSON output to protect.

src/utils/output.rs
```diff
@@ -141,6 +141,68 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// Stderr alone does not make a line safe: the avocado-desktop CLI runner merges
+/// stdout and stderr in causal order and parses the result, which is why the
+/// upgrade banner in `main` is suppressed under `--output json`. So under JSON
+/// the choice is between a record and silence, never prose.
+///
+/// Which one depends on where the caller sits, and the callers sit in two
+/// places:
+///
+/// - **Post-dispatch, inside a command already emitting JSON.** The five
+///   `lifecycle` sites reached from `vm start` / `vm reset` / `vm update`, and
```
jetm: **The call-graph claim is wrong again, and the auto-start JSON path still drops notices**

"The five lifecycle sites reached from `vm start` / `vm reset` / `vm update` … `is_json_output_active` is true here" does not hold. I enumerated all ten callers rather than trusting the prose:

- pre-dispatch (5): main.rs:2122, route.rs:96, :102, :134, :182 - all inside `ensure_routed_for_process`, which main.rs:2117 calls before the dispatch match.
- lifecycle (5): lifecycle.rs:293, :306, :321, :575, :994.

`vm start` (main.rs:4738) and `vm reset` (main.rs:4817) declare no `--output` at all; only `vm update` (:4863) does. And `lifecycle::start` has **four** callers, not three - the fourth is route.rs:106, inside `ensure_routed_for_process`, pre-dispatch, installing no guard. So the flag is true for exactly one of the four.

Failure path: `avocado sbom --output json > sbom.json` on macOS with `AVOCADO_VM_DIR` set and the VM down. Routing auto-starts it, `spawn_supervisor` fails to bind (lifecycle.rs:994), the notice reaches output.rs:190 with the flag still false, and it is dropped from both streams. `sbom` then fails to reach the SDK container with no record of why.

That is the exact failure this increment's own update.rs:59-63 comment describes - still live on every `needs_vm_routing` command.

This is the third consecutive round in which this doc block asserts a call graph the code does not have. That pattern is worth more than the individual fixes: each round the prose is corrected to a new claim that is also untrue, which suggests it is being written from intent rather than from the callers.

src/utils/output.rs
```diff
@@ -141,6 +141,68 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+///   contract exactly as prose would, so this case stays silent — the same
+///   trade `main` already made for its banner.
+///
+/// The pre-dispatch predicate has to be
+/// [`crate::utils::output_format::json_requested_on_command_line`] rather than
+/// the flag, precisely because the flag is what is not yet set there.
+fn print_stderr_notice(formatted: &str, message: &str) {
+    if crate::utils::output_format::is_json_output_active() {
+        crate::utils::output_format::emit_json_event(
+            &serde_json::json!({"event": "warning", "message": message}),
```
jetm: **`print_info_stderr` emits an event typed `"warning"`**

The event literal hardcodes `"event": "warning"`, but `print_stderr_notice` serves both severities - so `print_info_stderr` (:203) now reports INFO-severity notices to the desktop as warnings. The copy reads correct in isolation because `print_warning_above` (:84) carries the same literal, which is the hazard of the duplication rather than a defence of it.

Failure path: `avocado vm update --output json` on a VM needing guest DNS applied. lifecycle.rs:575 `print_info_stderr("applied guest DNS: 1.1.1.1, 8.8.8.8")` is reached via update.rs:290 → `lifecycle::start` → `apply_network_config`, with the new guard live. The desktop receives `{"event":"warning","message":"applied guest DNS: …"}` for a routine success step and surfaces a warning on an update that went fine.

New in this increment: before it, that notice emitted nothing at all.

src/utils/output.rs
```diff
@@ -141,6 +141,68 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+///   `lifecycle` sites reached from `vm start` / `vm reset` / `vm update`, and
+///   anything reached once a [`crate::utils::output_format::JsonOutputGuard`]
+///   is live. `is_json_output_active` is true here, stdout is already carrying
+///   a stream, and one more event joins it — the same pairing
+///   [`print_warning_above`] uses. These are non-fatal *failure* notices
+///   ("Don't fail the whole boot"), so losing them is how a caller comes to
+///   believe a VM whose SSH proxy is down updated cleanly.
+/// - **Pre-dispatch, before any command owns stdout.** `route.rs` and `main`,
+///   where no guard exists yet and the command about to run may answer with a
+///   single object (`sbom`). A prepended event would break that one-object
```
jetm: **`sbom`'s one-object risk is post-dispatch, not pre-dispatch**

The bullet justifies pre-dispatch silence on the grounds that `sbom` answers "with a single object", which places the risk in the pre-dispatch window. sbom/generate.rs:500 installs a `JsonOutputGuard` that stays live for the **entire** command, so under the new rule at :184 any notice reached during `sbom` emits a bare NDJSON line into the middle of the SPDX document.

The rule "guard live → emit an event" is only sound for commands whose contract is a *stream*. `sbom`'s contract is one object, and the doc puts the boundary on the wrong side of dispatch.

Failure path: a future author adds a `print_warning_stderr` anywhere reachable from `sbom`'s `execute` - which :166-173 explicitly invites - and `avocado sbom --output json > sbom.json` writes an SPDX document with an `{"event":"warning"}` line spliced into it. `tests/no_stdout_on_the_vm_path.rs` cannot catch it: that guard scans `src/utils/vm` only, and this write is indirect through `output.rs`.

No such path exists today, so this is latent rather than live - flagging it because the doc actively points the next author toward it.

jetm (commented):
Re-reviewed the increment since f7ceaab. One of the four is fully closed; three are partial, and four findings are inline.

**Closed:** the guard now catches `emit_json_event` and `emit_json_object`. Demonstrated by injecting both into `route.rs` in a scratch clone and watching the test go red naming each line, not by reading the needle list.

**Partial, and worth reading in this order:**

- The argv scan answers both inputs I gave last round correctly now, plus every extra spelling I probed - attached `--output=json`, mixed last-wins orderings, a repeated `--`, a valueless trailing `--output`, and `--` as the flag's own value. What remains is the same class the original finding named: `trailing_var_arg` does not require a `--` at all.
- The dropped-notice fix works for `vm update` and only `vm update`. `lifecycle::start` has four callers; the pre-dispatch one installs no guard, so every `needs_vm_routing` command still drops all five notices - which is exactly the failure this increment's own comment describes.
- The doc block is on its third consecutive corrected-and-still-wrong call-graph claim.

The severity bug is new in this increment rather than pre-existing, so it is worth catching before it ships.

Nine advisory notes were withheld. One is a project-rule matter rather than a preference and I will name it here: neither behaviour change has a test. Nothing calls `print_warning_stderr`/`print_info_stderr`/`print_stderr_notice` from the suite and nothing constructs `UpdateCommand`, so deleting the `is_json_output_active` check - or update.rs:64-67 outright - restores the original defect with the suite green.

Build is clean at 857aa52: 1445 + 1454 unit tests, all integration suites, clippy silent, fmt clean.

src/utils/output_format.rs
```diff
@@ -111,6 +111,71 @@ pub fn is_json_output_active() -> bool {
...
+/// - **Both spellings.** `--output=json` is valid clap input.
+///
+/// Known bound: a value-taking flag with `allow_hyphen_values` can swallow the
+/// literal `--output` (`--dnf-arg --output json`), which this reads as a format
+/// request. Resolving that needs the full arg spec — clap's job, not a scan's.
+fn json_requested_in(args: impl Iterator<Item = String>) -> bool {
+    let mut requested = false;
+    let mut args = args;
+    while let Some(arg) = args.next() {
+        if arg == "--" || PASS_THROUGH.contains(&arg.as_str()) {
```
jetm: **`PASS_THROUGH` matches any token, not a subcommand slot**

Both forms I reported are now stopped - measured at this head, not read:

- `["avocado","sdk","run","mytool","--output","json"]` → `false`
- `["avocado","sdk","run","--output","json","mytool"]` → `false`

All four declared pass-throughs behave. But the cutoff breaks on a bare `run` or `dnf` **anywhere** in argv rather than in a subcommand position, so a genuine `--output json` on a non-pass-through command now reads as human and the suppression this change exists to provide does not fire.

Measured pairs, scan result against what `Cli::try_parse_from` actually resolves in the same run:

```
["avocado","install","dnf","--output","json"]   scan false / clap Install(packages=["dnf"], output=Json)
["avocado","install","run","--output","json"]   scan false / clap Install(packages=["run"], output=Json)
["avocado","sdk","compile","run","--output","json"]              scan false
["avocado","install","--container-arg","run","--output","json"]  scan false
```

Failure path: `avocado install dnf --output json` installs the dnf RPM into a sysroot. `Commands::Install` is in `needs_vm_routing` (main.rs:2033), so `ensure_routed_for_process` runs pre-dispatch and route.rs:96 emits a raw ANSI `\x1b[93m[WARNING]\x1b[0m avocado-vm not running and AVOCADO_VM_DIR is unset...` into the stream avocado-desktop merges and parses - the exact corruption this change was written to prevent.

This is a **regression the increment introduces**: at 857aa52 the same argv returned `true` and was suppressed, because only `--` stopped the scan. Matching on a subcommand path - a `sdk`/`runtime`/`ext` parent immediately before the token, or a positional index - rather than on any token would close it.

src/utils/output.rs
```diff
@@ -141,6 +141,67 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// stream-versus-object; until something carries that, silence is the only
+/// answer that is right for every caller rather than for some of them.
+///
+/// The predicate is
+/// [`crate::utils::output_format::json_requested_on_command_line`] and not the
+/// `is_json_output_active` flag, because five of the ten callers sit in
+/// `ensure_routed_for_process`, which runs before the dispatch match and so
+/// before any guard exists.
+fn print_stderr_notice(formatted: &str) {
+    if stderr_notice_is_suppressed(crate::utils::output_format::json_requested_on_command_line()) {
```
jetm: **The suppression policy and its predicate are both unpinned**

The only new test exercises `stderr_notice_is_suppressed`, which is the identity function. Nothing tests that `print_stderr_notice` applies it, or that the predicate is the argv scan rather than the process flag - so both of this change's load-bearing claims can be reverted with the whole suite green.

Two mutations, each run against the full suite (`cargo test --offline`, all 15 binaries):

- Rewrite this line to `stderr_notice_is_suppressed(is_json_output_active())` - precisely the swap the doc at :173-177 says the code exists to avoid, and the one that reintroduces the original pre-dispatch bug because the flag is false before the dispatch match. **15/15 ok, 0 failures.**
- Delete the whole `if stderr_notice_is_suppressed(...) { return; }` block, removing the change's entire behaviour. **15/15 ok, 0 failures.**

CLAUDE.md's test-integrity rule asks that every test fail when the implementation is wrong. Worth noting you already applied that standard one function up - the `warning_sink` test comment at :285-287 rejects exactly this weaker form - and it did not carry down here. A test that captures the emitted line, or drives `print_stderr_notice` with an injected predicate, would pin it.

src/utils/output.rs
```diff
@@ -141,6 +141,67 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// here, and the same cost: under JSON these notices are lost outright, which
+/// for the non-fatal failures in `lifecycle` means a caller can read a clean
+/// success over a VM whose SSH proxy never came up.
+///
+/// An earlier version emitted an NDJSON event under JSON instead of going
+/// quiet, on the theory that a record joins a stream harmlessly. That is only
+/// true for commands whose JSON contract *is* a stream. `sbom` answers with a
+/// single object and holds a
+/// [`crate::utils::output_format::JsonOutputGuard`] for its whole execute, so
+/// the rule spliced a bare `{"event":"warning"}` line into the middle of an
```
jetm: **The stated reason for deleting the emit branch is not reachable**

The doc asserts the old rule "spliced a bare `{"event":"warning"}` line into the middle of an SPDX document". I traced all ten notice callers - main.rs:2122, route.rs:96/102/134/182, lifecycle.rs:293/306/321/575/994 - and that splice could not occur.

Every emitter runs pre-dispatch, before sbom installs its guard at generate.rs:500, and `SbomCommand::execute` calls none of them: it reaches `sysroot_scan` and the container helpers, which use `print_failure_notice` → `print_notice_above` and emit no event. So `avocado sbom --output json > sbom.json` at 857aa52 produced a clean document.

Meanwhile `lifecycle::start` has exactly four callers (route.rs:106 pre-dispatch, vm/start.rs:94, vm/reset.rs:111, vm/update.rs:277), and among `VmCommands` only Update, Config Get and Config List declare `--output`. So the deleted branch was reachable from exactly one command: `vm update --output json` - which at 857aa52 did put `{"event":"warning","message":"hibernation supervisor (pid ...) didn't bind ..."}` on the stream, and at this head emits nothing and exits 0.

The trade the doc describes as "silence is the only answer that is right for every caller" was priced against a caller that could not be hurt, and paid by the only caller that was being helped. Either say the sbom risk is latent-only, which is what it is, or let the stream case keep its event.

jetm (commented):
Re-reviewed the increment since 857aa52. One of the four is fixed; three findings inline.

**Fixed:** the severity bug is gone, by elimination rather than by carrying severity - `print_stderr_notice` lost its emit branch entirely, so an INFO can no longer surface as `event: "warning"`. Both severities still format and print correctly on the human path.

**The argv fix works for the cases I named and introduces a new one.** Both forms from last round now return `false`, and all four declared pass-throughs behave. But `PASS_THROUGH` is matched against every argv token rather than a subcommand slot, so `avocado install dnf --output json` - which clap resolves to `Install(packages=["dnf"], output=Json)` - now reads as human output. That is a regression against 857aa52, where the same line was correctly suppressed. Measured both ways in one run, scan result against `Cli::try_parse_from`.

**The dropped-notice finding is not fixed, and the increment now documents that as the intent.** `avocado vm update --output json` with a failed supervisor bind still reports a clean update over a VM whose SSH proxy is down; the scenario reproduces unchanged. I have anchored the disagreement to the doc rather than re-filing the bug, because the trade the comment describes was priced against a caller that could not be hurt - `sbom` never reaches a notice emitter - and paid by the only caller that was.

Worth saying plainly, since it is now the pattern rather than the incident: **this is the fourth consecutive round in which a doc claim about the call graph is one the code does not support.** The code has improved every round. The prose is the part that keeps failing review, and it is failing in a way that matters - two of the three findings here are arguments the comments make, not defects in the logic.

Four advisory notes were withheld. Two more things I could not anchor: `vm update`'s bare `println!` at update.rs:152/223/260/280 already corrupts its own `--output json` stdout and the new guard test covers only `src/utils/vm`, not `src/commands/vm` - both pre-existing and untouched here; and the CHANGELOG carries no entry for the notice-routing behaviour change at all.

Head is clean: `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, and the full suite all pass at 165351b.

hiagofranco:
Hi, pushed all three fixes, please take a look

jetm (commented):
Re-reviewed after the rebase onto 1f88d1b. **No code changed** - I diffed 165351b against c17a16b restricted to the files this commit owns and it is empty, and `git range-diff` across the two bases shows only CHANGELOG context pulled in from #192 and #199. So this is a clean rebase, not a revision.

I am re-posting the three findings from the previous round because the rebase **orphaned every inline comment on this PR**. All six of mine now report `outdated: true` with a null line, including the three posted hours before the push. That happens because the commit each was bound to is no longer in the PR's history - even though the lines themselves are byte-identical.

Two consequences worth knowing, since neither is visible from the PR page:

- The comments are collapsed as outdated, so the findings read as addressed when nothing about them changed.
- `open_threads` drops to zero, so the merge gate stops counting them. Nothing now blocks on three unfixed blocking findings.

**These are not new findings.** They are the same three, re-anchored to the current head so they are countable again: the `PASS_THROUGH` token match that regressed `avocado install dnf --output json`, the unpinned suppression policy that survives two mutations with the suite green, and the sbom splice the doc claims as past-tense fact that the call graph does not permit.

Nothing else to review here - the code is unchanged since my last pass.

src/utils/output_format.rs
```diff
@@ -111,6 +111,86 @@ pub fn is_json_output_active() -> bool {
...
+///
+/// - **Last occurrence wins.** `--output` is `ArgAction::Set`, so clap resolves
+///   `--output plain --output json` to json. Stopping at the first match reports
+///   human, and a wrapper appending `--output json` to a line that already
+///   carries a format is exactly how that arises.
+/// - **Nothing inside a pass-through is ours.** Four subcommands declare
+///   `trailing_var_arg = true, allow_hyphen_values = true` and have no
+///   `--output` of their own: `sdk run` (main.rs), `sdk dnf`, `runtime dnf`
+///   and `ext dnf`. A `--` separator stops the scan, but clap does not require
+///   one — it absorbs everything from the first unmatched token — so
```
jetm, line 167: **`PASS_THROUGH` matches any token, not a subcommand slot**

Both forms I reported are now stopped - measured at this head, not read:

- `["avocado","sdk","run","mytool","--output","json"]` → `false`
- `["avocado","sdk","run","--output","json","mytool"]` → `false`

All four declared pass-throughs behave. But the cutoff breaks on a bare `run` or `dnf` **anywhere** in argv rather than in a subcommand position, so a genuine `--output json` on a non-pass-through command now reads as human and the suppression this change exists to provide does not fire.

Measured pairs, scan result against what `Cli::try_parse_from` actually resolves in the same run:

```
["avocado","install","dnf","--output","json"]   scan false / clap Install(packages=["dnf"], output=Json)
["avocado","install","run","--output","json"]   scan false / clap Install(packages=["run"], output=Json)
["avocado","sdk","compile","run","--output","json"]              scan false
["avocado","install","--container-arg","run","--output","json"]  scan false
```

Failure path: `avocado install dnf --output json` installs the dnf RPM into a sysroot. `Commands::Install` is in `needs_vm_routing` (main.rs:2033), so `ensure_routed_for_process` runs pre-dispatch and route.rs:96 emits a raw ANSI `\x1b[93m[WARNING]\x1b[0m avocado-vm not running and AVOCADO_VM_DIR is unset...` into the stream avocado-desktop merges and parses - the exact corruption this change was written to prevent.

This is a **regression the increment introduces**: at 857aa52 the same argv returned `true` and was suppressed, because only `--` stopped the scan. Matching on a subcommand path - a `sdk`/`runtime`/`ext` parent immediately before the token, or a positional index - rather than on any token would close it.

src/utils/output.rs
```diff
@@ -141,6 +141,85 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// stream too.
+///
+/// `sbom` is the one caller the record was wrong for, and the earlier note here
+/// overstated how: `SbomCommand::execute` calls nothing in this module, so the
+/// record could not land in the *middle* of the document. It would have been
+/// **prepended** — one `{"event":"warning"}` from `ensure_routed_for_process`
+/// ahead of the single SPDX object — two objects where the contract says one.
+/// Latent rather than observed, and Docker-Desktop-only: every route.rs emitter
+/// sits behind `RoutingMode::Apply`.
+///
```
jetm, line 186: **The suppression policy and its predicate are both unpinned**

The only new test exercises `stderr_notice_is_suppressed`, which is the identity function. Nothing tests that `print_stderr_notice` applies it, or that the predicate is the argv scan rather than the process flag - so both of this change's load-bearing claims can be reverted with the whole suite green.

Two mutations, each run against the full suite (`cargo test --offline`, all 15 binaries):

- Rewrite this line to `stderr_notice_is_suppressed(is_json_output_active())` - precisely the swap the doc at :173-177 says the code exists to avoid, and the one that reintroduces the original pre-dispatch bug because the flag is false before the dispatch match. **15/15 ok, 0 failures.**
- Delete the whole `if stderr_notice_is_suppressed(...) { return; }` block, removing the change's entire behaviour. **15/15 ok, 0 failures.**

CLAUDE.md's test-integrity rule asks that every test fail when the implementation is wrong. Worth noting you already applied that standard one function up - the `warning_sink` test comment at :285-287 rejects exactly this weaker form - and it did not carry down here. A test that captures the emitted line, or drives `print_stderr_notice` with an injected predicate, would pin it.

src/utils/output.rs
```diff
@@ -141,6 +141,85 @@ pub fn warning_sink(renderer_active: bool, json_active: bool) -> WarningSink {
...
+/// here, and the same cost: under JSON these notices are lost outright, which
+/// for the non-fatal failures in `lifecycle` means a caller can read a clean
+/// success over a VM whose SSH proxy never came up.
+///
+/// An earlier version emitted an NDJSON `{"event":"warning"}` record under JSON
+/// instead of going quiet, on the theory that a record joins a stream
+/// harmlessly. For most callers it did: the pre-dispatch notices fire for every
+/// `needs_vm_routing` command that declares `--output`, and `build`, `install`
+/// and `provision` all answer in NDJSON, where one more record is well-formed.
+/// `vm update` is the case that lost most by the removal — the `lifecycle`
```
jetm, line 168: **The stated reason for deleting the emit branch is not reachable**

The doc asserts the old rule "spliced a bare `{"event":"warning"}` line into the middle of an SPDX document". I traced all ten notice callers - main.rs:2122, route.rs:96/102/134/182, lifecycle.rs:293/306/321/575/994 - and that splice could not occur.

Every emitter runs pre-dispatch, before sbom installs its guard at generate.rs:500, and `SbomCommand::execute` calls none of them: it reaches `sysroot_scan` and the container helpers, which use `print_failure_notice` → `print_notice_above` and emit no event. So `avocado sbom --output json > sbom.json` at 857aa52 produced a clean document.

Meanwhile `lifecycle::start` has exactly four callers (route.rs:106 pre-dispatch, vm/start.rs:94, vm/reset.rs:111, vm/update.rs:277), and among `VmCommands` only Update, Config Get and Config List declare `--output`. So the deleted branch was reachable from exactly one command: `vm update --output json` - which at 857aa52 did put `{"event":"warning","message":"hibernation supervisor (pid ...) didn't bind ..."}` on the stream, and at this head emits nothing and exits 0.

The trade the doc describes as "silence is the only answer that is right for every caller" was priced against a caller that could not be hurt, and paid by the only caller that was being helped. Either say the sbom risk is latent-only, which is what it is, or let the stream case keep its event.

src/utils/output_format.rs
jetm: Retracting this - it was posted in error and your change already fixes it.

`PASS_THROUGH` is now `[(&str, &str); 4]` matched as a `(parent, child)` pair, which is exactly the fix. `avocado install dnf --output json` no longer breaks the scan, because `install`/`dnf` is not one of the four pairs.

My apologies for the noise. I claimed in the review body that no code had changed since 165351b and re-posted this finding on that basis. That claim was wrong - see the top-level comment for what went wrong on my end.

src/utils/output.rs
jetm: Retracting this - posted in error, and your change addresses it.

`print_stderr_notice` now returns `Option<&str>` precisely so a test can see which predicate fed the decision, which is the gap I was pointing at: the behaviour was unobservable, so both mutations I ran left the suite green.

Apologies for the noise - I re-posted this claiming no code had changed since 165351b, and that claim was wrong. See the top-level comment.

src/utils/output.rs
jetm: Retracting this - posted in error, and your rewrite is better than what I asked for.

You did not just soften the claim: you worked out that the record would have been **prepended** rather than spliced mid-document, since `SbomCommand::execute` calls nothing in this module, and you noted it is Docker-Desktop-only because every route.rs emitter sits behind `RoutingMode::Apply`. Both corrections are right and neither was in my finding.

Apologies for the noise - I re-posted this claiming no code had changed since 165351b, which was wrong. See the top-level comment.

jetm:
Correction to my review a few minutes ago ([4961542468](https://github.com/avocado-linux/avocado-cli/pull/198#pullrequestreview-4961542468)): **its central claim is false and I have retracted all three findings it re-posted.**

I said the push from 165351b to c17a16b was "a clean rebase, not a revision" and that no code changed. It changed by 111 insertions and 31 deletions across `output.rs` and `output_format.rs`, and those changes address all three findings from my previous round:

- `PASS_THROUGH` became `(parent, child)` pairs, which fixes the `avocado install dnf --output json` regression.
- `print_stderr_notice` returns `Option<&str>` so the decision is observable from a test, which was the gap behind the unpinned-policy finding.
- The doc's sbom claim was corrected, and more precisely than I asked: prepended rather than spliced mid-document, and Docker-Desktop-only.

What went wrong on my end, since it is worth knowing if you see me do it again: I built the file list with a command substitution and passed it unquoted as a git pathspec. It did not word-split in that shell, so git received one multi-line string, matched nothing, and exited **0 with empty output**. I read empty-and-successful as "no changes" instead of checking that the pathspec had expanded. The rebase onto 1f88d1b made the mistake plausible - a plain two-dot diff really does show ~3500 lines of other people's merged work - which is why I did not question the empty result.

Sorry for the noise on your threads. The three findings are addressed as far as I can tell; I will re-review the actual increment properly rather than trusting this pass.

jetm (commented):
Re-reviewed the increment properly this time. My previous review on this head claimed no code had changed; that was wrong and I retracted it in https://github.com/avocado-linux/avocado-cli/pull/198#issuecomment-5329195477. This one supersedes it.

Scope I actually read: `165351b3` (my last real review) to `c17a16b8`, restricted to what this PR owns. The set difference between the two heads and the three main commits the rebase pulled in leaves exactly `src/utils/output.rs`, `src/utils/output_format.rs` and this PR's CHANGELOG hunk - 154 insertions, 50 deletions.

All three findings are fixed, and both new tests hold up under mutation. Swapping the `print_stderr_notice` callsite to `is_json_output_active()` fails `an_environment_notice_reads_argv_not_the_process_flag` with exactly `left: None, right: Some("[WARNING] x")`, which is what its comment predicts. Reverting `PASS_THROUGH` to bare-token matching fails `a_pass_through_name_used_as_a_value_is_not_a_pass_through` on the `install dnf --output json` case. The four pairs match the four `trailing_var_arg` sites one for one (main.rs:1519 `sdk run`, 1578 `sdk dnf`, 1787 `runtime dnf`, 4449 `ext dnf`), and holding `--output`'s consumed value out of `previous` is right. The CHANGELOG move fixes a real problem: `### Changed` and `### Breaking` had been splitting the `### Added` list, and Unreleased now reads Added / Changed / Breaking / Fixed in order.

Locally at `c17a16b8`: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, full suite green (1502 lib + 1511 bin + integration). Repo defines no required contexts, so that is behavioral evidence rather than a green required check.

Two things worth a look, one of them small.

The doc comment on `print_stderr_notice` now says `SbomCommand::execute` calls nothing in this module. It calls four - `src/commands/sbom/generate.rs:21` imports `print_info` and `print_success` from `crate::utils::output`, and `print_summary` reaches them at 1115, 1126, 1133 and 1148 from `execute`'s line 613. The conclusion still holds and the very next clause is what carries it: the record originates in `ensure_routed_for_process`, which runs before dispatch, so it lands ahead of the document rather than inside it. That timing argument is sufficient on its own, and the "calls nothing in this module" clause is both wrong and redundant next to it. Worth noting because the thing keeping sbom's own prose off stdout is narrower than "it never calls in here" - it is the `output_path.is_some()` arm at 613 plus `tui_is_active()`.

The bigger one is the tradeoff itself. The note names `vm update` as the caller that lost most, on the grounds that it is a stream and could have carried the record. `avocado vm update --output json` does not currently have a clean stream to protect: `src/commands/vm/update.rs` has four ungated `println!` on the apply path (152, 223, 260, 280), `json_mode` at 167 is computed after the first of them and only gates the progress bars, and `vm update` never enables a `JsonOutputGuard`, so `tui_is_active()` is false and nothing else suppresses them. Line 152 in particular emits a bare prose line and returns, so a `--output json` consumer gets no JSON at all. `--check` is properly gated; it is the apply path that is not. So the record looks dropped to protect a property the named beneficiary does not have. `tests/no_stdout_on_the_vm_path.rs` reasons that every emitter reachable from the path has to be stderr and not just the outermost one, which is the right argument - it just stops at `src/utils/vm`, and the same failure is live one directory over. Not asking for `update.rs` in this PR; it does change whether blanket silence is the right call, and pointing that guard at `src/commands/vm` too would catch it.

Small one: the `PASS_THROUGH` bounds paragraph reads as exhaustive about what the pair cannot separate, and there is a second direction. `--runs-on`, `--nfs-port` and `--sdk-arch` are `global = true` (main.rs:102-111), so a value of one sitting between the parent and the child makes `previous` the value rather than the parent - `avocado sdk --runs-on user@host dnf install foo` misses `("sdk", "dnf")` and the cutoff never fires, after which the wrapped tool's args get scanned. It fails in the safe direction and it is the same class as the `--dnf-arg --output json` bound already documented above it, so a sentence in that paragraph is all it needs.

hiagofranco:
Pushed both fixes, `pointing no_stdout_on_the_vm_path.rs at src/commands/vm` was rejected. That guard's rule is "never stdout", justified by src/utils/vm running before dispatch.

jetm (commented):
Confirmed. `2fb97ee` corrects all three - comment-only, no functional lines changed, so nothing to re-test.

The `SbomCommand::execute` line now says accurately what happens: `print_summary` does call `print_info`/`print_success`, on the `output_path.is_some()` arm specifically, and that arm is why the record still can't land in the document. The `vm update` framing no longer claims a clean beneficiary - it names the ungated `println!` still on the apply path, so the tradeoff is stated honestly rather than optimistically. And the `PASS_THROUGH` bounds note now covers the global-flag-value direction (`--runs-on` sitting between parent and child), landing safe the same way the existing bound does.

No other threads open.

jetm (approved):
Approving. All three prior findings are correctly addressed at `2fb97ee`, verified directly against the diff, and the change is comment-only - no functional lines moved, so nothing new needs testing. Repo defines no required contexts; the optional Test Suite, both build targets, Security Audit, and Windows compile are all green on this head.

PR #201, opened 2026-08-13
sdk run: quote argv elements instead of joining them raw

mobileoverlord:
## What

`sdk run` splices the user's argv into a shell script inside the container using a bare `cmd.join(" ")`, so the container shell re-splits it. This:

```
avocado sdk run -- bash -lc 'U=/opt/x; ls $U'
```

arrives in the container as:

```
bash -lc U=/opt/x; ls $U
```

The shell reads that as **two** commands — a throwaway `bash -lc U=/opt/x`, then `ls $U` where `$U` is expanded by the *outer* shell to nothing, i.e. plain `ls`.

It does not error. It prints plausible output for a different directory. I hit this while debugging an SDK sysroot and it produced two confidently wrong readings before I noticed — a debugging tool that lies is worse than one that fails.

## Fix

Quote each argv element with the existing `utils::runs_on::shell_escape` (promoted to `pub(crate)` — no new helper). Extracted `join_argv` so this is unit-testable without spinning a container.

## Behavior change

Argv is now treated as argv, matching `docker run` / `kubectl exec` convention. Anyone currently passing a single string of shell code:

```
avocado sdk run -- 'ls /foo && ls /bar'      # was: interpreted as shell
```

now gets that string as one command word, and should use the explicit form:

```
avocado sdk run -- bash -lc 'ls /foo && ls /bar'
```

which is what the flag combination already implies, and which only works correctly *after* this fix. Flagging it since it is user-visible.

## Tests

Two unit tests in `src/commands/sdk/run.rs` — argv-boundary preservation (the exact `bash -lc 'X=1; echo $X'` case) and embedded-single-quote escaping. `cargo test --bin avocado sdk::run` — 11 passed. fmt and clippy clean.

Separate from #200; no overlap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   is skipped with nothing to say rather than asserting a version it never read.
 
 ### Breaking
+- **`sdk run` passes its argv through literally; the container shell no longer
+  re-splits or expands it.** Anything passing a `$VAR`, a glob, or a `&&` chain
+  as a single argument and counting on the container shell to interpret it now
+  gets that string through as one command word: `avocado sdk run -- 'ls /foo &&
+  ls /bar'` no longer runs two commands, and `avocado sdk run -- echo '$HOME'`
+  prints a literal `$HOME`. Ask for a shell explicitly instead — `avocado sdk
+  run -- bash -lc 'ls /foo && ls /bar'` — which is the form the flags already
+  implied and which only works correctly after this change.
+
+  This one breaks silently at runtime: there is no error and no warning, so a
+  pinned-version bump in CI will not fail the build, it will just run a
+  different command. Audit any job that shells through `sdk run` before
+  upgrading.
+
+  **What it fixes.** The arguments after `--` were spliced into the container's
+  script with a bare `join(" ")`, so the shell inside re-parsed them. `avocado
+  sdk run -- bash -lc 'U=/opt/x; ls $U'` arrived as `bash -lc U=/opt/x; ls $U`,
+  which the shell read as two commands and whose `$U` the *outer* shell expanded
+  to nothing — printing plausible output for a different directory rather than
+  failing. Each element is quoted now, so argv is argv, matching `docker run`
+  and `kubectl exec`.
 - **`is_version_compatible` returns `Option<bool>` instead of `bool`.** Only
   affects consumers of the `avocado_cli` lib target. `None` means the remote
   version could not be read at all, which the old `bool` could not express -
@@ -49,23 +70,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   that collapse is what let an unparseable version fall through as a pass.
 
 ### Fixed
-- **`sdk run` no longer lets the container shell re-split its argv.** The
-  arguments after `--` were spliced into the container's script with a bare
-  `join(" ")`, so the shell inside re-parsed them. `avocado sdk run -- bash
-  -lc 'U=/opt/x; ls $U'` arrived as `bash -lc U=/opt/x; ls $U`, which the shell
-  read as two commands and whose `$U` the *outer* shell expanded to nothing —
-  printing plausible output for a different directory rather than failing.
-  Each element is quoted now, so argv is argv, matching `docker run` and
-  `kubectl exec`.
-
-  **This removes an expansion that some invocations relied on.** Anything
-  passing a `$VAR`, a glob, or a `&&` chain as a single argument and counting
-  on the container shell to interpret it now gets that string through
-  literally: `avocado sdk run -- 'ls /foo && ls /bar'` becomes one command
-  word, and `avocado sdk run -- echo '$HOME'` prints `$HOME`. Ask for a shell
-  explicitly instead — `avocado sdk run -- bash -lc 'ls /foo && ls /bar'` —
-  which is the form the flags already implied and which only works correctly
-  after this change.
 - **Rootfs and initramfs no longer reinstall on every run.** `avocado sdk
   install` wiped and rebuilt both sysroots from scratch on every invocation,
   even with nothing changed. Removal detection compared the lockfile against
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   that collapse is what let an unparseable version fall through as a pass.
 
 ### Fixed
+- **`sdk run` no longer lets the container shell re-split its argv.** The
+  arguments after `--` were spliced into the container's script with a bare
+  `join(" ")`, so the shell inside re-parsed them. `avocado sdk run -- bash
+  -lc 'U=/opt/x; ls $U'` arrived as `bash -lc U=/opt/x; ls $U`, which the shell
+  read as two commands and whose `$U` the *outer* shell expanded to nothing —
+  printing plausible output for a different directory rather than failing.
+  Each element is quoted now, so argv is argv, matching `docker run` and
+  `kubectl exec`.
+
+  **This removes an expansion that some invocations relied on.** Anything
+  passing a `$VAR`, a glob, or a `&&` chain as a single argument and counting
+  on the container shell to interpret it now gets that string through
+  literally: `avocado sdk run -- 'ls /foo && ls /bar'` becomes one command
+  word, and `avocado sdk run -- echo '$HOME'` prints `$HOME`. Ask for a shell
+  explicitly instead — `avocado sdk run -- bash -lc 'ls /foo && ls /bar'` —
+  which is the form the flags already implied and which only works correctly
+  after this change.
 - **Rootfs and initramfs no longer reinstall on every run.** `avocado sdk
   install` wiped and rebuilt both sysroots from scratch on every invocation,
   even with nothing changed. Removal detection compared the lockfile against
--- a/src/commands/sdk/run.rs
+++ b/src/commands/sdk/run.rs
@@ -11,6 +11,7 @@ use crate::utils::{
     config::{ComposedConfig, Config},
     container::{RunConfig, SdkContainer},
     output::{print_info, print_success, OutputLevel},
+    shell::shell_escape,
     target::validate_and_log_target,
 };
 
@@ -20,7 +21,7 @@ use crate::utils::{
 /// element must be quoted or the container shell re-splits it.
 fn join_argv(argv: &[String]) -> String {
     argv.iter()
-        .map(|a| crate::utils::runs_on::shell_escape(a))
+        .map(|a| shell_escape(a))
         .collect::<Vec<_>>()
         .join(" ")
 }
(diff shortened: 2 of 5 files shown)


jetm (commented):
The fix is right and I verified the premise rather than taking it on faith: `RunConfig.command` really is spliced into `bash -c` at four sites in `src/utils/container.rs` (944/1153/1543/1841), and the `--runs-on` path does not double-escape, because `build_docker_command` already wraps the whole script in `shell_escape`. Both new tests assert the exact escaped strings rather than truthiness, and `main.rs:3404` maps an empty argv to `None`, so `join_argv` never sees an empty slice. `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo test --lib join_argv` all pass at `e7806cb`.

One thing I would add before merging.

**This is a user-visible behavior change with no CHANGELOG entry.** Anyone who was single-quoting a `$VAR` or a glob into `sdk run` and relying on the container shell to expand it loses that silently. I confirmed both halves in bash: `bash -c "'bash' '-lc' 'X=1; echo $X'"` now prints `1` (the fix), and `bash -c "'echo' '\$HOME'"` prints a literal `$HOME` (the loss). The repo keeps a populated Keep-a-Changelog Unreleased block, so there is a natural home for it — I would file it under Fixed with a line about the expansion change, since the people affected will not find it in the diff.

Three nits, all optional:

- `src/commands/sdk/run.rs:275` — stale error text inside the function you touched: "You must either provide a --command (-c)". `avocado sdk run --help` shows no such flag; the command is a trailing positional and `-c` is unassigned (`-C` is `--config`). I checked with `cargo run -- sdk run --help`.
- `src/commands/sdk/run.rs:22` — layering: a generic shell-quoting helper now comes from `utils::runs_on`, the remote-execution module, into a local non-remote path, and via an inline `crate::utils::runs_on::shell_escape(..)` rather than a `use`. `utils::shell` would be the honest home.
- The same untrusted-argv-into-`bash -c` splice is still raw in `sdk/dnf.rs:115`, `ext/dnf.rs:358` and `runtime/dnf.rs:303` (`self.command.join(" ")`). Out of scope here, but the class of bug is not closed by this PR.

mobileoverlord:
All three taken, plus the fourth as a filed follow-up. Pushed in c785a69.

**CHANGELOG.** Added under `Fixed`, per your suggestion of where it belongs. I gave the loss its own paragraph rather than a clause, because your `bash -c "'echo' '\$HOME'"` check is exactly the thing nobody will connect to a `sdk run` release note six weeks from now — so the entry names both concrete forms that stop working (`'ls /foo && ls /bar'` as one arg, `echo '$HOME'`) and the `bash -lc` form to move to.

**Stale flag text.** Fixed. Worth noting it was wrong in a second way beyond the flag not existing: the message told you to provide `--command` while the actual requirement is a trailing positional, so following it literally gets you `error: unexpected argument '--command'`. Now reads "You must either provide a command to run or use --interactive (-i)", and the test asserting on that string moved with it.

**Layering.** Agreed, and done — `shell_escape` now lives in `src/utils/shell.rs` with its four tests, and both consumers take it by `use`. Your framing is what convinced me: the direction of the dependency was backwards, a local non-remote path reaching into the remote-execution module for a generic helper. The follow-up below is the other half of the argument — three more non-remote callers want it, and each one reaching into `runs_on` would have been worse.

**The dnf sites.** Confirmed all three are the identical `self.command.join(" ")` shape, and filed as #207 rather than folded in here. Reasoning: the fix is now genuinely one line each, but it carries the same user-visible expansion loss this PR does, and making that call for `sdk dnf` / `ext dnf` / `runtime dnf` is a decision to take on its own terms rather than inherit from a `sdk run` fix. You are right that the class is not closed — #207 says so and names all three lines.

`cargo fmt`, `clippy --all-targets --all-features -D warnings`, 1392 lib tests and the 11 `sdk::run` tests are green at c785a69.

CHANGELOG.md
```diff
@@ -49,6 +49,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   that collapse is what let an unparseable version fall through as a pass.
 
 ### Fixed
+- **`sdk run` no longer lets the container shell re-split its argv.** The
```
jetm: **User-visible CLI break filed under Fixed, not Breaking**

This entry's own bold sentence concedes it "removes an expansion that some invocations relied on", yet it sits under `### Fixed`. Meanwhile the `### Breaking` section in this same `## [Unreleased]` block (line 44) holds only a change the text itself scopes to "consumers of the `avocado_cli` lib target".

Failure path: a maintainer bumps the pinned avocado CLI version in CI and reads only `### Breaking` for upgrade impact. It says lib-target consumers only, so they ship. A job running `avocado sdk run -- '$BUILD_DIR/x'` now passes a literal `$BUILD_DIR` instead of the expansion - no compile error, no warning, and the pre-change behaviour was already wrong-but-plausible output rather than a hard failure.

The asymmetry is what makes this worth changing: the entry that *is* filed under Breaking would have failed loudly as a compile error, while this one fails silently at runtime. Swapping which section each sits in would match the actual blast radius.

jetm (commented):
Re-reviewed the increment since e7806cb - the changelog/flag-text/shell_escape commit answering my last pass. The `shell_escape` move is byte-identical and every call site still resolves; the reworded flag text checks out against the actual flags. One blocking finding inline, on where the changelog files this change rather than on the code.

Six advisory notes were withheld rather than appended here, so the blocking one stays readable.

CHANGELOG.md
mobileoverlord: Taken — the `sdk run` entry now sits under `### Breaking`, in 4d54dfa.

Your failure path is the whole argument: a maintainer reading only `### Breaking` for upgrade impact would have found a sentence scoping the blast radius to lib-target consumers, and shipped a CI job that silently runs a different command. Filed there now, and placed *ahead* of the `is_version_compatible` entry so the section reads worst-first. I also reordered the entry itself to lead on the expansion loss with the bug as the second half ("**What it fixes.**"), since under Breaking the loss is the headline.

One half of your suggestion I did not take: I left `is_version_compatible` in `Breaking` rather than swapping it down. It is a genuine signature break for lib consumers, and moving it to `Fixed` would mean a lib consumer upgrading and hitting a compile error that the changelog no longer flags as breaking — trading your silent-failure gap for a new loud-failure one. What made the asymmetry wrong was the CLI break being *absent* from Breaking, not the API change being present, so I fixed that half and ordered the section by blast radius instead.

Also added the thing that was missing from both placements: an explicit note that this break produces no error and no warning, so a pinned-version bump will not fail the build. That is the sentence someone auditing an upgrade actually needs, and it was implied by neither section header.

Code is untouched since c785a69 — this is a `CHANGELOG.md`-only commit (21 insertions, 17 deletions, all inside the `## [Unreleased]` block).

jetm (commented):
Re-reviewed the increment since `c785a693` - one commit, `4d54dfa`, CHANGELOG only. No source file moved, so the code carries over from the last pass.

Reclassifying this as Breaking is right, and the reason given is the one that matters: the old invocations do not fail, they run something else. A silent behaviour change is worse for an upgrader than a loud one, and Breaking is where they look.

Checked each claim in the new text against the code rather than against the old entry. The base had `cmd.join(" ")` at `src/commands/sdk/run.rs:315`; head routes through `join_argv`, which maps `shell_escape` over every element, and `shell_escape` wraps in single quotes with `'\''` for embedded quotes. So a `$HOME`, a `&&` chain and a quoted glob all arrive as one literal word, which is what the entry says. The `bash -lc 'ls /foo && ls /bar'` recommendation holds too - under the old join that became `bash -lc ls /foo && ls /bar`, so "only works correctly after this change" is accurate rather than promotional.

Nothing was lost moving it. Every example, the `U=/opt/x` walkthrough and the `docker run` / `kubectl exec` comparison all survive in the new shape, and the section order under Unreleased reads Added / Changed / Breaking / Fixed with no duplicated heading. CI is green on `4d54dfa` including the test suite.

No findings.

CHANGELOG.md
```diff
@@ -52,6 +52,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   is skipped with nothing to say rather than asserting a version it never read.
 
 ### Breaking
+- **`sdk run` passes its argv through literally; the container shell no longer
```
jetm, line 55: **Published HIL doc teaches the pattern this breaks.** `docs/src/docs-guides/hardware-in-the-loop.md:140-146` (peridio/docs) instructs `avocado sdk run cd /opt/_avocado \&\& mkdir -p ./qemux86-64/extensions/my-app/usr \&\& echo "hello from host" \> ./qemux86-64/extensions/my-app/usr/hello.txt`, followed by an "Escaping SDK run commands" callout stating that `&&` and `>` are escaped precisely so they "reach the container's shell unchanged." That is exactly the contract this PR retires, and the vendor's own guide isn't in the audit list this entry asks users to check.

Replayed the documented argv against both code paths:
- Before: `cd <dir> && mkdir -p ./ext/usr && echo hello from host > ./ext/usr/hello.txt` -> exit 0, file created.
- After: each argv element quoted separately -> `bash: line 1: cd: too many arguments`, exit 2, nothing created.

The `echo`/redirect half in isolation is worse: it exits 0, prints `hello from host > ./out.txt` to stdout, and writes nothing - so someone following this guide sees a success and a missing file. Worth a companion docs fix landing with this change, not after.

CHANGELOG.md
```diff
@@ -52,6 +52,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  re-splits or expands it.** Anything passing a `$VAR`, a glob, or a `&&` chain
+  as a single argument and counting on the container shell to interpret it now
+  gets that string through as one command word: `avocado sdk run -- 'ls /foo &&
+  ls /bar'` no longer runs two commands, and `avocado sdk run -- echo '$HOME'`
+  prints a literal `$HOME`. Ask for a shell explicitly instead — `avocado sdk
+  run -- bash -lc 'ls /foo && ls /bar'` — which is the form the flags already
+  implied and which only works correctly after this change.
+
+  This one breaks silently at runtime: there is no error and no warning, so a
```
jetm, line 64: **"no error, no warning" doesn't hold for `&&` chains** - and the entry's own first example two sentences up falsifies it.

```
$ bash -c "'ls /foo && ls /bar'"
bash: line 1: ls /foo && ls /bar: No such file or directory
```

That's the exact `avocado sdk run -- 'ls /foo && ls /bar'` example from line 58 above, reproduced as what `join_argv` actually sends the container shell (a single element, `shell_escape`d into one quoted word since it's already one argv element). It exits 127 with a clear error, not silently. The `cd` recipe in the first comment on this PR exits 2, also loudly.

Only the expansion cases (`echo '$HOME'`) are genuinely silent - the `&&`/`>` chain cases fail loudly. Recommend splitting the claim: expansion breaks silently, but a `&&`/`|`/`>`/`;` chain passed as one argument now hard-fails at runtime, which is actually the *easier* case to catch in CI, not a risk to bury under "no error, no warning."

Separately, the audit list (`$VAR`, a glob, a `&&` chain) omits the assignment-prefix form, which bash only honors on a fully unquoted word: `avocado sdk run -- CC=clang make` now breaks silently (the one real case) while matching none of the three things this paragraph tells the reader to grep for.

jetm (commented):
Re-review after the head moved. Read the diff and every enclosing function, traced `RunConfig.command` through all four `full_command` sinks, ran the escape/sdk::run test suites and clippy, and reproduced the CHANGELOG's own worked example against both code paths. Two findings on the CHANGELOG entry itself, inline below - the code change is sound (escaping is complete POSIX single-quoting, verified against 16 payloads and 6 injection attempts, remote double-escape composes correctly).

jetm (approved):

PR #202, opened 2026-08-13
config: treat runtime package compile sections as active

mobileoverlord:
## What

`find_active_compile_sections` counts a compile section as active if it is reached from an extension or from `runtimes.<n>.kernel.compile`. A section reached through `runtimes.<n>.packages.<pkg>.compile` was never counted.

Consequences, in order of how confusing they are:

1. `need_target_dev` stays `false`, so **no target sysroot is created at all** — `$AVOCADO_PREFIX/sdk/target-sysroot` is simply absent (verified: `sdk/` contains only `target-repoconf` and `x86_64`).
2. The section's own `packages:` list is **silently discarded**.
3. The compile script still runs, with `--sysroot` pointing at a nonexistent directory. GCC tolerates that and falls back to built-in paths.

So a perfectly valid config has no effect, and the build fails much later on a missing target header or library with nothing pointing back at the cause.

## How it was found

`references#23` adds a `uboot-custom-imx8mp-evk` reference wired as:

```yaml
runtimes:
  dev:
    packages:
      uboot:
        compile: uboot
sdk:
  compile:
    uboot:
      packages:
        libgcc-s-dev: '*'      # silently ignored
```

It failed first on `sys/types.h`, then — after that was chased down — on `cannot find -lgcc` at the U-Boot link, because target `libgcc.a` only exists in `libgcc-s-dev` and that request was being dropped. The reference currently works around it by declaring the section under `kernel:` instead, with a comment pointing here; that workaround can be reverted once this lands.

## Tests

Three tests in `src/utils/config.rs`:
- the runtime-package case (fails on `main`, passes here)
- scalar package values (`avocado-runtime: '*'`) are not mappings and must not be misread
- the existing `kernel.compile` and dedup tests still pass

`cargo test --bin avocado` — 1401 passed. fmt and clippy clean.

Independent of #200 and #201.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,9 +58,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   That scan is also scoped to the current target now, the way the sibling
   extension scan already was. It previously read every runtime in the file
-  regardless of `runtimes.<name>.target` or `--runtime`, which in a
-  multi-target config provisioned target-dev for runtimes the user had not
-  asked to build.
+  regardless of `runtimes.<name>.target`, which in a multi-target config
+  provisioned target-dev for runtimes the user had not asked to build. The scan
+  reads the merged runtime config, so a `compile:` reference declared inside a
+  `target-<t>:` override block counts too — selecting runtimes from resolved
+  config while scanning unresolved config meant no target-dev sysroot was
+  installed for such a section, yet `runtime build` still ran its compile
+  script.
+- **A runtime that names its `target:` only inside a `target-<t>:` block is no
+  longer treated as targeting everything.** Override resolution strips the
+  non-matching blocks before target selection reads the runtime, so such a
+  runtime arrived with no `target:` key at all and matched the
+  "no target declared, applies to every target" branch. `avocado sdk install
+  --target qemux86-64` would install a raspberrypi4-only runtime's compile
+  packages into the x86-64 target sysroot. A `target-<t>:` block for the target
+  being built still keeps a runtime in scope even when it declares no `target:`
+  of its own. This also scopes the extension and runtime steps of `avocado
+  install`, which shared the selection logic through a duplicate that has been
+  collapsed onto the fixed one.
 - **Rootfs and initramfs no longer reinstall on every run.** `avocado sdk
   install` wiped and rebuilt both sysroots from scratch on every invocation,
   even with nothing changed. Removal detection compared the lockfile against
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   that collapse is what let an unparseable version fall through as a pass.
 
 ### Fixed
+- **A runtime package's `compile:` section now counts as active.** Only
+  `kernel.compile` and extension packages were scanned, so a section reached
+  through `runtimes.<name>.packages.<pkg>.compile` was treated as unused: no
+  target-dev sysroot was provisioned and the section's own `packages:` were
+  dropped without a word. The build then failed much later on missing target
+  headers or libraries, with nothing pointing back at the cause.
+
+  That scan is also scoped to the current target now, the way the sibling
+  extension scan already was. It previously read every runtime in the file
+  regardless of `runtimes.<name>.target` or `--runtime`, which in a
+  multi-target config provisioned target-dev for runtimes the user had not
+  asked to build.
 - **Rootfs and initramfs no longer reinstall on every run.** `avocado sdk
   install` wiped and rebuilt both sysroots from scratch on every invocation,
   even with nothing changed. Removal detection compared the lockfile against
--- a/src/utils/config.rs
+++ b/src/utils/config.rs
@@ -5607,11 +5607,74 @@ pub fn load_config<P: AsRef<Path>>(config_path: P) -> Result<Config> {
     Config::load(config_path)
 }
 
+/// Does this runtime declare its `target:` *only* inside `target-<t>:` override
+/// blocks, none of which is the current target?
+///
+/// `resolve_overrides_in_value` strips every non-matching `target-<t>:` block
+/// before a merged runtime is handed back, so a runtime shaped like
+///
+/// ```yaml
+/// runtimes:
+///   rpi:
+///     target-raspberrypi4:
+///       target: raspberrypi4
+/// ```
+///
+/// arrives with no `target` key at all when the current target is something
+/// else. Read naively that looks target-agnostic, which lands it in the
+/// include-for-all-targets branch and defeats the scoping: `sdk install
+/// --target qemux86-64` would install this runtime's u-boot compile packages
+/// into the x86-64 target sysroot.
+///
+/// Deliberately conservative - it reports true only when the config gives no
+/// sign whatsoever that the runtime applies to `target`. A `target-<target>:`
+/// block that exists but declares no `target:` of its own keeps the runtime
+/// target-agnostic and included, since the author plainly wrote something for
+/// this target.
+///
+/// The deprecated bare-target-name form (`raspberrypi4:` rather than
+/// `target-raspberrypi4:`) is not covered here; using it already warns.
+fn declares_target_only_for_other_targets(
+    raw_runtime: Option<&serde_yaml::Value>,
+    target: &str,
+) -> bool {
+    let Some(mapping) = raw_runtime.and_then(|v| v.as_mapping()) else {
+        return false;
+    };
+
+    // An unconditional `target:` needs nothing from us: the caller read it out
+    // of the merged value and already decided.
+    if mapping.get("target").is_some() {
+        return false;
+    }
+
+    let mut names_another_target = false;
+    for (key, sub_value) in mapping {
+        let Some(block_target) = key.as_str().and_then(|k| k.strip_prefix("target-")) else {
+            continue;
+        };
+        if block_target == target {
+            return false;
+        }
+        if sub_value.get("target").and_then(|t| t.as_str()).is_some() {
+            names_another_target = true;
+        }
+    }
+
+    names_another_target
+}
+
 /// Find runtimes that are relevant for the specified target.
 ///
 /// A runtime is relevant if:
 /// - It has a `target:` field matching the given target, OR
-/// - It has no `target:` field (matches all targets)
+/// - It has no `target:` field anywhere (matches all targets)
+///
+/// A runtime whose only `target:` sits inside a `target-<t>:` override block is
+/// relevant just to the targets it names there - see
+/// [`declares_target_only_for_other_targets`], which exists because override
+/// resolution has already stripped those blocks by the time the merged runtime
+/// is read here.
 ///
 /// If `requested_runtime` is Some, only that runtime is considered.
 pub fn find_target_relevant_runtimes(
@@ -5643,7 +5706,10 @@ pub fn find_target_relevant_runtimes(
                         if runtime_target == target {
                             relevant_runtimes.push(runtime_name.to_string());
                         }
-                    } else {
+                    } else if !declares_target_only_for_other_targets(
+                        runtime_section.get(runtime_name_val),
+                        target,
+                    ) {
                         // No target specified - include for all targets
                         relevant_runtimes.push(runtime_name.to_string());
                     }
@@ -5770,9 +5836,24 @@ pub fn find_active_compile_sections(
 
     if let Some(runtimes) = parsed.get("runtimes").and_then(|r| r.as_mapping()) {
         for runtime_name in &target_runtimes {
-            let Some(runtime_val) = runtimes.get(runtime_name.as_str()) else {
+            let Some(raw_runtime) = runtimes.get(runtime_name.as_str()) else {
                 continue;
             };
+
+            // Read the *merged* body, not the raw one. The names above were
+            // selected through `get_merged_runtime_config`, so reading the
+            // un-resolved mapping here would select on resolved config and then
+            // scan unresolved config: a compile ref declared inside a
+            // `target-<t>:` block is invisible to the raw read, while runtime
+            // build -- which reads the merged value -- still executes it. That
+            // pairing is worse than either half alone. `need_target_dev` stays
+            // false so no target-dev sysroot is installed, yet the compile-deps
+            // stamp is still written and later accepted on existence alone, so
+            // the compile script runs against a sysroot holding none of the
+            // packages it declared.
+            let merged_runtime =
+                config.get_merged_runtime_config(runtime_name, target, config_path)?;
+            let runtime_val = merged_runtime.as_ref().unwrap_or(raw_runtime);
 
             if let Some(section_name) = runtime_val
                 .get("kernel")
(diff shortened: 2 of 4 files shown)


jetm (commented):
The fix is correct and the premise checks out: `find_active_compile_sections` has exactly one production caller (`src/commands/sdk/install.rs:347`), where it gates `need_target_dev` and drives the target-dev package list, so the failure the commit body describes is real. The `packages.<pkg>.compile` shape is genuine and documented (`docs/features/sdk-cross-compilation-rpm-packaging.md:108`), and `find_compile_dependencies_in_ext` already handles the identical shape for extensions, so this closes a real asymmetry. No panic risk on the scalar case — `as_mapping()` and `get("compile")` both return `None` for `avocado-runtime: '*'`, which your second test pins. `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo test --lib find_active_compile_sections` all pass at `457ce34` (4 tests, including the two new ones).

One thing worth a look before merge.

**The new loop iterates every runtime, ignoring `runtimes.<n>.target` and `--runtime`.** `src/utils/config.rs:5770`. Its sibling `find_active_extensions`, called from the same place (`install.rs:185`), scopes through `find_target_relevant_runtimes` first. The gap is pre-existing on the `kernel.compile` line, so this is not a regression you introduced — but it widens it from the rare `kernel.compile` case to the common `packages.<pkg>.compile` one, which is the point of the PR. In a multi-target config that means provisioning target-dev for runtimes the user did not ask for.

One nit: no `CHANGELOG.md` Unreleased/Fixed entry for a user-visible bugfix. Weak signal though — only 5 of the last 15 src-touching commits added one, so the convention does not look enforced here.

Disclosure on my own verification: I tried the delete-the-hunk check to prove the new test actually fails without the production change, and my tooling blocked the edit, so I did not observe the RED. The static argument holds — `test_find_active_compile_sections_via_runtime_package` has no extensions and no `kernel.compile`, so the new block is the only path that can produce `["uboot"]` — but I am telling you that as reasoning rather than as something I watched fail.

mobileoverlord:
Both taken, in 63b69ba.

**Target scoping.** Fixed rather than deferred, on your reasoning: the gap predates the branch on the `kernel.compile` line, but this PR is what moves it from a rare case to the common one, so it belongs here. `find_active_compile_sections` now takes the same `(config, target, config_path, requested_runtime)` tuple as `find_active_extensions` and scopes through `find_target_relevant_runtimes` first, so the two scans called from `install.rs` agree about which runtimes exist.

**On the RED you couldn't observe** — worth closing that loop, since your static argument was right but you were explicit about not having watched it fail. I reverted the loop body to iterating every runtime and ran it:

```
test test_find_active_compile_sections_scopes_to_the_target ... FAILED
  left: ["uboot-rpi4", "uboot-x86"]
 right: ["uboot-x86"]
```

That is the new test, and the failure output is the bug verbatim — both targets' compile sections. The four older tests stayed green under that same mutation, which is what shows the coverage was genuinely absent rather than incidental.

The new test asserts from both directions (`qemux86-64` → `uboot-x86`, `raspberrypi4` → `uboot-rpi4`) so it fails on a scan that returns everything *and* on one that returns nothing. The four existing tests now build a real `Config` from a temp file instead of a bare `serde_yaml::Value`, since scoping needs the merge to resolve `runtimes.<n>.target`.

**CHANGELOG.** Added under `Fixed`, covering the original bug and the scoping as one entry. Your read of the convention matches mine — it is inconsistent in recent history — but the argument for adding it does not depend on that: the failure mode here is a build dying much later on missing target headers, and that is precisely the kind of thing someone greps a changelog for.

`cargo fmt`, `clippy --all-targets --all-features -D warnings` and 1393 lib tests green at 63b69ba.

src/utils/config.rs
```diff
@@ -5749,9 +5754,26 @@ pub fn find_active_compile_sections(
...
+    // `find_active_extensions` scopes its own scan. Reading every runtime here
+    // provisions target-dev for runtimes belonging to some other target, which
+    // is wasted work at best and the wrong toolchain at worst.
+    let target_runtimes =
+        find_target_relevant_runtimes(config, parsed, target, config_path, requested_runtime)?;
+
     if let Some(runtimes) = parsed.get("runtimes").and_then(|r| r.as_mapping()) {
-        for (_runtime_name, runtime_val) in runtimes {
+        for runtime_name in &target_runtimes {
+            let Some(runtime_val) = runtimes.get(runtime_name.as_str()) else {
```
jetm: **Selection is merged, body read is raw**

The loop picks runtime names via `find_target_relevant_runtimes`, which resolves `target-<t>:` blocks through `get_merged_runtime_config` - but then reads the runtime body straight out of the un-resolved `parsed` mapping. So a compile ref declared inside a `target-<t>:` override block is invisible here, while runtime build (which reads the merged value) still executes it.

Reproduced with this config:

```yaml
runtimes.dev.target-qemux86-64.packages.uboot:
  compile: uboot-x86
  install: install-uboot.sh
```

plus an `sdk.compile.uboot-x86.packages`. `get_merged_runtime_config("dev", "qemux86-64")` yields compile ref `Some("uboot-x86")`, but `find_active_compile_sections` returns `[]`.

The consequence chain is what makes it blocking rather than cosmetic: `need_target_dev` is false so no target-dev sysroot is installed; the compile-deps stamp is still written (install.rs:750 is gated only on `--no-stamps`); runtime build accepts that stamp on existence alone, since `runtime/build.rs` pushes only `StampComponent::Runtime` current-inputs so `validate_stamp` gets `None` and returns `Current`. `collect_runtime_compile_entries` then runs `uboot-compile.sh` against a sysroot holding none of its declared packages - the exact late failure on missing target headers this PR's CHANGELOG entry says is fixed.

src/utils/config.rs
```diff
@@ -5749,9 +5754,26 @@ pub fn find_active_compile_sections(
...
+    // no target sysroot was provisioned and the section's own `packages:` were
+    // silently dropped -- the build then failed much later on absent target
+    // headers/libraries with nothing pointing back at the cause.
+    //
+    // Scoped to the runtimes this target selects, the same way
+    // `find_active_extensions` scopes its own scan. Reading every runtime here
+    // provisions target-dev for runtimes belonging to some other target, which
+    // is wasted work at best and the wrong toolchain at worst.
+    let target_runtimes =
+        find_target_relevant_runtimes(config, parsed, target, config_path, requested_runtime)?;
```
jetm, line 5630: **Narrowing misses a `target-` declared target key**

When a runtime declares `target:` only inside a `target-<t>:` block, `resolve_overrides_in_value` strips that block for every non-matching target. The merged runtime then has no `target` key at all, so `find_target_relevant_runtimes` falls into its include-for-all-targets branch - defeating the scoping this call was added to perform.

Reproduced with:

```yaml
runtimes.rpi.target-raspberrypi4.target: raspberrypi4
runtimes.rpi.packages.uboot.compile: uboot-rpi4
```

Active sections come back as `["uboot-rpi4"]` for `qemux86-64` and also `["uboot-rpi4"]` for `raspberrypi4`. So `avocado sdk install --target qemux86-64` installs the raspberrypi4 u-boot compile packages into the x86-64 target sysroot - verbatim the "wrong toolchain at worst" case the comment added at 5764-5767 says it prevents.

The root cause sits in the untouched helper, but this call is the first to depend on it for compile-section scoping, which is why it surfaces here.

CHANGELOG.md
```diff
@@ -49,6 +49,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **A runtime package's `compile:` section now counts as active.** Only
+  `kernel.compile` and extension packages were scanned, so a section reached
+  through `runtimes.<name>.packages.<pkg>.compile` was treated as unused: no
+  target-dev sysroot was provisioned and the section's own `packages:` were
+  dropped without a word. The build then failed much later on missing target
+  headers or libraries, with nothing pointing back at the cause.
+
+  That scan is also scoped to the current target now, the way the sibling
+  extension scan already was. It previously read every runtime in the file
+  regardless of `runtimes.<name>.target` or `--runtime`, which in a
```
jetm: **CHANGELOG claims `--runtime` scoping that does not exist**

The entry says the scan previously read every runtime "regardless of `runtimes.<name>.target` or `--runtime`", which reads as both now being honoured. Only target is.

install.rs:352 passes `requested_runtime: None`, and `SdkInstallCommand` has no runtime field - so `avocado install -r dev` (main.rs:255, which delegates to `SdkInstallCommand` at install.rs:205) still gets an unscoped scan. A reader of this entry expects `-r dev` to narrow to the named runtime, and it does not.

Either thread `self.runtime` through, or drop the `--runtime` clause from the entry.

jetm (commented):
Re-reviewed the increment since 457ce34 - the scan-scoping and changelog commit answering my last pass. Three blocking findings inline; the two in `config.rs` were reproduced with throwaway probe tests, and both are cases where the narrowing now misses a section it should still find.

Seven advisory notes were withheld rather than appended here, so the blocking ones stay readable.

src/utils/config.rs
mobileoverlord: Fixed in 885c67f. The scan now reads the merged body — `get_merged_runtime_config` per selected runtime, falling back to the raw mapping when there is no merged value.

I reproduced your config as a failing test before touching the fix, since you flagged last round that you couldn't observe the RED on your own check. `probe_finding_one_compile_ref_under_target_override` returned `[]` against `["uboot-x86"]` at 63b69ba, and passes now. It ships as `test_find_active_compile_sections_reads_target_override_blocks`.

Your framing of it as "selection is merged, body read is raw" is what made the fix obvious — the two reads have to agree on which config they're looking at, and the sibling `find_active_extensions` had already picked merged. I also confirmed the two findings are independently necessary rather than one root cause seen twice: with this fix in and the 5769 narrowing reverted, your `target-qemux86-64` case passes and the `target-raspberrypi4` case still leaks; with the narrowing in and this read reverted, exactly the reverse.

The consequence chain you traced is in the code comment, because it is the part that would not survive a later refactor: someone reasonably concluding "the raw read is equivalent here, selection already filtered" needs to know that the stamp is written regardless (install.rs gated only on `--no-stamps`) and that `validate_stamp` sees `None` and returns `Current`, so the failure lands in `runtime build` on a sysroot that was never populated rather than anywhere near this function.

src/utils/config.rs
mobileoverlord: Fixed in 885c67f. Your repro is now `test_find_active_compile_sections_honors_target_declared_under_override` — it returned `["uboot-rpi4"]` for `qemux86-64` at 63b69ba, exactly the leak you described, and returns `[]` now while `raspberrypi4` still returns `["uboot-rpi4"]`.

You're right that the root cause is in the untouched helper, so that is where the fix went: `declares_target_only_for_other_targets`, consulted only in the `else` branch where the merged runtime has no `target` key. When the raw runtime declares `target:` inside some `target-<t>:` block and none of those blocks is the current target, the runtime is no longer treated as target-agnostic.

**Where I drew the line, since it is a judgment call rather than a mechanical fix.** The helper is deliberately conservative: it bails out the moment a `target-<current>:` block exists at all, even one declaring no `target:` of its own. So a mixed config —

```yaml
runtimes:
  multi:
    target-raspberrypi4:
      target: raspberrypi4
    target-qemux86-64:
      packages: { extra: '*' }
```

— keeps `multi` in scope for `qemux86-64`. That case is genuinely ambiguous: the declared `target:` keys read as an enumeration that excludes x86-64, while the `target-qemux86-64:` block reads as the author writing something for x86-64. I resolved it toward including, because the failure modes are not symmetric — a spurious include is the wasted work you flagged last round, a spurious exclude silently drops a runtime the author configured. `test_find_active_compile_sections_keeps_runtime_with_a_block_for_this_target` pins that half, so the choice is visible rather than incidental. Say the word if you read that case the other way.

**One thing I extended past your finding.** `commands/install.rs` had an independently maintained copy of `find_target_relevant_runtimes` — same logic, comments already drifted — used for the extension and runtime steps of `avocado install`. Fixing only the shared helper would have left `avocado install` scoping its SDK step one way and its ext/runtime steps another, which is a worse state than the uniform bug. That copy now delegates to the shared helper, so the fix lands on all three paths and ~55 lines of duplicate go away.

Not covered, and called out in the helper's doc comment: the deprecated bare-target-name form (`raspberrypi4:` instead of `target-raspberrypi4:`), which `resolve_overrides_in_value` also strips for sibling declared targets and which therefore has the same hole. It already warns on use, and covering it means duplicating that function's bare-key/`supported_targets` logic here. Happy to file it if you'd rather it not sit as a comment.

CHANGELOG.md
mobileoverlord: Confirmed and dropped, in 885c67f. You're right on both halves: `SdkInstallCommand` has no `runtime` field, `InstallCommand` never sets one when it constructs the command, and both scans inside it pass `requested_runtime: None` — so `install -r dev` gets an unscoped scan for compile sections *and* for `find_active_extensions`. The clause was describing scoping that does not exist.

I took the drop rather than the thread-through, and filed #209 for the other option. Reasoning: threading `self.runtime` into `find_active_compile_sections` alone would swap your false claim for a real inconsistency — compile sections scoped by runtime, SDK deps for extensions not. Threading it into both is the coherent version and is probably right, but it narrows which extensions get their SDK dependencies installed for a `-r`-scoped install, which is a user-visible change wanting its own tests and CHANGELOG entry rather than arriving inside a compile-section fix. #209 spells out the call sites and the suggested shape.

While rewriting the entry I added the `target-<t>:` half your other two findings exposed, since the original text claimed a scoping that the merged/raw split meant it did not actually deliver, and split the target-declared-under-override fix into its own bullet — it changes runtime selection for extension and runtime installs too, not just the compile-section scan, so it does not belong buried in a sub-paragraph of a `sdk install` entry.

src/utils/config.rs
```diff
@@ -5643,7 +5706,10 @@ pub fn find_target_relevant_runtimes(
                         if runtime_target == target {
                             relevant_runtimes.push(runtime_name.to_string());
                         }
-                    } else {
+                    } else if !declares_target_only_for_other_targets(
```
jetm, line 5504: **Narrowing skipped when the merged runtime config is `None`.** This `declares_target_only_for_other_targets` check only runs in the `Some(merged)` arm. The sibling `else` branch a few lines down (`config.rs:5716-5729`, unchanged by this PR) reads the raw runtime config directly and pushes it whenever there's no top-level `target` key - with no narrowing call at all. `get_merged_runtime_config` returns `Ok(None)` when override-stripping leaves an empty mapping, which is exactly a runtime whose entire body sits inside `target-<t>:` blocks.

Reproduced: a runtime declaring `target:` only under `target-raspberrypi4:` (no top-level `target`). `declares_target_only_for_other_targets` itself returns `true` (exclude) when probed directly, `get_merged_runtime_config` returns `None` for `qemux86-64`, yet `find_target_relevant_runtimes(..., "qemux86-64")` still returns this runtime - the bypass in the `None` branch overrides it. This is verbatim the CHANGELOG's second bullet, defeated by an empty merge.

All five new tests keep a key outside the override block (which routes them onto the fixed `Some` arm) - drop the stray top-level key from `honors_target_declared_under_override` and it should fail, since it would then exercise this exact branch.

jetm (commented):
One inline finding below. Three more that don't anchor to a diff line - each is in code this PR doesn't touch, but becomes reachable or inconsistent because of what it does touch:

**Empty runtime set now activates every top-level extension.** `find_active_extensions` (`config.rs:5778`, unchanged) treats an empty `target_runtimes` as "standalone extension config" and marks every top-level extension active - the mirror image of the finding above. Once `find_target_relevant_runtimes` narrows to `[]` for a real, target-scoped runtime, this branch fires and the active-extension set gets *wider*, not narrower. Measured by toggling the narrowing in one build: pre → `runtimes=["rpi"]`, `active_ext=["ext-a"]`; post → `runtimes=[]`, `active_ext=["ext-a","ext-b"]` (`ext-b` was never referenced). `avocado install` then reads that same empty runtime list the opposite way (`find_required_extensions`, `install.rs:447`) and concludes there's nothing to install - so the two paths disagree with each other.

**Two more copies of the old selection logic, both untouched.** `BuildCommand::get_runtimes_to_build` (`src/commands/build.rs:463-503`) is the pre-fix logic branch-for-branch - it's not in this diff. `SignCommand` (`src/commands/sign.rs:132-158`) is a partial clone with the same shape. Before this PR all three copies agreed (all wrong); now `avocado build`/`avocado sign` will select a runtime that `avocado install`/`sdk install` refuse for the same target. Worth a follow-up PR collapsing all three onto the one helper this PR just fixed.

**`sdk compile` doesn't share the new narrowing either.** `get_compile_sections_from_config` (`src/commands/sdk/compile.rs`) enumerates all of `sdk.compile` with no target scoping. `sdk install --target X && avocado sdk compile` (no explicit `--sections`) now runs sections `sdk install` deliberately skipped for that target - every in-tree caller passes explicit sections, so this only reaches a human running the bare command, but it's the same "late, unattributable failure" this PR exists to remove from the other direction.

On the two questions from the last review: confirmed both. The mixed-config include-on-ambiguity call is the right default (a spurious exclude silently drops configured work). And yes, the bare-target-name gap is real, not theoretical - confirmed it leaks the same way for both matching and non-matching targets when `supported_targets` is declared without a nested override.

jetm (dismissed):

PR #203, opened 2026-08-13
purge package-manager state from rootfs and initramfs images

mobileoverlord:
Found while reviewing #199. Splitting it out because it stands on its own and touches both image paths, not just the cpio.

## What

`dnf` installs into the shared sysroots leave package-manager bookkeeping behind, and `cp -a` carries it into the work copy that gets imaged. Measured on a `qemux86-64` build:

| tree | size | pkg-manager state |
|---|---|---|
| `rootfs` | 138M | 2.0M `var/lib/rpm`, 6.4M `var/cache/dnf` |
| `initramfs` | 123M | 2.5M `var/lib/rpm`, 4.3M `var/lib/dnf`, 6.4M `var/cache/dnf` |

dnf's logs come along too — `var/log/dnf.log`, `dnf.rpm.log`, `hawkey.log`, each line wall-clock stamped. `rootfs/install.rs` omits `$DNF_SDK_HOST_OPTS`, which is what redirects `logdir`/`cachedir`/`persistdir` at the SDK prefix, so the single omission that puts `var/cache/dnf` and `var/lib/dnf` in the sysroot puts the logs in `var/log`. Caught by @jetm in review.

Two problems with shipping it.

**It's dead weight.** There is no runtime package manager on target, so nothing ever reads any of it. That's ~13M off the rootfs and ~14M off the initramfs (11% of the staged tree).

**It blocks reproducibility.** Every package records a wall-clock stamp in the rpmdb:

```
busybox-udhcpc  INSTALLTIME=1773929326  INSTALLTID=1773929326
libsmartcols1   INSTALLTIME=1773929329  INSTALLTID=1773929326
```

240 of them in the initramfs. On top of that `var/lib/dnf/history.sqlite` records the transaction itself, and `var/cache/dnf` holds generated repodata plus per-repo `.solv` solvfiles. While any of that is in the tree, the same package set produces different image bytes on every install — a runtime build and a standalone build will never agree — regardless of how the archive or `mkfs` step is invoked.

To be exact about what this earns: removing it is **necessary and not sufficient**. `cpio --reproducible` is `--device-independent` on GNU cpio 2.15 and does not touch mtime, and the `rm` itself restamps the directories it empties. See the #199 section below.

## How

One `rm -rf` plus one `rm -f` per script, placed after `post_install` so state left by a hook's own `dnf` call is caught too.

Deliberately **not** added to `DEFAULT_ROOTFS_POST_INSTALL` / `DEFAULT_INITRAMFS_POST_INSTALL`: a user who defines `post_install` replaces those lists wholesale, and would silently lose both the size win and reproducibility.

## Safety

The pristine sysroots keep their rpmdb. Checked every consumer:

- `AVOCADO_OS_BUILD_ID` queries read `$ROOTFS_SYSROOT` / `$INITRAMFS_SYSROOT`
- installroot seeding in `ext install`, `ext dnf`, `runtime install`, `runtime dnf` copies from `$AVOCADO_PREFIX/rootfs`

None of them touch the work copy. Tests assert the sysroot rpmdb is not removed, so that stays true.

## Interaction with #199

Overlapping hunk — both insert into the same spot between the release-file injection and the archive step, and `initramfs/image.rs` gains a `mod tests` in both. On rebase the purge must land **before** #199's mtime-normalization pass, otherwise the `rm` re-stamps `/var/lib` and `/var/cache` with wall-clock time and undoes part of what that pass is for.

Neither PR alone gives a reproducible initramfs: #199 normalizes metadata, this normalizes contents. Both are needed.

## Test

Two regression tests, one per script: all six paths are purged from the work copy and the purge precedes `mkfs`/`cpio`. Deleting either `rm` line fails them.

These originally also asserted `!contains("$ROOTFS_SYSROOT/var/lib/rpm")`, which @jetm pointed out was vacuous — that string has no occurrences to begin with, and the build-ID query it claimed to protect is spelled `--root "$ROOTFS_SYSROOT"` with `--dbpath`. Removed rather than repaired; the sysroot's safety is argued above and is not something a substring check establishes.

Full suite green, clippy clean with `-D warnings`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   second field.
 
 ### Changed
+- **Rootfs and initramfs images no longer ship package-manager state.** The
+  rpmdb, `var/lib/dnf`, `var/cache/dnf` and dnf's own logs (`dnf.log`,
+  `dnf.rpm.log`, `hawkey.log`) are removed from the staged copy before the
+  image is built — measured at ~13MB of a qemux86-64 rootfs and 14MB of a
+  123MB initramfs. Nothing on target reads any of it; these systems have no
+  runtime package manager. Only the staged copy is touched, so the shared
+  sysroot keeps its database and `ext install` / `runtime install` still seed
+  their installroots from it.
+
+  Anything inspecting a built image for installed packages — `rpm -qa` against
+  a loop-mounted rootfs, say — needs to query the sysroot or the lockfile
+  instead. This is a necessary step toward reproducible images but not a
+  sufficient one on its own; archive mtime normalization is separate.
 - **A skipped remote version check no longer reports as a passed one.** When
   the remote's `--version` output cannot be parsed, `--runs-on` used to print
   `[SUCCESS] Remote avocado version: <whatever it read>`, which is a green line
--- a/src/commands/initramfs/image.rs
+++ b/src/commands/initramfs/image.rs
@@ -101,11 +101,13 @@ if [ -d "$INITRAMFS_SYSROOT/usr" ]; then
 
     # Purge package-manager bookkeeping from the work copy before archiving.
     # Same reasoning as the rootfs image — see the comment in
-    # `generate_rootfs_build_script`. Measured 14MB of a 123MB qemux86-64
-    # initramfs (2.5M rpmdb, 4.3M var/lib/dnf, 6.4M var/cache/dnf), all of it
-    # unreproducible and none of it read by anything in an initrd.
+    # `generate_rootfs_build_script`, including why dnf's logs come too and why
+    # this is necessary but not sufficient for reproducibility. Measured 14MB
+    # of a 123MB qemux86-64 initramfs (2.5M rpmdb, 4.3M var/lib/dnf, 6.4M
+    # var/cache/dnf), none of it read by anything in an initrd.
     echo "Purging package-manager state from initramfs image"
     rm -rf "$INITRAMFS_WORK/var/lib/rpm" "$INITRAMFS_WORK/var/lib/dnf" "$INITRAMFS_WORK/var/cache/dnf"
+    rm -f "$INITRAMFS_WORK/var/log/dnf.log" "$INITRAMFS_WORK/var/log/dnf.rpm.log" "$INITRAMFS_WORK/var/log/hawkey.log"
 
     # Build initramfs image using configured filesystem format
     INITRAMFS_FS="{initramfs_filesystem}"
@@ -443,7 +445,14 @@ mod tests {
             "",
         );
 
-        for path in ["var/lib/rpm", "var/lib/dnf", "var/cache/dnf"] {
+        for path in [
+            "var/lib/rpm",
+            "var/lib/dnf",
+            "var/cache/dnf",
+            "var/log/dnf.log",
+            "var/log/dnf.rpm.log",
+            "var/log/hawkey.log",
+        ] {
             assert!(
                 script.contains(&format!("\"$INITRAMFS_WORK/{path}\"")),
                 "{path} must be purged from the work copy before archiving"
@@ -458,11 +467,5 @@ mod tests {
             .find("cpio --reproducible")
             .expect("cpio step present");
         assert!(purge_at < cpio_at, "purge must precede cpio creation");
-
-        // The shared sysroot keeps its rpmdb — the build ID query reads it.
-        assert!(
-            !script.contains("$INITRAMFS_SYSROOT/var/lib/rpm"),
-            "the shared sysroot's rpmdb must not be removed"
-        );
     }
 }
--- a/src/commands/initramfs/image.rs
+++ b/src/commands/initramfs/image.rs
@@ -99,6 +99,14 @@ if [ -d "$INITRAMFS_SYSROOT/usr" ]; then
         echo "AVOCADO_OS_BUILD_ID=$INITRAMFS_BUILD_ID" >> "$INITRAMFS_WORK/usr/lib/os-release-initrd"
     fi
 
+    # Purge package-manager bookkeeping from the work copy before archiving.
+    # Same reasoning as the rootfs image — see the comment in
+    # `generate_rootfs_build_script`. Measured 14MB of a 123MB qemux86-64
+    # initramfs (2.5M rpmdb, 4.3M var/lib/dnf, 6.4M var/cache/dnf), all of it
+    # unreproducible and none of it read by anything in an initrd.
+    echo "Purging package-manager state from initramfs image"
+    rm -rf "$INITRAMFS_WORK/var/lib/rpm" "$INITRAMFS_WORK/var/lib/dnf" "$INITRAMFS_WORK/var/cache/dnf"
+
     # Build initramfs image using configured filesystem format
     INITRAMFS_FS="{initramfs_filesystem}"
     INITRAMFS_OUTPUT="$OUTPUT_DIR/avocado-image-initramfs-$TARGET_ARCH.$INITRAMFS_FS"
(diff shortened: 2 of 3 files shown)


jetm (commented):
The size win is real and I could not find a consumer this breaks. I traced every rpmdb reader — the build-ID query at `rootfs/image.rs:160`, `lockfile.rs:107-155`, the installroot seeding at `ext/install.rs:603`, `ext/dnf.rs:270`, `runtime/install.rs:491`, `runtime/dnf.rs:220` — and each targets a pristine sysroot or a separate installroot, never the work copy. Nothing later reads the work copy, no code mounts or extracts a built image, and the purge sits correctly after permissions, post_install and both os-release injections. `set -e` is safe here since a recursive remove on a missing path exits 0 and both work vars carry `:-` defaults. `cargo fmt`, `clippy -D warnings` and `cargo test --bin avocado purges_package_manager` all pass at `7f78038`.

Where I would push back is on the second half of the stated goal. The size reduction lands; reproducibility does not.

**Two identical builds still emit different bytes.** `cpio --reproducible` is `--device-independent` on GNU cpio 2.15 and does not touch mtime, and the removal itself restamps `var/lib` and `var/cache` with wall-clock time. I reproduced it: two identical trees, purge, `cpio --reproducible -o -H newc` — `var/lib` mtimes 13:40:48 vs 13:40:50, archives diverge at byte 518, different sha. `initramfs/image.rs:108`.

**And the purge misses dnf's own logs.** `rootfs/install.rs:951` omits `$DNF_SDK_HOST_OPTS` (`container.rs:2470`) — the same omission that puts `var/cache/dnf` and `var/lib/dnf` in the sysroot also puts `dnf.log`, `dnf.rpm.log` and `hawkey.log` in `$SYSROOT/var/log`, and those carry per-line wall-clock timestamps. I confirmed the mechanism with `dnf5 --dump-main-config`: `logdir` is installroot-prefixed. `fetch.rs:468+487` reads like the repo already knowing about this path. Same on initramfs.

Nits:

- `src/commands/ext/image.rs:895` — extension `.raw` images mkfs straight from the ext sysroot, which `ext/install.rs:603` seeds with a full rootfs rpmdb copy. Same problem, not covered here, and `--exclude-path` is already wired there for `var_excludes`.
- `src/commands/rootfs/image.rs:599` — the `!contains("$ROOTFS_SYSROOT/var/lib/rpm")` test guard is vacuous: that string has 0 occurrences at the merge-base, and the build-ID query it claims to protect is spelled `--root "$ROOTFS_SYSROOT"` with `--dbpath`. Same at `initramfs/image.rs:464`. The three positive assertions do real work and fail if the purge line is deleted; this one asserts nothing.
- No `CHANGELOG.md` Unreleased entry for a user-visible artifact change. Closest precedent in the same subsystem (#185) added one.

Two things I could not verify, stated as gaps rather than findings: my sandbox refused every `mkfs.erofs` invocation including `--help`, so whether the rootfs erofs shares the mtime problem (`-T` is passed, `--ignore-mtime` is not) is untested; and I could not run a real SDK-container build, so the `var/log/dnf.log` point rests on reading the code paths rather than on an observed file.

mobileoverlord:
Pushed in 4a9774e. Both findings taken, plus two of the three nits.

**The `var/log` miss is real and now fixed.** This is the best catch in the review, because the mechanism you inferred is the *same* omission that produces the state the PR already purges — `rootfs/install.rs` runs `$DNF_SDK_HOST $DNF_SDK_TARGET_REPO_CONF` without `$DNF_SDK_HOST_OPTS`, and that variable is the one setting `logdir`, `cachedir` and `persistdir` at the SDK prefix. So the 6.4M `var/cache/dnf` I measured is the observed half of exactly the thing you reasoned about for `logdir`. That is stronger than a code read: the cache being there is proof the redirect is absent.

Both images now `rm -f` `dnf.log`, `dnf.rpm.log` and `hawkey.log`, and both tests assert all six paths. Deleting the new line fails them (`var/log/dnf.log must be purged from the work copy before imaging`).

**Reproducibility.** Agreed, and the wording is scoped now — comments and PR body both say removal is necessary and not sufficient. Two notes on where we already agreed: the body's "Interaction with #199" section called out the restamp (`the rm re-stamps /var/lib and /var/cache with wall-clock time`) and said neither PR alone gives a reproducible initramfs. What was wrong was the section *heading* — "It's why the images aren't reproducible" reads as a claim to have fixed it, and a heading is what people take away. It says "It blocks reproducibility" now, which is what I can defend.

Your cpio measurement settles the `--reproducible`/`--device-independent` question for good, so thank you for running it rather than reading the man page.

**Vacuous test guard.** Deleted from both files, not repaired. You are right that it asserts nothing, and the honest fix is removal rather than rewriting it into a check of something else: the sysroot's safety rests on the consumer trace in the PR body, not on a substring. A test that cannot fail is worse than an absent one because it reads as coverage.

**`ext/image.rs:895`** — that is #205, which was open when you wrote this and which you have since reviewed. Excluded there rather than deleted, for the reason you confirmed on that PR.

**CHANGELOG.** Added under `Changed`, naming the one consumer this can surprise: anything running `rpm -qa` against a built image rather than against the sysroot.

On your two stated gaps — the erofs `-T` question is #204's territory and I have flagged it as open there too; nobody in this review has erofs-utils available, so it stays unanswered rather than assumed.

src/commands/rootfs/image.rs
```diff
@@ -172,6 +172,37 @@ if [ -d "$ROOTFS_SYSROOT/usr" ]; then
...
+    # half, not this one's.
+    #
+    # Runs after post_install so state left by a hook's own dnf call is caught
+    # too. Safe for identity and for extension priming: the build ID above
+    # queries $ROOTFS_SYSROOT, and the installroot seeding in `ext install` /
+    # `runtime install` copies from $AVOCADO_PREFIX/rootfs — all the pristine
+    # sysroot, never this work copy.
+    echo "Purging package-manager state from rootfs image"
+    rm -rf "$ROOTFS_WORK/var/lib/rpm" "$ROOTFS_WORK/var/lib/dnf" "$ROOTFS_WORK/var/cache/dnf"
+    rm -f "$ROOTFS_WORK/var/log/dnf.log" "$ROOTFS_WORK/var/log/dnf.rpm.log" "$ROOTFS_WORK/var/log/hawkey.log"
```
jetm: **dnf log purge removes nothing; it is a no-op**

dnf never writes its logs into an `--installroot`, so these three operands match nothing in the staged rootfs. The mechanism the new comment states at line 189 - "the same omission that lands `var/cache/dnf` and `var/lib/dnf` in the sysroot also lands `dnf.log`, `dnf.rpm.log` and `hawkey.log` in `var/log`" - is not how dnf behaves.

dnf 4.24's `dnf/cli/cli.py:1033` loops `prepend_installroot` over `('cachedir', 'persistdir')` only. `logdir` is never installroot-relative, so the logs land in the SDK prefix rather than the image.

Checked against six real avocado build volumes on this box (five imx93-frdm, one qemux86-64): zero `.log` files anywhere under any `rootfs/` or `initramfs/` tree, `var/log -> volatile/log` resolves to an empty directory in every one, and all four dnf logs sit in `sdk/x86_64/var/log`.

So the `rm -f` exits 0 having deleted nothing while the build prints "Purging package-manager state from rootfs image". The two new test cases assert only that the generated script *contains* these strings, so they pass without detecting any of it.

src/commands/initramfs/image.rs
```diff
@@ -99,6 +99,16 @@ if [ -d "$INITRAMFS_SYSROOT/usr" ]; then
...
 
+    # Purge package-manager bookkeeping from the work copy before archiving.
+    # Same reasoning as the rootfs image — see the comment in
+    # `generate_rootfs_build_script`, including why dnf's logs come too and why
+    # this is necessary but not sufficient for reproducibility. Measured 14MB
+    # of a 123MB qemux86-64 initramfs (2.5M rpmdb, 4.3M var/lib/dnf, 6.4M
+    # var/cache/dnf), none of it read by anything in an initrd.
+    echo "Purging package-manager state from initramfs image"
+    rm -rf "$INITRAMFS_WORK/var/lib/rpm" "$INITRAMFS_WORK/var/lib/dnf" "$INITRAMFS_WORK/var/cache/dnf"
+    rm -f "$INITRAMFS_WORK/var/log/dnf.log" "$INITRAMFS_WORK/var/log/dnf.rpm.log" "$INITRAMFS_WORK/var/log/hawkey.log"
```
jetm: **Same no-op log purge on the initramfs path**

Identical dead `rm -f` here, and the reworded comment above it now points the reader at the rootfs comment "including why dnf's logs come too" - propagating the same incorrect mechanism to a second site.

`initramfs/var/volatile/log` (the target of `var/log`) is empty in every build volume I inspected, including the qemux86-64 one the PR body measures. The line deletes nothing on any build.

CHANGELOG.md
```diff
@@ -31,6 +31,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   second field.
 
 ### Changed
+- **Rootfs and initramfs images no longer ship package-manager state.** The
+  rpmdb, `var/lib/dnf`, `var/cache/dnf` and dnf's own logs (`dnf.log`,
+  `dnf.rpm.log`, `hawkey.log`) are removed from the staged copy before the
```
jetm: **Changelog documents removing files that were never staged**

The entry tells users dnf's logs "are removed from the staged copy before the image is built". No dnf log has ever been in the staged copy, so this documents a removal that does not happen.

The cost is that a reader believes wall-clock-stamped log files were stripped from their image and stops looking for them. And if a future config change ever did put them there, nothing in this change would catch it - the purge is unconditional and its tests only match strings in the generated script.

The rpmdb/cache half of this entry is accurate; it is specifically the log clause that describes something the code does not do.

jetm (commented):
Re-reviewed the increment since 7f78038. Two of the three claims in the subject hold up: the dropped assertion really was vacuous and nothing real went with it, and the scoped claim is accurate for the rpmdb/cache half. The log purge is the problem - three blocking findings inline, all one root cause.

Two advisory notes were withheld rather than appended here. One item has no diff line to anchor to: the PR body's Safety section still says "Tests assert the sysroot rpmdb is not removed, so that stays true", which the same body's Test section now contradicts - worth fixing the body.

src/commands/rootfs/image.rs
mobileoverlord: Confirmed — your dnf/cli.py reading and the six-volume check both hold. Fixed in 253fe08: the log operands are gone rather than kept as insurance (an unconditional dead purge whose tests string-match the script reads as coverage and provides none), the comment now states the real mechanism (logdir is never installroot-prefixed; the logs live under the SDK prefix), and both tests pin the ABSENCE of the log operands so the wrong mechanism cannot quietly return.

src/commands/initramfs/image.rs
mobileoverlord: Fixed in 253fe08 — dead rm dropped here too, and the comment now points at the corrected rootfs rationale (why the logs are NOT on the list) instead of propagating the wrong one. Also rebased over the mtime-normalization work that landed on main, with the purge ordered before the mtime pass so the directories the purge restamps get normalized.

CHANGELOG.md
mobileoverlord: Fixed in 253fe08 — the entry now documents only the rpmdb/cache removal that actually happens, and states explicitly that dnf logs never land in the staged copy (and where they actually live), so nobody stops looking for them in the wrong place.

PR #206, opened 2026-08-13
fix(vm): sync the new runtime into var instead of destroying it

mobileoverlord:
Stacked on #196 — targets `fix-vm-update-reset-var` so the diff reads as a delta. Squashing into #196 is fine; the two are one change in intent.

## Why this replaces the reset

#196 deletes `var.btrfs` on a seed change. That also deletes `/var/lib/docker`, which holds the SDK image layers and the `avo-<uuid>` volume that **is** `$AVOCADO_PREFIX` — hence the release note telling people to re-run `avocado install` and `avocado build`.

That was never necessary. The avocado-owned half of var is already content-addressed and versioned (`images/`, `runtimes/`, `active`), and `avocadoctl` already installs a runtime alongside the current one and switches `active` only once it is staged and verified. The guest can adopt the release's runtime out of the new seed and leave everything else alone.

The btrfs work has to happen guest-side regardless: this VM exists *because* the host is macOS or Windows, neither of which can mount a btrfs image.

## Change

- `vm update` records `runtime.pending_var_seed_sha` instead of unlinking the disk.
- `vm start` attaches that seed read-only and names its sha **and filesystem UUID** on the cmdline. The guest echoes the sha into a stamp, which is what lets the marker be cleared for *this* sync. Every failure leaves it set, so the next start retries — that is the whole recovery story.
- **Never attached when the live disk was byte-copied from that same seed.** `seed_var_disk` is a `std::fs::copy`, so the two carry an identical btrfs fsid and the kernel would read them as one multi-device filesystem. A correctness gate, not an optimisation.

## The ordering bug this found

The first version addressed the seed by `serial=`, which needs `if=none` plus an explicit `-device`. An explicit `-device virtio-blk-pci` takes a **lower PCI slot** than the devices qemu creates for `if=virtio` — so the seed became `/dev/vda`, the rootfs and var disks shifted, and the VM dropped into emergency mode with `root=/dev/vda` pointing at a btrfs seed. Fixing the guest's lookup broke the two consumers that address disks positionally: the kernel cmdline and the guest fstab.

Every drive now stays on `if=virtio` so attaching one is purely additive, and the guest finds the seed by filesystem UUID. Reading that UUID is a 16-byte read at a fixed superblock offset rather than shelling out to `blkid` — the hosts this VM exists for have neither blkid nor btrfs. Nothing is cfg-gated; it behaves the same on macOS, Linux and Windows.

Tests pin the ordering property directly: no seed emits one `if=virtio` drive, a seed emits exactly two, the seed never precedes the rootfs, and no explicit `-device` appears.

## Also included

`ext/build.rs` now reads `[Install]` from drop-ins, not just the unit body. systemd honours a `WantedBy=` declared in a drop-in, so the standard way to enable a vendor unit shipping without an `[Install]` was invisible to the build. Parsed one file at a time — a single `sed` over several files shares an input stream, so the `[Install]` range would run past a file boundary. Test runs the generated fragment with `sh` against a real sysroot and asserts the `.wants` symlink appears; it fails against the previous parser.

## What this removes from #196

Nothing is destroyed, so `--reset-var`, the typed-`update` prompt and the `var_reset` event all go. The JSON event becomes `var_seed_sync_pending`, which host applications can ignore rather than act on. `record_var_size_floor` goes too — it existed only so a grown disk survived the delete.

`min_cli_version` — the most valuable part of #196 — is untouched.

## Testing

1412 unit tests, clippy clean.

**End to end on a real VM**: stock 0.3.0 migrated to a locally built release, 8/8 assertions — var.btrfs the same inode throughout, active runtime changed, all extensions merged, and the user's Docker volume, image layers and `/var` marker intact. Kernel 6.6.123 → 6.18.35, extensions 2024.1.0 → 0.1.0.

## Depends on

avocado-vm#3 (guest unit) and avocado-vm#4. Without #3 the drive is attached and nothing reads it — exactly today's behaviour, so they can land in either order.

### DIFF
--- a/src/commands/vm/update.rs
+++ b/src/commands/vm/update.rs
@@ -9,28 +9,33 @@
 //! - `seed_only` — the `var` image. Fetched on first install and, on a
 //!   version bump, re-fetched and re-seeded. See below.
 //!
-//! ## Why a version bump resets `var`
+//! ## Why a version bump needs more than the boot artifacts
 //!
-//! `seed_only` was written to preserve user state across image bumps,
-//! on the assumption that the in-VM agent would migrate the existing
-//! btrfs at boot. That
-//! migration does not exist yet, and `var` holds more than user state:
-//! the VM's own system extensions live in `/var/lib/avocado/images` and
-//! are merged onto `/usr` and `/etc` at boot. Carrying `var` across a
-//! bump therefore boots a new kernel and rootfs against the *old*
-//! userspace.
+//! `var` holds more than user state: the VM's own system extensions
+//! live in `/var/lib/avocado` and are merged onto `/usr` and `/etc` at
+//! boot. Replacing only kernel/initramfs/rootfs therefore boots a new
+//! kernel against the *old* userspace, and nothing catches it — each
+//! extension's `extension-release` carries `ID=_any`, systemd-sysext's
+//! "matches any OS" wildcard, so mismatched extensions merge silently.
 //!
-//! Nothing catches that. Each extension's `extension-release` is
-//! synthesized at merge time carrying `ID=_any`, which is
-//! systemd-sysext's explicit "matches any OS" wildcard — so the
-//! mismatched extensions merge silently and the VM looks healthy.
+//! It cannot be fixed by carrying the new seed's `var` wholesale
+//! either: the same filesystem holds the user's Docker volumes and
+//! installed SDKs (`$AVOCADO_PREFIX` is a Docker named volume inside
+//! the VM), which are expensive to rebuild and entirely theirs.
 //!
-//! Until the guest can migrate in place, a version bump takes the new
-//! seed and discards the live disk. That costs the user their Docker
-//! volumes, SDK images and `/data` work, so they must re-run `avocado
-//! install` and `avocado build` — a tradeoff deliberately accepted over
-//! shipping a VM whose halves don't match. An unchanged `var` sha means
-//! no reset.
+//! So the two lifetimes are separated at the granularity that already
+//! distinguishes them. The avocado-owned half — `runtimes/`, `images/`
+//! and the `active` pointer — is content-addressed and versioned, so
+//! the guest can install the new runtime *alongside* the old one out of
+//! the new seed and switch `active` only once it is fully staged.
+//! `vm update` records the seed sha here; `vm start` attaches that seed
+//! read-only; the guest does the work in early boot, before extensions
+//! merge. Nothing the user owns is touched, and every failure leaves
+//! the VM on its previous runtime rather than on none.
+//!
+//! The btrfs work has to happen guest-side regardless: this VM exists
+//! because the host is macOS or Windows, neither of which can mount a
+//! btrfs image.
 //!
 //! Behaviour with a running VM: query lifecycle, stop it cleanly,
 //! perform the swap, restart with the same `start` options. The
@@ -60,10 +65,6 @@ pub struct UpdateCommand {
     pub channel: Option<String>,
     pub check_only: bool,
     pub assume_yes: bool,
-    /// Authorizes the var reset when combined with `--yes`. Interactive
-    /// runs consent at the typed prompt instead, so this is only
-    /// consulted when `assume_yes` is set.
-    pub reset_var: bool,
     pub output: OutputFormat,
 }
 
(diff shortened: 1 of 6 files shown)


mobileoverlord:
End-to-end tested — stock avocado-vm 0.3.0 migrated to a locally built release with the seed attached by this change, 8/8. Results in the PR body under **Testing**: `var.btrfs` the same inode throughout, active runtime changed, all extensions merged, and the user's Docker volume, image layers and `/var` marker intact. Kernel 6.6.123 → 6.18.35.

Nothing needed re-running `avocado install` or `avocado build` afterwards, which is the whole point of replacing the reset in #196.

Guest half: avocado-vm#3 + #4, with avocadoctl#18 + #19 in the guest binary.

src/utils/vm/lifecycle.rs
```diff
@@ -685,6 +706,104 @@ fn seed_var_disk(paths: &VmPaths, manifest: &Manifest, artifact_dir: &Path) -> R
...
+    paths: &VmPaths,
+    manifest: &Manifest,
+    artifact_dir: &Path,
+) -> Option<(String, PathBuf)> {
+    let pending = super::config::VmConfig::load(paths)
+        .ok()?
+        .runtime?
+        .pending_var_seed_sha?;
+
+    if !paths.var_disk().exists() {
```
lee-reinhardt, line 763: This `exists()` guard is dead in the `start()` path: `seed_var_disk` runs earlier (line 155) and recreates a missing disk from this same seed, so the disk always exists by the time this runs. And `vm reset` / `rebuild --reset-data` delete the disk without clearing `pending_var_seed_sha`, so the next start byte-copies the seed via `seed_var_disk` and then attaches the seed beside its copy — the two-devices-one-fsid state the doc comment forbids, reachable via a documented command. Suggest guarding by comparing `btrfs_fsid` of the live disk against the seed instead of existence, and clearing the marker wherever the var disk is deleted.

src/commands/vm/update.rs
```diff
@@ -316,37 +298,43 @@ impl UpdateCommand {
...
+            // or Windows and has no btrfs.
+            //
+            // Nothing is destroyed: the new runtime is installed
+            // alongside the old and `active` only moves once it is
+            // fully staged, so every failure leaves the VM on its
+            // previous runtime rather than on none.
+            let new_sha = new_manifest
+                .artifact("var")
+                .map(|a| a.sha256.clone())
+                .expect("should_sync_var_seed only fires on a var artifact");
```
lee-reinhardt, line 316: `should_sync_var_seed` matches any `seed_only` artifact, but this resolves `artifact("var")` specifically — a manifest whose seed_only artifact has a different name panics here mid-update, after the VM teardown. Derive the sha from the same artifact the gate matched, or make the gate require `artifact("var")`.

src/utils/vm/lifecycle.rs
```diff
@@ -685,6 +706,104 @@ fn seed_var_disk(paths: &VmPaths, manifest: &Manifest, artifact_dir: &Path) -> R
...
+    target: &super::ssh::SshTarget,
+    expected_sha: &str,
+) {
+    let Ok((applied, _)) = target
+        .exec(&format!("cat {GUEST_SEED_STAMP} 2>/dev/null || true"))
+        .await
+    else {
+        return;
+    };
+    if !applied.trim().eq_ignore_ascii_case(expected_sha) {
```
lee-reinhardt, line 790: The stamp is a bare sha that persists in /var across boots, so it can vouch for a sync that did not happen this boot. Example: update to sha X, guest syncs and stamps X; later the channel rolls back and forward so `pending_var_seed_sha` is X again, but this boot the guest sync fails (runtime GC'd, copy failed) — the old stamp still reads X, the marker clears, and the retry-on-next-start story silently ends while the VM runs the wrong runtime. A per-update nonce on the cmdline, echoed into the stamp, would make only this boot's sync able to clear this update's marker.

src/utils/vm/lifecycle.rs
```diff
@@ -685,6 +706,104 @@ fn seed_var_disk(paths: &VmPaths, manifest: &Manifest, artifact_dir: &Path) -> R
...
+        crate::utils::output::print_warning(
```
lee-reinhardt, line 791: This warning is suppressed under `--output json`, and there is no applied/failed counterpart to `var_seed_sync_pending` — so a host app driving `vm start` in JSON mode cannot tell "extensions updated" from "sync failed, still on the old runtime" and will report success on stale extensions. Worth emitting a `var_seed_sync_applied` / `var_seed_sync_failed` event on this path.

PR #208, opened 2026-08-17
fix(image): fold permissions into the OS build id so they OTA

nicksinas:
ENG-2437 

## Problem

A rootfs change that only edits `permissions:` (users/groups/passwords) builds fine locally but **never lands on a device over `avocado deploy`**. The device logs:

```
OS already at target version (AVOCADO_OS_BUILD_ID=…), skipping OS bundle download
```

…and no-ops.

## Root cause

`AVOCADO_OS_BUILD_ID` is a pure function of the rootfs **rpmdb** — `uuid5(namespace, sha256(sorted NEVRA of every installed RPM))`. The `permissions:` section rewrites `/etc/{passwd,shadow,group}` in the image work dir but **never touches the package DB**, so an identical package set yields an identical build id regardless of the auth-file bytes. The deploy-side gate compares build ids and short-circuits both the download and the apply of the OS bundle, so the change silently never ships.

(Same class of blind spot existed for the initramfs build id.)

## Fix

Fold a hash of the assembled auth files into the build id **alongside** the NEVRA hash:

- Rootfs: `uuid5(ns, "$PKG_HASH:$AUTH_HASH")`
- Initramfs: `uuid5(ns, "$INITRAMFS_PKG_HASH:$INITRAMFS_AUTH_HASH")`

`AUTH_HASH` is produced by a new shared helper, `render_auth_files_hash`, that lives beside the existing `render_hook_block` / `resolve_install_hooks` / `render_users_groups_script` helpers and is imported by `initramfs::image` — so **rootfs and initramfs derive their build id the exact same way**, from one implementation.

```sh
AUTH_HASH=$(cat \
    "$ROOTFS_WORK/etc/passwd" \
    "$ROOTFS_WORK/etc/shadow" \
    "$ROOTFS_WORK/etc/group" \
    "$ROOTFS_WORK/etc/gshadow" 2>/dev/null | LC_ALL=C sort | sha256sum | awk '{print $1}')
```

- **`LC_ALL=C sort`** makes the digest independent of the order the permissions section appends entries in (the `users:`/`groups:` maps have no guaranteed iteration order), so it's fully deterministic.
- **`2>/dev/null`** tolerates a missing `gshadow` (minimal images) — absent files contribute nothing rather than aborting.
- Computed **after** the permissions section runs, so it observes the change.

## Why this doesn't cause spurious updates

The id is a content hash: an **unchanged** `permissions:` block hashes identically → same id → no OTA. It moves **if and only if** the auth-file content actually changes. This is the minimal correct mechanism — the deploy gate reads nothing but `AVOCADO_OS_BUILD_ID`, so a permissions-only change *has* to move the id to land.

## One-time migration note

Because the build-id formula changes (`$PKG_HASH` → `$PKG_HASH:$AUTH_HASH`), every existing image re-ids once on the first rebuild after this lands, triggering a **single** OS OTA even when nothing else changed. Steady state is churn-free thereafter.

## Tests

- `test_render_auth_files_hash_snippet` — exact-output unit test on the pure helper + reuse under the initramfs var names.
- `test_build_id_folds_auth_files` (rootfs **and** initramfs) — structural + ordering assertions (auth hash computed after the permissions section, folded into the uuid5 input).
- Full lib suite: 1452 passing; `cargo fmt` / `cargo clippy` clean on the changed files.

## Scope / follow-up

This is the correctness fix only. A separate `--force` override for `avocado deploy` (an escape hatch for any future rootfs change that doesn't move the id) was intentionally split out into its own PR — this fix largely subsumes the `permissions:` case, so `--force` is a nice-to-have rather than a dependency.

Fixes ENG-2437.

### DIFF
--- a/src/commands/rootfs/image.rs
+++ b/src/commands/rootfs/image.rs
@@ -49,6 +49,30 @@ echo \"Applied systemd presets\"; fi",
     "echo \"Generated ld.so.cache\"",
 ];
 
+/// Render the shell snippet that hashes an image's identity/auth files
+/// (`/etc/{passwd,shadow,group,gshadow}`) into `out_var`.
+///
+/// Folded into the build id alongside the package NEVRA hash so a
+/// `permissions:`-only change — which rewrites these files but never the
+/// rpmdb — still moves the build id and therefore OTAs (ENG-2437). Shared
+/// by the rootfs and initramfs build scripts so both derive their build id
+/// the same way.
+///
+/// `work_var` names the image work-dir shell variable (e.g. `ROOTFS_WORK`,
+/// `INITRAMFS_WORK`). `LC_ALL=C sort` makes the hash independent of the
+/// order the permissions section appended entries in (the `users:`/`groups:`
+/// maps have no guaranteed iteration order); missing files (e.g. no
+/// base-passwd) contribute nothing rather than aborting.
+pub fn render_auth_files_hash(work_var: &str, out_var: &str) -> String {
+    format!(
+        r#"    {out_var}=$(cat \
+        "${work_var}/etc/passwd" \
+        "${work_var}/etc/shadow" \
+        "${work_var}/etc/group" \
+        "${work_var}/etc/gshadow" 2>/dev/null | LC_ALL=C sort | sha256sum | awk '{{print $1}}')"#
+    )
+}
+
 /// Render a list of user-supplied shell commands as an indented block,
 /// preceded by a one-line "Running … hooks" echo for log clarity. Empty
 /// input returns an empty string so the surrounding script stays clean.
@@ -162,7 +186,13 @@ if [ -d "$ROOTFS_SYSROOT/usr" ]; then
     # shift with the container's collation.
     PKG_NEVRA=$(rpm --dbpath /var/lib/rpm -qa --queryformat '%{{NEVRA}}\n' --root "$ROOTFS_SYSROOT" | LC_ALL=C sort)
     PKG_HASH=$(echo "$PKG_NEVRA" | sha256sum | awk '{{print $1}}')
-    OS_BUILD_ID=$(python3 -c "import uuid; print(uuid.uuid5(uuid.UUID('{namespace_uuid}'), '$PKG_HASH'))")
+
+    # Fold the assembled auth files into the build id so a `permissions:`-only
+    # change — which the NEVRA set above is blind to — still moves the id and
+    # OTAs (ENG-2437). See render_auth_files_hash.
+{auth_hash_block}
+
+    OS_BUILD_ID=$(python3 -c "import uuid; print(uuid.uuid5(uuid.UUID('{namespace_uuid}'), '$PKG_HASH:$AUTH_HASH'))")
 
     # Inject identity into os-release (work copy for the image, sysroot for stone)
     # Strip any prior injected fields from the work copy before appending
@@ -218,6 +248,7 @@ fi"#,
         rootfs_filesystem = rootfs_filesystem,
         post_install_block = post_install_block,
         permissions_section = permissions_section,
+        auth_hash_block = render_auth_files_hash("ROOTFS_WORK", "AUTH_HASH"),
     )
 }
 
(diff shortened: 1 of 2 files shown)


src/commands/rootfs/image.rs
```diff
@@ -49,6 +49,30 @@ echo \"Applied systemd presets\"; fi",
...
+/// order the permissions section appended entries in (the `users:`/`groups:`
+/// maps have no guaranteed iteration order); missing files (e.g. no
+/// base-passwd) contribute nothing rather than aborting.
+pub fn render_auth_files_hash(work_var: &str, out_var: &str) -> String {
+    format!(
+        r#"    {out_var}=$(cat \
+        "${work_var}/etc/passwd" \
+        "${work_var}/etc/shadow" \
+        "${work_var}/etc/group" \
+        "${work_var}/etc/gshadow" 2>/dev/null | LC_ALL=C sort | sha256sum | awk '{{print $1}}')"#
```
mobileoverlord: `2>/dev/null` hides the message but not the exit status, so this doesn't tolerate a missing file — it aborts the build.

The standalone image paths run this fragment under `set -euo pipefail` (`rootfs/image.rs:362`, `initramfs/image.rs:287`). With `pipefail`, a missing `gshadow` makes `cat` fail the whole pipeline, and `AUTH_HASH=$(...)` takes that status, so `set -e` kills the build — with no output at all, since the only diagnostic went to `/dev/null`:

```
$ bash -c 'set -euo pipefail; H=$(cat w/etc/{passwd,shadow,group,gshadow} 2>/dev/null | LC_ALL=C sort | sha256sum | awk "{print \$1}"); echo "$H"'
exit=1     # nothing printed
```

This is exactly the minimal-image case the redirect was added for. It bites the initramfs hardest — a minimal initramfs plausibly has neither `/etc/shadow` nor `/etc/gshadow`, and the block is emitted unconditionally even when no `permissions:` is configured.

Worth noting `runtime build` happens to survive it: that script runs via plain `bash -c` + entrypoint `set -e` with no `pipefail`, so the assignment takes `awk`'s status. So the failure mode is "`avocado rootfs image` / `avocado initramfs image` break, `avocado runtime build` doesn't" — an unpleasant one to debug.

A brace group fixes it (verified: exit 0, stable digest, with `gshadow` absent):

```suggestion
        r#"    {out_var}=$({{ cat \
        "${work_var}/etc/passwd" \
        "${work_var}/etc/shadow" \
        "${work_var}/etc/group" \
        "${work_var}/etc/gshadow" 2>/dev/null || true; }} | LC_ALL=C sort | sha256sum | awk '{{print $1}}')"#
```

`test_render_auth_files_hash_snippet` asserts the exact string so it needs the same update — worth adding an assertion that the snippet is `pipefail`-safe while you're there.

mobileoverlord (commented):
Approach looks right and appropriately scoped — content hash folded in beside the NEVRA hash, one shared helper for both images, computed after the permissions section. Ordering checks out in both generated scripts, and the sysroot os-release gets the same id the manifest patcher reads back (`runtime/build.rs:2455`), so no split-brain. Migration note is accurate.

One blocking issue inline: the missing-file tolerance doesn't survive `pipefail`.

Two follow-ups filed rather than piled onto this PR:
- ENG-2440 — `overlay:` contents aren't content-hashed by the install stamps, so an overlay edit leaves the sysroot stale and never reaches the image at all. Same blind spot, one layer earlier.
- ENG-2441 — derive the id from the assembled work tree instead of an enumerated component list, on one shared rootfs/initramfs code path, so this doesn't need a third patch. Notes the `HashMap` → `BTreeMap` determinism prerequisite (which is also what the `LC_ALL=C sort` here is working around).

src/commands/rootfs/image.rs
nicksinas: confirmed the pipefail abort and fixed it with your brace-group suggestion in f95ce9b. Updated the exact-string test and added a pipefail-safety assertion

jetm (commented):
Two blocking findings inline, both reproduced. The core intent (fold permissions into the build id so a `permissions:`-only change OTAs, per ENG-2437) is right and needed; the hash input isn't order-stable yet, and the id rotates unconditionally for every project regardless of whether `permissions:` is even set. Everything else - the four-file staging guard, the pipefail-safe `{ …; } || true` wrapping, the passwd/shadow sed application - checked out clean (1446 tests pass, clippy/fmt clean, ran the rendered script directly under `set -euo pipefail` with files missing).

src/commands/rootfs/image.rs
```diff
@@ -49,6 +49,36 @@ echo \"Applied systemd presets\"; fi",
...
+/// (`/etc/{passwd,shadow,group,gshadow}`) into `out_var`.
+///
+/// Folded into the build id alongside the package NEVRA hash so a
+/// `permissions:`-only change — which rewrites these files but never the
+/// rpmdb — still moves the build id and therefore OTAs (ENG-2437). Shared
+/// by the rootfs and initramfs build scripts so both derive their build id
+/// the same way.
+///
+/// `work_var` names the image work-dir shell variable (e.g. `ROOTFS_WORK`,
+/// `INITRAMFS_WORK`). `LC_ALL=C sort` makes the hash independent of the
```
jetm, line 62: **`LC_ALL=C sort` can't make this deterministic - the inputs it's sorting are already order-dependent.** `sort` normalizes *line order* only. It can't normalize the uid/gid values baked into each line, or the member order *within* one `/etc/group` line. `PermissionsConfig.users`/`groups` are `HashMap<String, serde_yaml::Value>` (`config.rs:840-841`), and `mapping_from_hashmap` (`permissions.rs:380`) copies them via `for (k, v) in map` with no sort - Rust's `HashMap` iteration order is randomized per process. `stamps.rs:1283-1292` documents exactly this failure mode for the same reason ("HashMap iteration order varies per process... so the keys have to be sorted here or the hash is unstable between runs") and sorts before hashing; `mapping_from_hashmap` doesn't.

Ran the real `mapping_from_hashmap` -> `render_users_groups_script` path in 4 separate processes with identical input: 3 distinct user-emission orders. For a user declared without an explicit `uid:` (optional per `configs/default.yaml:82-86`), that changes the assigned uid/gid, and for two users sharing a supplementary group, it changes the member order within one `/etc/group` line - both survive the line-level `sort` and change `AUTH_HASH`.

Before this PR the build id was `PKG_HASH`-only and stable across rebuilds. After, a rebuild that changed nothing in the config can OTA the whole fleet, contradicting "deterministic build ID" at line 138. Fix upstream of the hash - sort the keys in `mapping_from_hashmap`, the same way `packages_for_hash` already does.

src/commands/rootfs/image.rs
```diff
@@ -162,7 +192,13 @@ if [ -d "$ROOTFS_SYSROOT/usr" ]; then
...
     PKG_NEVRA=$(rpm --dbpath /var/lib/rpm -qa --queryformat '%{{NEVRA}}\n' --root "$ROOTFS_SYSROOT" | LC_ALL=C sort)
     PKG_HASH=$(echo "$PKG_NEVRA" | sha256sum | awk '{{print $1}}')
-    OS_BUILD_ID=$(python3 -c "import uuid; print(uuid.uuid5(uuid.UUID('{namespace_uuid}'), '$PKG_HASH'))")
+
+    # Fold the assembled auth files into the build id so a `permissions:`-only
+    # change — which the NEVRA set above is blind to — still moves the id and
+    # OTAs (ENG-2437). See render_auth_files_hash.
+{auth_hash_block}
+
+    OS_BUILD_ID=$(python3 -c "import uuid; print(uuid.uuid5(uuid.UUID('{namespace_uuid}'), '$PKG_HASH:$AUTH_HASH'))")
```
jetm, line 201: **This rotates the build id for every project, even one with no `permissions:` block at all.** The uuid5 input changes unconditionally from `'$PKG_HASH'` to `'$PKG_HASH:$AUTH_HASH'`, and `AUTH_HASH` is the digest of the *base* packages' own passwd/shadow/group - never empty, even with zero `permissions:` config. The id is appended to os-release before `mkfs.erofs` runs (line 206 vs 220), so this changes the image bytes, sha256, `image_id`, spot hashes, and the re-signed AMF for every build across the merge boundary.

First build after this merges, zero config change, zero package change: rootfs + initramfs + os-bundle all re-upload and every device takes a full OS OTA for nothing. Worth a CHANGELOG rollout note - this repo already carries one for smaller changes (the `avocado --version` entry under `[Unreleased]`) - so downstream fleets know to expect one unscheduled full OTA rather than reading it as a regression.

jetm (approved):

mobileoverlord (approved):
Correct and symmetric: reuses the rootfs side's render_auth_files_hash, folds it into the uuid5 input after the permissions section runs (the test pins that ordering, which is the part that could silently regress), and closes the permissions-only-change-never-OTAs hole.

PR #210, opened 2026-08-19
fix(stamps): hash overlay contents for verbatim overlays too

nicksinas:
Fixes ENG-2440.

## Problem

Edit a file under your `overlay/` dir, run `avocado build`, deploy — and the built image still contains the **old** file. No warning, no error. `avocado install` first fixes it, which is why the habitual `avocado install && avocado build` masks it.

## Root cause

`overlay:` is applied to the sysroot by `cp` (in install), not by RPM. The install stamp is a **prerequisite gate, not a work-skipping cache**: `avocado build` only validates that the stamp exists/isn't stale, it never re-runs install. So an *under*-invalidating stamp doesn't cause redundant work — it causes silently wrong output.

The overlay content digest (sorted path + mode + content-sha256) already existed in `overlay_preprocess::overlay_content_digest`, but `fold_overlay_content_hash` **gated it behind `PreprocessSpec::is_enabled()`** — so a *verbatim* overlay (the common case, no `preprocess:`) hashed only the config value (dir name + mode), never file contents. Editing an overlay file left the stamp current.

## Fix

- Drop the `is_enabled()` gates in `overlay_content_digest` and `fold_overlay_content_hash` so the content digest is **always** folded in — raw bytes for verbatim overlays (the existing `process_file_bytes` returns raw when nothing is selected for preprocessing), post-`{{ }}` content for preprocessed ones. All three call sites (rootfs, initramfs, ext-build) already route through `fold_overlay_content_hash`, so all are fixed at once.
- Move `parse_overlay_config` into `overlay_preprocess.rs` as the single shared parser, so a **bare-string** overlay (`overlay: mydir`) hashes the right tree instead of the `"overlay"` default.

Now editing an overlay file makes the stamp stale and `avocado build` tells the user to re-run install — exactly as it already does for a `post_install` edit.

## Note on the sibling ticket

Even once the file reaches the image, `AVOCADO_OS_BUILD_ID` was blind to it (a `cp` never touches the rpmdb), so the OTA gate no-ops — same class as ENG-2437. That's fixed generally in ENG-2441 (build id from the assembled work tree), not here.

## Tests

- Inverted the two tests that encoded the bug (`rootfs_verbatim_overlay_ignores_file_contents` → `…_hashes_file_contents`; the "disabled yields no digest" assertion → verbatim digest tracks raw bytes).
- Added bare-string overlay coverage.
- Full lib suite green; `cargo fmt`/`clippy` clean on the changed files.

## Out of scope (noted in the ticket)

- Merge-mode overlay never *deletes* — removing a file from `overlay/` doesn't remove it from the persistent sysroot (needs a clean or opaque mode). No hashing scheme fixes this.
- `compute_runtime_build_input_hash` is write-only (deploy validates existence only) — separate decision.

### DIFF
--- a/src/commands/rootfs/install.rs
+++ b/src/commands/rootfs/install.rs
@@ -5,28 +5,7 @@ use std::collections::{HashMap, HashSet};
 use std::path::Path;
 use std::sync::Arc;
 
-/// Parse the `overlay:` config value into `(dir, opaque)`.
-/// Accepts either a plain string (`"path/to/dir"`) or a mapping
-/// (`{ dir: "path/to/dir", mode: "opaque" | "merge" }`).
-fn parse_overlay_config(value: &serde_yaml::Value) -> (String, bool) {
-    if let Some(dir_str) = value.as_str() {
-        (dir_str.to_string(), false)
-    } else if let Some(table) = value.as_mapping() {
-        let dir = table
-            .get("dir")
-            .and_then(|d| d.as_str())
-            .unwrap_or("overlay")
-            .to_string();
-        let opaque = table
-            .get("mode")
-            .and_then(|m| m.as_str())
-            .map(|m| m == "opaque")
-            .unwrap_or(false);
-        (dir, opaque)
-    } else {
-        ("overlay".to_string(), false)
-    }
-}
+use crate::utils::overlay_preprocess::parse_overlay_config;
 
 /// Build the shell snippet that applies an overlay directory into a sysroot.
 /// `overlay_dir` is the path relative to `/opt/src` (the project root inside the container).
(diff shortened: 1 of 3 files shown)


mobileoverlord (approved):
Complete against ENG-2440: the ungated content digest fixes all three call sites (rootfs/initramfs/ext-build) through the one existing path, the bare-string overlay parse fix covers a miss the ticket didn't catch, and the bug-encoding tests are inverted. The ticket's item 2 (collapse the rootfs/initramfs input-hash duplicates) turned out to already be on main via #185, so nothing is outstanding. Rebased onto current main; 3391 tests, clippy, fmt green locally.

